PR #7, opened 2014-04-13
APE6 - Enhanced Character Separated Values table format

taldcroft:
APE6 is primarily a specification of a new standard for the interchange of tabular data in a text-only format. The proposed format handles the key issue of serializing column specifications and table metadata by using a JSON-encoded data structure which is included in the file as a series of #-prefixed comment lines. The actual tabular data are stored in a standard comma-separated-values (CSV) format, giving compatibility with a wide variety of non-specialized CSV table readers. Using JSON makes it extremely easy for applications to read both the standardized data format elements (e.g. column name, type, description) as well as complex metadata structures. Support for schemas to describe and validate the metadata is part of the standard along with support for non-ASCII unicode character encoding.

The implementation in astropy.io.ascii is relatively straightforward and will provide a significant benefit of allowing text serialization of most astropy Table objects, persistent storage, and subsequent interchange with other users.

### DIFF
--- a/APE6.rst
+++ b/APE6.rst
@@ -1,5 +1,5 @@
-Data-table Text Interchange Format
-----------------------------------
+Enhanced Character-Separated-Values format
+-----------------------------------
 
 author: Tom Aldcroft
 
@@ -19,7 +19,7 @@ APE6 is primarily a specification of a new standard for the interchange of
 tabular data in a text-only format.  The proposed format handles the key issue
 of serializing column specifications and table metadata by using a YAML-encoded
 data structure.  The actual tabular data are stored in a standard
-comma-separated-values (CSV) format, giving compatibility with a wide variety of
+character-separated-values (CSV) format, giving compatibility with a wide variety of
 non-specialized CSV table readers.
 
 Using YAML makes it extremely easy for applications *and humans* to read both
@@ -27,21 +27,16 @@ the standardized data format elements (e.g. column name, type, description) as
 well as complex metadata structures.  YAML also lends itself to simple table
 modifications by humans in a plain text editor.
 
-The implementation in ``astropy.io.ascii`` is relatively straightforward and
+The reference Python implementation in ``astropy.io.ascii`` is relatively straightforward and
 will provide a significant benefit of allowing text serialization of most astropy
 Table objects, persistent storage, and subsequent interchange with other users.
 
 Existing standards
 --------------------
 
-The well-known `XKCD comic <https://xkcd.com/927/>`_ aptly mocks the
-introduction of new standards.  Let's just stipulate that this standard may be a
-bad idea, but talk about it anyway.  Following the lead of the astronomical data
-community in recently documenting shortcomings of the FITS standard, we start by
-discussing existing widely-used standards and why they are lacking.
-
-Note that the existing `Tabular Data Package`_ standard, while not
-used widely in astronomy, may be entirely suitable for our purposes.
+Following the lead of the astronomical data community in recently
+documenting shortcomings of the FITS standard, we start by discussing
+existing widely-used standards and why they are lacking.
 
 CSV
 ^^^^^
@@ -179,7 +174,7 @@ this APE we have become aware of a very similar standard known as the
 `Tabular Data Package
 <http://dataprotocols.org/tabular-data-package/>`_.  This provides a
 fully-formed protocol for publishing and sharing tabular-style data
-which is conceptually very similar to the proposed DTIF format, with
+which is conceptually very similar to the proposed ECSV format, with
 the exception of using two files, one pure JSON for the header and one
 pure CSV for the data.  The JSON header follows a schema defined by
 the `JSON Table Schema
@@ -195,15 +190,16 @@ YAML instead of JSON for serialization precludes direct use of the TDP.
 Detailed description
 ---------------------
 
-The proposed Data-table Text Interchange Format (DTIF) has the following
+The proposed Enhanced CSV (ECSV) format has the following
 overall structure:
 
-- A header section which consists of lines that start with the "#" character
-  and provide the table definition via a YAML-encoded data structure.
-- An initial line in the header section which identifies the file as DTIF and
+- A header section which consists of lines that start with the ``#`` character
+  and provide the table definition and data format via a YAML-encoded data structure.
+- An initial line in the header section which identifies the file as ECSV and
   provides a version number.
 - A CSV-formatted data section in which the first line contains the column names
-  and subsequent lines contains the data values
+  and subsequent lines contains the data values.
+
 
 Why YAML?
 ^^^^^^^^^^
@@ -214,19 +210,19 @@ other relevant metadata.  Those formats essentially invent custom serialization
 specifications that must be carefully implemented from scratch by any
 reader/writer application.
 
-DTIF takes the approach of defining a minimal standard for the the underlying
+ECSV takes the approach of defining a minimal standard for the the underlying
 data structure that is needed to define a table.  Then that structure is
 encoded or decoded using YAML.  Libraries for encoding and decoding YAML are
 widely used, very efficient, and easily available in all the most-frequently
 used programming languages (with the notable exception of IDL).
 
-Translating to / from the data structure provided by a DTIF header into the native
+Translating to / from the data structure provided by a ECSV header into the native
 structure that an application uses should generally be quite easy because the
 functional elements (e.g. column name, type) are ubiquitous.  Generally
 speaking manipulating data structures programmatically is easier than parsing
 textual data structure fields.
 
-The DTIF standard does not require that the YAML encoding be "pretty", but it
+The ECSV standard does not require that the YAML encoding be "pretty", but it
 is highly-recommended that applications format the YAML header to be legible
 to humans.  This is important because a key feature of YAML is that it is
 meant to be easily readable, and thus modifiable, by humans.
@@ -234,8 +230,9 @@ meant to be easily readable, and thus modifiable, by humans.
 Example
 ^^^^^^^^^^
 
-A couple of quick examples will put this in context.  First let's create a
-table and give it some custom attributes::
+A couple of quick examples using the reference Python implementation
+will put this in context.  First let's create a table and give it some
+custom attributes::
 
   >>> from astropy.table import Table
   >>> t = Table([[1, 4], [2, 3]], names=['a', 'b'])
@@ -250,11 +247,11 @@ table and give it some custom attributes::
     001   2
     004   3
 
-Now we write this to a file using the DTIF format and print it::
+Now we write this to a file using the ECSV format and print it::
 
-  >>> t.write('example.dtif', format='ascii.dtif')
-  >>> cat example.dtif
-  # %DTIF 1.0
+  >>> t.write('example.ecsv', format='ascii.ecsv')
+  >>> cat example.ecsv
+  # %ECSV 1.0
   # ---
   # columns:
   # - {name: a, unit: m / s, type: int64, format: '%03d'}
@@ -271,7 +268,7 @@ are included in CSV format with a space delimiter.
 Now we can read back the table and see that it has survived the round-trip
 to a text file::
 
-  >>> t2 = Table.read('example.dtif', format='ascii.dtif')
+  >>> t2 = Table.read('example.ecsv', format='ascii.ecsv')
   >>> print(t2)
     a    b 
   m / s  km
@@ -279,7 +276,7 @@ to a text file::
     001   2
     004   3
 
-To illustrate the full features of DTIF we create a table with meta data
+To illustrate the full features of ECSV we create a table with meta data
 (keywords and comments) as well as more detailed column meta data::
 
   >>> from astropy.utils import OrderedDict
@@ -294,8 +291,8 @@ To illustrate the full features of DTIF we create a table with meta data
 Now we write the table to standard out::
 
   >>> import sys
-  >>> t.write(sys.stdout, format='ascii.dtif')
-  # %DTIF 1.0
+  >>> t.write(sys.stdout, format='ascii.ecsv')
+  # %ECSV 1.0
   # ---
   # columns:
   # - {name: a, unit: m / s, type: int64, format: '%5.2f', description: Column A}
@@ -303,7 +300,7 @@ Now we write the table to standard out::
   #   type: int64
   #   meta:
   #     column_meta: {a: 1, b: 2}
-  # table_meta: !!omap
+  # meta: !!omap
   # - keywords: !!omap
   #   - {z_key1: val1}
   #   - {a_key2: val2}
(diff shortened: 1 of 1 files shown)


taldcroft:
@eteq @mdboom @embray @astrofrog - just a reminder that my *hope* is to get https://github.com/astropy/astropy/pull/2319 into 1.0, and to some extent that depends on APE-6.

I guess one question for me is whether the standards in ASDF need to be considered and somehow incorporated here to allow future integration.  You'll note from the previous discussion with @cdeil that I de-scoped APE-6 somewhat to be more tied to astronomy and more specifically for astropy.  In that context I think this is good enough right now, but I don't want to push something just out of my exuberance to have 2319 in 1.0.

embray:
How many people have had a chance to look at this?  Would it be worth advertising on the Astropy mailing list?  There's also an astrodataformats mailing list I'm on, while this is slightly outside the scope of that list I'm sure there are lots of people on it that would be interested in looking at this.

taldcroft:
@embray - I just advertised this on the astropy list.  It was previously discussed on astropy-dev.

PaulKuin:
I am perhaps partial to the solution for astronomical tables that I negotiated in 1992/3 with CDS (while  being the GSFC Astronomical Data Center head). Our goals were: store tables and metadata in a readable format; allow 1-to-1 conversion to FITS (binary) tables; allow one format to address multiple tables; provide some basic information on the tables, like format and including references, history; keep the tables themselves in files with just data.  I still stand by that solution, and CDS has since introduced XML based metadata which I am not familiar with.  Now known generally as the CDS table format, the requirements for easy use and access to further information on the data are still valid. I am sure there are many ways to implement this, but it would be good to consider to keep them completely interoperable. I think this is a drawback of the current astropy tables, but not all tables need to be up to the same level as some are for personal use, development, or whatever. So a certain amount of flexibility is needed in the amount of required metadata dependant on it purposed usage.  I have not yet checked out APE, but have gotten the impression that it would be useful to consider that at some point it may become too confusing or perhaps impossible to support all possible table implementations.

taldcroft:
@PaulKuin - thanks for the feedback.  CDS is discussed in the APE-6 text (https://github.com/taldcroft/astropy-APEs/blob/ape6/APE6.rst#cds--apj-machine-readable-table).  This format does has a long heritage and a wide user base, but in my opinion it also has limitations in this context.  I think that making a new format which addresses some of the limitations while being interoperable with CDS would be difficult and end up counter to the goal of being CSV-compliant and simple.  Using XML is counter to the idea of human-readable/writeable files.

One of the features which I think is quite valuable in ECSV is that in the simple cases the header reduces down to something very short and easily readable.  E.g.
```
# %ECSV 1.0
# ---
# columns:
# - {name: a, unit: m / s, type: int64, format: '%03d'}
# - {name: b, unit: km, type: int64, description: This is column b}
a b
001 2
004 3
```
In all cases the format is pretty straightforward for machines to read as well, which would hopefully work toward wider adoption.

PaulKuin:
Yes, true enough, but what about other metadata?   The idea was to have a
"ReadMe" file to put the metadata.  I think that a compromise would be to
limit the header metadata in a data file to the names of the column and a
pointer to another file, like the 'ReadMe', containing the rest of the
metadata.

In other words, what you are proposing is to restrictive (trying to put all
the metadata in the file header) and not extensible (something that the
ReadMe file of metadata could do).  In my honest opinion you are
reinventing the wheel.

Sorry to have to say that.

    Paul

On Thu, Jan 15, 2015 at 11:12 PM, Tom Aldcroft <notifications@github.com>
wrote:

> @PaulKuin <https://github.com/PaulKuin> - thanks for the feedback. CDS is
> discussed in the APE-6 text (
> https://github.com/taldcroft/astropy-APEs/blob/ape6/APE6.rst#cds--apj-machine-readable-table).
> This format does has a long heritage and a wide user base, but in my
> opinion it also has limitations in this context. I think that making a new
> format which addresses some of the limitations while being interoperable
> with CDS would be difficult and end up counter to the goal of being
> CSV-compliant and simple. Using XML is counter to the idea of
> human-readable/writeable files.
>
> One of the features which I think is quite valuable in ECSV is that in the
> simple cases the header reduces down to something very short and easily
> readable. E.g.
>
> # %ECSV 1.0
> # ---
> # columns:
> # - {name: a, unit: m / s, type: int64, format: '%03d'}
> # - {name: b, unit: km, type: int64, description: This is column b}
> a b
> 001 2
> 004 3
>
> In all cases the format is pretty straightforward for machines to read as
> well, which would hopefully work toward wider adoption.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/astropy/astropy-APEs/pull/7#issuecomment-70180746>.
>



-- 

* * * * * * * * http://www.mssl.ucl.ac.uk/~npmk/ * * * *
Dr. N.P.M. Kuin      (n.kuin@ucl.ac.uk)
phone +44-(0)1483 (prefix) -204927 (work)
mobile +44(0)7806985366  skype ID: npkuin
Mullard Space Science Laboratory  – University College London  –
Holmbury St Mary – Dorking – Surrey RH5 6NT–  U.K.

hamogu:
I don't see that the CDS format and APE6 are in competition. CDS is a general format for astronomical tables that is sanctioned by major astronomical institutions (CDS, ApJ, A&A) and it is set up to contain metadata that follows a certain standard and is of long-term interest.

The format suggested in APE6 is can store a table I am currently working on between two python sessions or I can quickly email it to a collaborator who uses the exact same software as I do. It handles arbitrary metadata in an automatic and machine readable way, e.g. I can have a column whose metatdata is a nested list of python dictionaries. Such a structure would be hard to encode in a CDS table and even if I managed to put it in the README file, this specific file would only be readable for me and my specific code.

In my opinion, APE6 and CDS seem to address opposite requirements: CDS has a standardized format with fixed conventions (e.g. all CDS table use "e_x" as column names for the uncertainty on column "x") while APE6 allows for total free form meta data - that is great for data exchange between collaborators or persistent storage between my python sessions. APE6 tries to go a little bit beyond that by defining a handful of "standard keywords" (e.g. "name" and "unit") and keeping the data in a CSV format that many programs can read, but it does not aim for the complete meta-data header that is found in most CDS tables.

taldcroft:
@PaulKuin - As @hamogu has mentioned, the allowed metadata is essentially unrestricted.  There is a `meta` tag for the entire table and a `meta` tag for each column (in addition to a few predefined meta attributes like `name`, `format`, `unit`, `description`).  This `meta` can be any data structure that is serializable by YAML, so the metadata set in a CDS file could be easily encoded.  Then it can be read back by anyone with a YAML parser (i.e. most common languages).

To my knowledge none of the available astronomical ASCII data formats support (essentially) arbitrary meta data and place the burden of parsing header meta on a standard and widely available library (YAML).  In that sense I would argue that this is not entirely re-inventing the wheel.  

Of course there *is* the Tabular Data Package (TDP) that uses JSON.  This preceded ECSV and shares much of the design philosophy.  But after some discussion here we decided that YAML would be a better fit since it is far easier to read/write for humans.  On the point of re-inventing TDP as ECSV, definitely guilty.

PaulKuin:
@notifications@github.com ,  @hamogu <https://github.com/hamogu> -

In 1993 we were faced with Tables entered by hand that needed to go through
a process of verification. The CDS format provided restrictions so that the
process could be pipelined. restricting the name space was important for
CDS in order to automate loading the databases used for SIMBAD etc. .  A
tight standard also allows for applications to be written to use the data.

But quit rightly you mention good reasons to use YAML parsing, some of
which are very similar to those we had when developing the CDS format. The
main drawback of using the CDS format now is that it is providing barriers
rather then take them away for the kind of use that you envision. Frankly,
if the table generation works as it does now, with options to add metadata
(+metadata field) a bit like HDS allows, that sounds to me to be the way to
go.  If data structures can be supported that sounds even better.
The challenge there will be the readability.

 A separate application could be made to enforce the kind of restrictions
that the CDS format needs so that the data can be converted with relatively
the same effort as it takes now to do it by hand. The astropy tables would
thus only partly overlap with CDS tables, could make CDS table creation not
more difficult, perhaps even easier, and add lots of needed functionality.

PBarmby:
I am no kind of tabular-data-interchange-format expert, but I have spent quite a bit of time wrestling data out of IPAC/CDS/LaTeX etc formats into something usable. The standard proposed here seems like it would make my life easier, so I'm all for it. I would argue strongly in favour of having *all* metadata in the same file as the actual tabular data; I think there is too much potential for trouble with a separate ReadMe/header file.

As for "is this astronomy-specific enough?", well, it addresses a need that astronomers have, so that seems like enough to me.

taldcroft:
One possible change here relates to having a format that is easily embeddable in ASDF.  This was part of the driver for using YAML in ECSV (formerly known as DTIF which used JSON).  In the current [asdf-standard doc](http://asdf-standard.readthedocs.org/en/latest/) there is an example of an ndarray table with a header like:
```
!core/ndarray
  source: 0
  shape: [64]
  dtype:
    - name: coordinate
      dtype:
        - name: ra
          dtype: float64
        - name: dec
          dtype: float64
    - name: kernel
      dtype: float32
      shape: [3, 3]
  byteorder: little
```
The YAML for ECSV looks like:
```
  columns:
   - {name: a, unit: m / s, type: int64, format: '%5.2f', description: Column A}
   - name: b
     type: int64
     meta:
       column_meta: {a: 1, b: 2}
```
So one possibility is to change the top-level `columns` to `dtype` and also the column `type` to `dtype`.  I chose the original names for human readability and to be a bit more implementation agnostic (where dtype obviously comes from numpy).  However, the schema proposed in ASDF naturally supports nested tables via multiple layers of `dtype`.  I think that in theory the same could be achieved in ASDF using `columns` to introduce a tabular element while reserving `type` to specify only a simple data type, but it isn't clear that is the best way.

@mdboom @embray @perrygreenfield ?

mdboom:
I can see some advantages to both approaches to describing the layout of the data.  I think the ASDF one is more flexible in what structures it can describe, however ECSV allows for the extra metadata.  Maybe the thing to do is to follow ASDF's pattern, but add a place for arbitrary metadata and a unit?

Of course, both of these are limited to what Numpy can do -- for an example of where it might go in the future, Continuum IO's datashape project captures a lot more flexibility and real-world needs when describing data that might be closer to where all this ends up if we start thinking outside of the Numpy box.  But it's probably premature to make that leap yet.  My long-term plan is that a future version of ASDF may support a more general data description "language", but what we have now is 80% there and easy to support in a number of languages.

And lastly, on the term `dtype`.  I guess I was unintentionally following a Numpy bias there.  Since ASDF is by no means set in stone, maybe we change it to `datatype`, which seems sufficiently generic, but also more specific than just `type` which I find maybe a bit too ambiguous/overloaded.

taldcroft:
Ah, I like `datatype` as being descriptive and neutral.  If nobody raises objections I'll change ECSV to use `datatype` in place of the current `columns` and `type`.

taldcroft:
Made the change in 9480c56.

mdboom:
I've made the dtype -> datatype rename for ASDF in https://github.com/spacetelescope/pyasdf/pull/52 and https://github.com/spacetelescope/asdf-standard/pull/46

APE6.rst
```diff
@@ -0,0 +1,499 @@
...
+  a b
+  1.00 2
+  4.00 3
+
+In this case there are Ordered dictionary elements which must be preserved
+during serialization.  This is done via the ``!!omap`` element tag.  The
+reference Python implementation in Astropy uses custom Loader and Dumper
+classes to allow serialization to and from the Python ``OrderedDict`` class.
+Other implementations must likewise use an ordered mapping when reading and the
+``!!omap`` tag when writing for ordered mappings in the data structure.
```
embray: Perhaps it's useful to add some words about why order preservation for metadata is useful.  I've had to explain that a surprising number of times in the context of ASDF.  Some people think it's optional, but I think there's a strong case to be made that it is an essential feature.

APE6.rst
```diff
@@ -0,0 +1,499 @@
...
+An ECSV file begins with the the table header which contains the
+necessary information to define the table columns and metadata.  This
+is expressed as a YAML-encoded data structure which has a small set of
+required keywords and standard specifiers.
+
+Each line of the YAML-encoded data structure must start with the two
+characters ``# `` (hash followed by space) to indicate the presence of
+header content.  The first line which does not start with ``#``
+signifies the end of the header.  Subsequent lines starting with ``#``
+are treated as file comment lines.
```
embray: I think that if we will allow subsequent non-YAML lines we should require, or at least strongly recommend, ending the YAML portion with an explicit document end line
```
# ...
```

APE6.rst
```diff
@@ -0,0 +1,499 @@
...
+
+Within the header section, lines which start with ``##`` are treated as
+comments and can be ignored by readers.  There is no requirement for
+ECSV writers to emit such comment data.  Relevant comment strings
+should be serialized within the ``meta`` keyword structure.
```
embray: I think this line is confusing.  When you say "relevant comment strings" I assume you mean comments that are specified in the "comments" section of the YAML document.  That's fine, but on first reading this sounded contradictory.  The language here is a bit soft--"there is no requirement"--suggesting that a reader *may* try to parse these comments and include them in the ``meta`` keyword.  I would suggest that such comments *must* be ignored and no attempt made to parse them or round-trip them.

APE6.rst
```diff
@@ -0,0 +1,499 @@
...
+
+The data section follows immediately after the header.  Lines in the
+data section consisting only of zero or more whitespace characters
+(space and/or tab) are ignored.
+
+The first non-blank line in the data section must contain the column
+names formatted according to the CSV writer being used.  This allows
+most CSV reader applications to successfully read ECSV files and
+automatically infer the correct column names.  ECSV readers should
+validate that the column names in this line match those in the header.
```
embray: What should a reader do if they don't match?  I think a warning is sufficient, so long as the number of columns at least match.  Though what a "warning" means depends entirely on the implementation, and could be optionally upgraded to an error in a "strict" mode.

APE6.rst
```diff
@@ -0,0 +1,499 @@
...
+Following the column name line the data values are serialized according to
+the following rules:
+
+- Each row must contain the same number of delimiter-separated fields.
+- Fields are separated by the delimiter character, which can be either the
+  space or comma character.
+- Any field may be quoted with double quotes.
+- Fields containing a line-break, double-quote, and/or the delimiter character
+  must be quoted
+- Boolean fields are represented as the string ``False`` or ``True``.
```
embray: Case-sensitivity?

APE6.rst
taldcroft: What do you usually say as your first example of why it's needed?  For me it is crucial for human readability and file round-tripping, but I don't know of examples where there is a real contextual dependence on keyword order.

APE6.rst
```diff
@@ -0,0 +1,499 @@
...
+- A header section which consists of lines that start with the ``#`` character
+  and provide the table definition and data format via a YAML-encoded data structure.
+- An initial line in the header section which identifies the file as ECSV and
+  provides a version number.
+- A CSV-formatted data section in which the first line contains the column names
+  and subsequent lines contains the data values.
+
+Version 0.9 of the ECSV format specification and the reference Python
+implementation assumes ASCII-encoded header and data sections.  Support
+for unicode (in particular UTF-8) may be added in subsequent versions.
```
embray: Speaking of encodings, would we allow separate encodings for the data and header?  I think this could be useful.  Or should that discussion be tabled?

APE6.rst
taldcroft:
```
Each line of the YAML-encoded data structure must start with the two
characters ``# `` (hash followed by space) to indicate the presence of
header content.  All content within this header section must be parseable
as a single YAML document.  The first line which does not start with ``#``
signifies the end of the header and the start of the data section.  Subsequent 
lines within the data section starting with ``#`` are ignored.
```
I changed "treated as file comment lines" to "ignored" to make it clear that these "comments" are not expected to be available as part of the output and thus will not round-trip.

We *could* just say that comments in the data section are simply not allowed.  That's easy enough in the Python implementation and would reduce downstream ambiguity.  I'd be fine with that since it gives more clarity to the distinction between header and data.

PaulKuin: Comments inside data files:

Sometimes it is nice to be able to uncomment a data record but keep it in
its place in the data file. Useful for testing, or to flag problems with
that record that can be addressed later. If that cannot be done, the
solution is usually a bit more hassle, like versioning it or so.



On Wed, Jan 21, 2015 at 7:50 PM, Tom Aldcroft <notifications@github.com>
wrote:

> In APE6.rst
> <https://github.com/astropy/astropy-APEs/pull/7#discussion_r23328482>:
>
> > +``'description'``, and ``'meta'``.  This is not a ECSV requirement but is
> > +recommended for human accessibility.
> > +
> > +Header details
> > +^^^^^^^^^^^^^^^^
> > +
> > +An ECSV file begins with the the table header which contains the
> > +necessary information to define the table columns and metadata.  This
> > +is expressed as a YAML-encoded data structure which has a small set of
> > +required keywords and standard specifiers.
> > +
> > +Each line of the YAML-encoded data structure must start with the two
> > +characters ``# `` (hash followed by space) to indicate the presence of
> > +header content.  The first line which does not start with ``#``
> > +signifies the end of the header.  Subsequent lines starting with ``#``
> > +are treated as file comment lines.
>
> Each line of the YAML-encoded data structure must start with the two
> characters ``# `` (hash followed by space) to indicate the presence of
> header content.  All content within this header section must be parseable
> as a single YAML document.  The first line which does not start with ``#``
> signifies the end of the header and the start of the data section.  Subsequent
> lines within the data section starting with ``#`` are ignored.
>
> I changed "treated as file comment lines" to "ignored" to make it clear
> that these "comments" are not expected to be available as part of the
> output and thus will not round-trip.
>
> We *could* just say that comments in the data section are simply not
> allowed. That's easy enough in the Python implementation and would reduce
> downstream ambiguity. I'd be fine with that since it gives more clarity to
> the distinction between header and data.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/astropy/astropy-APEs/pull/7/files#r23328482>.
>



-- 

* * * * * * * * http://www.mssl.ucl.ac.uk/~npmk/ * * * *
Dr. N.P.M. Kuin      (n.kuin@ucl.ac.uk)
phone +44-(0)1483 (prefix) -204927 (work)
mobile +44(0)7806985366  skype ID: npkuin
Mullard Space Science Laboratory  – University College London  –
Holmbury St Mary – Dorking – Surrey RH5 6NT–  U.K.

taldcroft: OK, I'm leaving the possibility of comment lines in the data but specifying that they should not be parsed or returned by the reader.

PaulKuin: one more thing comes to mind.

i modified my reader when i needed to read in QDP files which use the !  in
column one to flag a commented out line. Can you make the symbol used in
col 1 a parameter/

(qdpo files are used much in the HEA community)


On Thu, Jan 22, 2015 at 1:21 AM, Tom Aldcroft <notifications@github.com>
wrote:

> In APE6.rst
> <https://github.com/astropy/astropy-APEs/pull/7#discussion_r23349532>:
>
> > +``'description'``, and ``'meta'``.  This is not a ECSV requirement but is
> > +recommended for human accessibility.
> > +
> > +Header details
> > +^^^^^^^^^^^^^^^^
> > +
> > +An ECSV file begins with the the table header which contains the
> > +necessary information to define the table columns and metadata.  This
> > +is expressed as a YAML-encoded data structure which has a small set of
> > +required keywords and standard specifiers.
> > +
> > +Each line of the YAML-encoded data structure must start with the two
> > +characters ``# `` (hash followed by space) to indicate the presence of
> > +header content.  The first line which does not start with ``#``
> > +signifies the end of the header.  Subsequent lines starting with ``#``
> > +are treated as file comment lines.
>
> OK, I'm leaving the possibility of comment lines in the data but
> specifying that they should not be parsed or returned by the reader.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/astropy/astropy-APEs/pull/7/files#r23349532>.
>



-- 

* * * * * * * * http://www.mssl.ucl.ac.uk/~npmk/ * * * *
Dr. N.P.M. Kuin      (n.kuin@ucl.ac.uk)
phone +44-(0)1483 (prefix) -204927 (work)
mobile +44(0)7806985366  skype ID: npkuin
Mullard Space Science Laboratory  – University College London  –
Holmbury St Mary – Dorking – Surrey RH5 6NT–  U.K.

APE6.rst
taldcroft: I would leave that for later.

APE6.rst
taldcroft: @PaulKuin - do you want to read QDP files to be readable with astropy?  That's a different question. In that case the comment character is already configurable via the `basic` reader, but right now we don't support trailing line comments.  It would be quite straightforward though to make a new `qdp` reader that works with qdp files.

PaulKuin: To be frank, yes, reading/writing qdp files with astropy would be useful.
The rationale is that  I exchange these files with people using QDP, while
I prefer to process in python.  I think you are right, this might be better
done as a qdp reader.

I did not realise you were talking about training comments in the data
file. I have had trouble with them if not done well. In my previous comment
I was talking about commenting out a whole record, which I have found
useful.  Just want to make that clear.

On Thu, Jan 22, 2015 at 1:15 PM, Tom Aldcroft <notifications@github.com>
wrote:

> In APE6.rst
> <https://github.com/astropy/astropy-APEs/pull/7#discussion_r23373958>:
>
> > +``'description'``, and ``'meta'``.  This is not a ECSV requirement but is
> > +recommended for human accessibility.
> > +
> > +Header details
> > +^^^^^^^^^^^^^^^^
> > +
> > +An ECSV file begins with the the table header which contains the
> > +necessary information to define the table columns and metadata.  This
> > +is expressed as a YAML-encoded data structure which has a small set of
> > +required keywords and standard specifiers.
> > +
> > +Each line of the YAML-encoded data structure must start with the two
> > +characters ``# `` (hash followed by space) to indicate the presence of
> > +header content.  The first line which does not start with ``#``
> > +signifies the end of the header.  Subsequent lines starting with ``#``
> > +are treated as file comment lines.
>
> @PaulKuin <https://github.com/PaulKuin> - do you want to read QDP files
> to be readable with astropy? That's a different question. In that case the
> comment character is already configurable via the basic reader, but right
> now we don't support trailing line comments. It would be quite
> straightforward though to make a new qdp reader that works with qdp files.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/astropy/astropy-APEs/pull/7/files#r23373958>.
>



-- 

* * * * * * * * http://www.mssl.ucl.ac.uk/~npmk/ * * * *
Dr. N.P.M. Kuin      (n.kuin@ucl.ac.uk)
phone +44-(0)1483 (prefix) -204927 (work)
mobile +44(0)7806985366  skype ID: npkuin
Mullard Space Science Laboratory  – University College London  –
Holmbury St Mary – Dorking – Surrey RH5 6NT–  U.K.

taldcroft:
OK, I think that I've addressed all comments except one.

There is an idea from @PaulKuin to allow a configurable comment character (e.g. `!`), but I think for the initial 0.9 release I would like to stick with the minimal idea of a fixed '#' comment character.  If it turns out in practice that this is a real limitation then we can expand the standard.  This will be easier than restricting it later.

So ... is this ready?  The last commit would be updating the final `Disposition` section.  This PR should be merged before https://github.com/astropy/astropy/pull/2319 so that I can put links to APE-6 in the astropy docs.

taldcroft:
I've pushed commit 1963a2f which updates the Decision rationale and status to reflect the final document.  This is obviously predicated on actual acceptance.

Coordinating committee - is this ready for a decision today?

@astrofrog @eteq @perrygreenfield

eteq:
As of right now, this APE has been accepted.  

@taldcroft - FYI for future APEs: normally the coordination committee member who does the merging fills out the decision rationale and related bookeeping.  I'm not sure it says that anywhere, so it's fine that you did that, but I made some additional changes in the final merged version to flesh out the discussion a bit more.

taldcroft:
@eteq @astrofrog @perrygreenfield - thanks!  

And yes, I would have been quite happy to leave writing the decision rationale to someone else. =>  PR #9.

PR #9, opened 2015-01-26
Add detail to Decision rationale in template

taldcroft:
To make it more clear to APE authors about the final process for the `Decision rationale` section.


eteq:
Good idea - merging!

PR #12, opened 2016-03-24
Timeseries object for Astropy (APE9)

Cadair:
This is still in very early draft.


abigailStev:
Pinging myself so I can follow development and discussion.

ehsteve:
I am very interested in this. Thanks for putting this together @Cadair.

ehsteve:
and thanks to everyone that participate in the discussion at PyAstro16 :P

APE9.rst
```diff
@@ -0,0 +1,111 @@
...
+
+type: Standard Track
+
+status: Discussion
+
+
+Abstract
+--------
+
+The goal of a timeseries object in astropy is to provide a core set of
```
ehsteve: remove the goal of sharing. that's a general astropy goal.

APE9.rst
```diff
@@ -0,0 +1,111 @@
...
+functionality that can be shared between astropy, affiliated packages and the
+wider community.
+
+The scope of this timeseries object is designed to remain focused on the core
```
ehsteve: Reword suggestion, `the scope of this time series object is to provide the capability to support a unit and meta-data aware time series that supports basic analysis tasks relevant to the community. Such an object should also be able to easily export itself into other formats such as pandas series`

APE9.rst
```diff
@@ -0,0 +1,111 @@
...
+functionality:
+
+#. Extending timeseries with extra rows
+#. Concatenating multiple timeseries objects
+#. Sorting
+#. Slicing / selecting time ranges (indexing)
+#. Re-binning and re-sampling timeseries
+#. Interpolating to different time stamps.
+#. Masking
+#. Support for subtraction and addition.
```
ehsteve: maybe add (e.g. background)

APE9.rst
```diff
@@ -0,0 +1,111 @@
...
+
+#. Converting between time systems
+#. Astropy unit support
+#. Support variable width time bins.
+
+It is proposed that this functionality be implemented via a subclass of the
+``QTable`` class, with a few requirements specific to the timeseries class:
+
+#. A 'Time' index column exists which enforces unique indexes.
+#. The table is always sorted in terms of increasing time.
```
ehsteve: if this is true then it does not need to provide sorting capability.

APE9.rst
```diff
@@ -0,0 +1,111 @@
...
+
+
+Binned Data vs Sampled Data
```
ehsteve: I like calling out this distinction!

ehsteve:
I think it would be helpful to list a number of examples of analyses that we would want to perform on these.

APE9.rst
ehsteve: Should we mention performance? It should work well and quickly for large data sets (million entries) as well as small.

APE9.rst
Cadair: it needs to be able to sort itself, the list above is not end-user functionality.

duncanmmacleod:
I am interested in this feature, and think it would be a great addition. [GWpy](//github.com/gwpy/gwpy) would likely move to using/inheriting this object, rather than the current `gwpy.timeseries.TimeSeries` (which inherits from `astropy.units.Quantity` in the first place).
I'm happy to help with development and testing as desired.

abigailStev:
I am in support of this APE! And I'm available to help with testing.

eteq:
I like it.

eaydin:
I love it!

abigailStev:
I would like for @StingraySoftware to move towards using a TimeSeries object (or at least basing our `Lightcurve` class on it) for lightcurves.

APE9.rst
```diff
@@ -0,0 +1,111 @@
...
+There are two different types of data that a timeseries class or classes would
+have to support, sampled data and binned data. Sampled data is taken at a
+precise time stamp whereas binned data is a number of counts contained within a
+some time window. These two different types of timeseries data require different
+handling of methods such as re-binning or re-sampling, as well as for sampled
+data the width of the bins must be stored.
+
+This APE proposes that we place the following restrictions on binned data:
+
+#. Contiguious bins, i.e. the start of the i+1th bin is the end of the ith bin.
```
SolarDrew: I think we decided this was not going to be necessary.

APE9.rst
```diff
@@ -0,0 +1,111 @@
...
+This document proposes an implementation based on Table, with extra constraints
+and specific functionality dedicated to timeseries type data.
+
+Detailed description
+--------------------
+
+Many different areas of astrophysics have to deal with 1D timeseries data, i.e.
+either sampling a continuous variable at fixed times or counting some events
+binned into time windows. These types of applications require some basic
+functionality:
```
SolarDrew: How many of these features already exist in `QTable`?

Cadair: Some, but all would need to be customised for timeseries data (i.e. to preserve the index correctly.)

APE9.rst
```diff
@@ -0,0 +1,142 @@
...
+
+#. Extending timeseries with extra rows
+#. Concatenating multiple timeseries objects
```
Cadair: Re-word to clarify that concatenation is extra columns.

APE9.rst
```diff
@@ -0,0 +1,142 @@
...
+
+To distinguish between binned and sampled timseries we propose having two ways
+of constructing a timeseries, ``BinnedTimeSeries`` and ``SampledTimeSeries``.
+This distinction in the name of the classes will allow a easier to understand
+API. Some examples of constructing these objects is given below.
+
+
+Initialize a ``SampledTimeSeries`` with a time and a data column::
+
+  ts = SampledTimeSeries(time=['2016-03-22T12:30:31', '2016-03-22T12:30:32', '2016-03-22T12:30:40'],
```
Cadair: Should we be able to construct from a start time and a list of "deltas"?

APE9.rst
```diff
@@ -0,0 +1,142 @@
...
+data the width of the bins must be stored.
+
+
+To distinguish between binned and sampled timseries we propose having two ways
+of constructing a timeseries, ``BinnedTimeSeries`` and ``SampledTimeSeries``.
+This distinction in the name of the classes will allow a easier to understand
+API. Some examples of constructing these objects is given below.
+
+
+Initialize a ``SampledTimeSeries`` with a time and a data column::
```
Cadair: This section should also demonstrate constructing these objects with Time and TimeDelta so that it's clear that's a thing.

APE9.rst
```diff
@@ -0,0 +1,142 @@
...
+binned into time windows. These types of applications require some basic
+functionality:
+
+#. Extending timeseries with extra rows
+#. Concatenating multiple timeseries objects
+#. Sorting
+#. Slicing / selecting time ranges (indexing)
+#. Re-binning and re-sampling timeseries
+#. Interpolating to different time stamps.
+#. Masking
```
Cadair: and missing value (NaN) lol

abigailStev:
Important points from our PyAstro18 discussion: 

- Need support for nanosecond time precision from an arbitrary zero point (like TimeDelta), including gps time (e.g., https://github.com/astropy/astropy/issues/6190). 
- Detail point 1 refers to "concatenating" rows or datapoints, detail point 2 refers to "concatenating" columns or more information for existing datapoints. 
- Sorting appears to be minimally useful for what we can think of at the meeting, though shouldn't be impossible. 
- Plotting/dataviz will not be uniquely provided but should be made easily possible via, e.g., matplotlib and shown within the tutorials and examples. Make sure it works with unit/datetime support!
- What about NaNs? Masking should support/be able to handle a NaN value in one column but not another.
- For linearly, evenly sampled data, all we really need to store are the starttime and deltatime. In GW and X-ray timing, there can easily be hundreds of thousands to millions of points for one segment, and 100s of energy bands with the same time binning. Make it possible to never calculate the time column in the first place in this special sub-case? (might not exist in first version of this, but totally necessary for next iteration). This will muck up sorting...
- Always put the start time of the timeseries in the metadata. For linearly evenly finely sampled data, also store deltatime in the metadata.

mhvk:
Led here via #7391 - just a general comment: I feel that this is a bit like spectra or wcsaxes, something we definitely eventually want to support, but for which, to me at least, it makes much more sense to start with an affiliated package to play around with and find out what is really needed.

This feeling is informed partially by having seen `NDData` been initially an abstract structure that was beautiful but hardly used, and then see us end up with a mash of that structure and the more directly useful one developed for ccd imaging analysis. (That said, given the idea to base oneself on `QTable` -- great!! -- maybe the risk here is much more minimal.)

gully:
Hey, chiming in here...
There seem to be synergies with ongoing efforts at the Kepler/K2 Guest Observer office.  We have been developing [lightkurve](http://lightkurve.keplerscience.org/), which has notions of time series, and surrounding functionality:
- TargetPixelFile classes
- LightCurve classes
- LightCurveFile classes

... and more.  The tools have focused on space-based optical, high-precision, high-cadence, monochromatic lightcurves, such as Kepler/K2 and, soon, TESS.

I've envisioned a framework that would combine `lightkurve` with ground-based photometry, like ASASSN.  This APE seems like the right place to kick start a conversation about the topic, so I'll leave it at that for now.

astrofrog:
@gully - the lightkurve package looks great! We're going to try and wrap up this APE and will be started on a prototype implementation, so just wanted to check whether you have any thoughts on the APE as it stands right now. If there was an implementation along the lines of this APE, do you think it would be usable in lightkurve? Would you be interested in contributing to the development?

gully:
Some thoughts---
- As the APE notes, pandas has lots of (but not identical) shared functionality with the proposed effort.  Is it conceivable / possible / desirable to subclass `DataFrame` rather than a `QTable`?  What are the pros/cons?  It seems like the main features are units support and variable bin size, but could those be extended to pandas?  `lightkurve` currently supports a `to_pandas` method, inspired by the astropy equivalent.
- It would be conceivable and probably desirable to retrofit `lightkurve`'s `LightCurve` class to inherit from the proposed `BinnedTimeSeries` object.
- One main "killer app" is a "show me all time-series photometry for this source" tool, e.g. Figure 1 of [Gully-Santiago et al. 2017](http://adsabs.harvard.edu/abs/2017ApJ...836..200G) or the more familiar Tabby's Star heterogeneously-sourced lightcurves.  I envision a tool that queries Gaia (the subset with time series available), ASASSN (requires a Captcha presently), KELT (not presently queryable), Kepler/K2, TESS (soon), AAVSO, and others (previous ASAS, CoRoT, etc.).  Even building this tool with a subset of the most dependably programmatically queryable sources would facilitate lots of science.
- A great feature exists in [SNcosmo](https://sncosmo.readthedocs.io/)-- the ability to synthesize and combine time series photometry across multiple different observation bands.  I suspect this functionality is well beyond the scope of this APE, but such feature requests will inevitably arise when researchers confront the question of "what to do with disparate filters", once we have convenient access to a objects housing them.

gully:
Ping @chedges @barentsen: see above for my thoughts on synergies of this Astropy Proposal for Enhancement and our ongoing lightkurve efforts.

APE9.rst
```diff
@@ -0,0 +1,142 @@
...
+
+status: Discussion
+
+
+Abstract
+--------
+
+The goal of a timeseries object in astropy is to provide a core set of
+functionality for representing data where time is the physical event list. This
+includes both continuous values binned in time or event list type data.
```
dstansby: What does "continuous values binned in time mean"? That the timestamps are a time range instead of an individual time?

APE9.rst
```diff
@@ -0,0 +1,142 @@
...
+functionality:
+
+#. Extending timeseries with extra rows
+#. Concatenating multiple timeseries objects
+#. Sorting
+#. Slicing / selecting time ranges (indexing)
+#. Re-binning and re-sampling timeseries
+#. Interpolating to different time stamps.
+#. Masking
+#. Support for subtraction and addition (e.g. background).
```
dstansby: "Support for subtraction and addition" - is this just the same as being able to easily manipulate underlying data?

APE9.rst
dstansby: Why does the table always have to be sorted by time? Why shouldn't a user be able to sort by another non-index column?

astrofrog:
Oh hai @abigailStev @ehsteve @duncanmmacleod @eteq  @eaydin @mhvk @gully @dstansby :wave:

@bsipocz and I (with some input from @Cadair and @SolarDrew) have started working on an implementation of **astropy.timeseries** based on this APE, and we've reached a point where it would be really useful to have input before continuing. So if you have some time to either read over some documentation, or even try out some of the functionality, please head over to:

 👉 https://github.com/aperiosoftware/astropy/pull/2 👈 

Several of you indicated that you'd be happy with helping with the development, so this is a good time to join in since we can start dividing up tasks now that the main infrastructure is in place.

gully:
ping-ing @barentsen @chedges who maybe interested in similarities with http://docs.lightkurve.org/

astrofrog:
While implementing this in https://github.com/aperiosoftware/astropy/pull/2, I've run into a design question regarding the main time series classes, and would be interesting to hear opinions since this should then be written in this APE. In my view, there are two main ways we can name/organize the main classes (though one could think of variants on these too):

1. Have a ``SampledTimeSeries`` and a ``BinnedTimeSeries`` class, which inherit from a ``BaseTimeSeries`` class (which contains some common infrastructure but isn't intended to be used directly). ``SampledTimeSeries`` would have a ``time`` column, and ``BinnedTimeSeries`` would have a ``start_time`` and ``bin_size`` column.

2. Have a single ``TimeSeries`` class that has an optional column called e.g. ``bin_size`` - if set, the time series is binned, and if not, then it's assumed that the measurements are at exact/discrete times. In this case, the time column would be called ``time`` in all cases.

Option 1 seems more explicit, in particular in the sense that it makes it clearer that the time column refers to the start time of the bin (not e.g. the center time), but I wonder whether it will be more confusing for users to have to think about the two separate classes, and Option 2 seems more user-friendly apart from the potential for confusion about what ``time`` means. What do people think?

duncanmmacleod:
@astrofrog, is there room in this for a third class, which is a `RegularlySampledTimeSeries`, that just has `start_time` and `bin_size` **attributes**, where the `time` column could then be dynamically created as `numpy.arange(size) * sample_size + start_time`, rather than being carried around the whole time? Sorry if I'm just late to the party.

astrofrog:
@duncanmmacleod - I think that could be accommodated with the existing classes, with a special ``Time`` sub-class (so that it's just an optimization without the need for a new API)

duncanmmacleod:
@astrofrog, excellent! I don't understand how that will work, but trust that it will.

abigailStev:
I agree with Duncan! A `RegularlySampledTimeSeries` class is what I would use the most.

astrofrog:
@duncanmmacleod @abigailStev - currently the example implementation I linked to above allows you to do for example:

```
ts = SampledTimeSeries(time='2016-03-22T12:30:31', time_delta=3 * u.s, n_samples=1000)
```

The idea is that currently this just computes all the times, but in future we could make it so that the time column is dynamically calculated for efficiency, but this would just happen behind the scenes 'magically'. So this is what I mean by not actually needing a new class for that use case.

abigailStev:
@astrofrog Ok! Yeah the idea is that since they're regularly sampled (and often quite long for my applications, like, 5e5 time bins in one segment), we don't want to always drag around the time column in addition to the 'counts' or 'rate' column.

astrofrog:
@abigailStev - that makes sense! Just out of curiosity, are your time series actually of binned photon events, and if so, is it important from your point of view to distinguish between the sampled and binned time series? (or would you actually treat them the same?)

mhvk:
@astrofrog - the question of separate or the same class is one of those questions that I fear can only be answered by doing it...  But I *think* two classes makes more sense. If you do go that route, I think the binned time series should indeed have `start_time`, but unlike in your text, I think the `time` column should by default evaluate to the middle of the bin, not the start. That would make it much more useful, e.g., for a radial-velocity study based on fairly long spectroscopic observations.

Agreed that regularly sampled data should be part of this, perhaps even a subclass since it allows some handy speedups. But at some level those are implementation details. The text seems a little short on specific methods, etc., that you'd hope to provide.

dstansby:
I am +1 for option A, ie. having separate classes for binned and sampled data. I think the is much more clear than option B. If in practice bin size is too small to care about, then a user is free to use a `SampledTimeSeries`, but I think it is good to force them to make the choice, which in turn will (maybe) force them to consider if it is the appropriate thing to do for a particular dataset.

barentsen:
My +1 goes to a single `TimeSeries` object for three reasons:
1. A single object is easier for new users; no need to read up about intricate differences between two classes to get started.
2. I can't think of any significant differences between the methods offered by `SampledTimeSeries` and `BinnedTimeSeries`. (But maybe someone else can?)
3. As Tom points out, many sampled time series are effectively binned by the detector across some exposure time.  It seems sensible for every `TimeSeries` object to be explicit about this.

Just my two cents.  (This is an exciting APE, thanks everyone!!)

abigailStev:
@astrofrog Binned time series. I very much agree with @barentsen's point 3 -- this is true for us, and then we often bin up the time series to have larger time bins than the intrinsic time resolution of the data.

mhvk:
I think there is a real difference between a time series of events/counts such as one gets from Fermi and a stack of exposures. They also imply rather different strategies for binning, fourier analysis, etc. Another very different type of time series is what one gets from a bolometer scan (think, south pole telescope; WMAP; ...). 

Generally, it may be good to think about providing a good match to existing time series formats, such as Fermi/Chandra even lists and the typical operations one does on those (hence, cc @taldcroft, @cdeil). 

One particular feature that I would look for in event lists is the ability to easily bin in multiple columns, e.g., to construct an image by binning in X, Y, a spectrum by binning in energy, a phase lightcurve by binning in phase (calculated from the times, of course). The relevant numpy routing is `bincount`, but it would need to be multidimensional. Quite possibly, `gammapy` already does this...

hamogu:
A comment on eventlists raised by @mhvk: In typical X-ray analysis, eventlists and lightcurves are different things. Taking Chandra as an example, ACIS has a read-out time of 3.2 so you get a list of detected photons with a time resolution of 3.2 s, and each photon essentially has a time, an energy and a position. These event lists can then be binned on time to create a lightcurve. In principle, one could do a lightcurve with 3.2 s time bins, but that would (for typical objects) give you many bins with 0 photons, a few with 1, and even less with 2 or more photons. For many applications that's not very useful, e.g. finding flares, looking for periods, so instead we make lightcurves that have e.g. 30 min in each bin. This would then be a sampled time series, but in practice is often used like a binned time series, e.g. people run a Lomb-Scargle periodogram using the mid-point of the bin.

@abigailStev : I know there are other observing modes for brighter objects with more counts, but the vast majority of Chandra or XMM observations uses "full frame" imaging. I'm all for a base class that can be used for everything, I'm just saying that even from the X-ray perspective, lightcurves that are so long that the `time` column needs to be computed on-the-fly to save memory are rare (meaning not required by most users).

hamogu:
My thoughts on @astrofrog original question: I agree with @barentsen and would prefer a single `TimeSeries` class. 
From the practical point, I don't see many functions that treat the two cases separately yet, e.g. if we had a function for `LombScarglePeriodogramForSampledData` and `LombScarglePeriodigramForBinnedData` then that would make sense.(*) In practice, almost all the operations I currently do on lightcurves ignore the difference. In some cases, that's mathematically correct, in others (e.g. fitting a function to the shape of the lightcurve) it's not correct, but for lightcurves with relatively small bins the difference between sampled or binned is much smaller than the statistical uncertainty anyway.

I expect that those people who work in fields where the difference between samples and binned matters already know what they are doing and can set a flag `DoThisToLightcurve(lightcurve, binned=True)` or select the right function when needed. 

*: Actually, I only think it makes sense to differentiate once there are dozens of functions that work differently on binned and sampled lightcurves and I just don't think that that's going to be available outside of very specialized packages - and those packages can subclass the astropy `TimeSeries` into subclasses themselves if they need to.

mhvk:
If one does go for a single class, please be sure that `time` then is the *centre* of the bin, not the start; that will avoid countless silly errors...

barentsen:
Serious question: on the stock markets, do financial analysts use different tools for *transactional data* (photon events) and *historical stock prices* (lightcurves)?

I imagine a case could be made for a separate `PhotonEventList` class, which might offer several specialized methods that are not applicable to a generic `TimeSeries` class, e.g. `extract_image(time_range=..., energy_range=...)` and `extract_lightcurve(energy_range=...)`.

The design of such a class should be driven by real use cases of course, i.e. common Chandra/Fermi/XMM analysis tasks.

astrofrog:
Thanks everyone for your feedback so far! Just a few notes based on an in-person discussion with @Cadair and @SolarDrew today:

* Specialized classes, such as photon event lists: as for NDData, any specialized class should be a sub-class of the base classes rather than trying to fit all functionality into the base classes. For example, a ``PhotonEventList`` (which IMHO is a time series) could have specific methods for binning in space or time (as mentioned by @barentsen and @mhvk)

* We can see arguments for one and for two base classes. On one hand users might find it easier to just use a single class, but there are conveniences (such as properties to get the start/end time of bins or the bin width) that would justify having a separate class for binned time series. So our new proposal is to have a ``TimeSeries`` class which are what we were calling ``SampledTimeSeries`` before, and to then have ``BinnedTimeSeries`` that is a sub-class of ``TimeSeries`` that adds a column for the bin size, and convenience properties for the start/end time of each bin. In this way, any function that works with a ``TimeSeries`` would work with a ``BinnedTimeSeries`` by default and could choose to do something extra because of the binning if they want (but wouldn't require re-writing all functions twice as @hamogu pointed out). A consequence of this is that in binned time series, ``time`` would indeed refer to the center of the bin (alleviating @mhvk's concerns).

* We think it still makes sense to sub-class Table/QTable, rather than wrap them as @cdeil suggests. Time series objects really are tables and all the existing methods on tables should apply. In our preliminary implementation, we haven't really run into anything that needs to be disabled because it isn't applicable.

Let us know if you have any thoughts about this, and in particular the new proposal for the ``TimeSeries`` and ``BinnedTimeSeries`` classes.

hamogu:
@astrofrog A `PhotonEventList` differs from a standard one-band `TimeSeries` in that there could be many entries with the same time (if more than one photon was detected in those 3.2 s), but at different x,y detector coordinates. In a `PhotonEventList` thus `Time` is **not** a unique index and "sorting in time" can be ambiguous. This is getting into High-energy astronomy specifically, but in my view, a `PhotonEventList` is just a table, which can then be binned in time (typically with some selection criteria on energy or x,y position) to produce a `TimeSeries`.

dstansby:
> @astrofrog A `PhotonEventList` differs from a standard one-band `TimeSeries` in that there could be many entries with the same time (if more than one photon was detected in those 3.2 s), but at different x,y detector coordinates. In a `PhotonEventList` thus `Time` is **not** a unique index and "sorting in time" can be ambiguous. This is getting into High-energy astronomy specifically, but in my view, a `PhotonEventList` is just a table, which can then be binned in time (typically with some selection criteria on energy or x,y position) to produce a `TimeSeries`.

More generally, I think you're describing something that is a function of time and one or more other variables (in this case spatial). This is also common in heliospheric physics when we measure the velocity distribution function of particles; measurements are counts as a function of time, speed, and two angular coordinates (`f(t, v, theta, phi)`). Support for some sort of pandas or xarray multi-indexing would be really great to have, but is probably another project altogether unless `Table` already has that feature.

astrofrog:
@hamogu @cdeil @dstansby - thanks for your comments! Just a few notes below:

> In a similar direction to what I think @hamogu said above, I'm also not sure if people need or would want to represent photon event lists as TimeSeries objects.

That's fine - I hadn't appreciated the non-unique time of event lists. I guess what I meant is just that when people are talking about domain-specific functionality here, this could be included in subclasses of ``TimeSeries`` if ``TimeSeries`` is the appropriate class to subclass from.

> The current APE doesn't mention use cases yet that you want to address with these classes. Can you add some info on that? Is it mostly about light curves or are there other things also?

This should be clarified - I think we can be broader than just lightcurves, for instance radial velocity curves or gravitational wave time series. It doesn't mean we can't envisage having subclasses for each of these types though (but the idea here is to start very general)

> The biggest difference in your current implementation is __init__. Looking at the current APE here I see many examples how to create TimeSeries with one time and one data column. Is it possible to create TimeSeries with multiple data columns (like flux and flux_error for example) or if I already have a Table object with a time column?

In the current implementation I'm working on, yes. I agree we should expand on this in the APE.

> I think explaining this or giving an example in the APE would be very useful. If __init__ becomes too complex you could add alternative constructors like from_table, or move the current __init__ to a from_something and keep the main __init__ for creation from Table.

That's a good idea, I'll see what would make sense with the current implementation.

abigailStev:
In terms of event lists: while this is how my data is produced, I use these events (in a `Table`) to populate a light curve with even time sampling then do my analysis with those light curves. So, an `EventList`-type object wouldn't get much use from me (and others in X-ray [spectral-]timing).

@hamogu: I wouldn't discount the entirety of X-ray timing and spectral-timing. Data from RXTE, NICER, and Astrosat LAXPC (and in the future, eXTP and maybe STROBE-X) require the ability to handle long time series, even just for a single segment (like, 32 seconds of data) if it's finely binned.

Using time in the center of the bin (re: @mhvk) works for me as long as this is *very* clearly documented!! (knowing where in the bin your timestamp is has caused many a headache in the past) 

I agree with @astrofrog about having broad use-cases like radial velocity and grav. wave time series! And I agree with @dstansby about having support for multiple dimensions (like spatial or energy information) in one time series segment (i.e., with the same timestamps) -- this would be super duper awesome.

mwcraig:
I'm late to this, but have a couple things to add:

+ I like the `TimeSeries`/`BinnedTimeSeries` option more than the original `Sampled...` and `Binned...`
+ I'd propose (this may be a little extreme) that in the `BinnedTimeSeries` there not be a `time` property at all. Instead, I'd go with the more explicit `start_time`, `end_time` and `mid_time` (or something along those lines). It is a little more verbose but has the virtue of being very clear what time you were using without needing to track down the documentation.

gbelanger:
Good morning, 

I just read through this thread and find the discussion very useful and important. I have implemented, over the past 12 years, a bunch of packages for time domain analysis in Java, and have had to make all the choices that are being discussed above in terms of how to represent time-related data. 

Because my background is high energy astrophysics, it was clear from the very start that I needed to have EventList and TimeSeries as different objects. As it was mentioned above, EvenList can have other characteristics in addition to arrivalTime such as x, y, energy, and a quality flag, for example. In the TimeSeries this is not the case. But there can be fundamentally different kinds of time series, like what I've called CodedMaskTimeSeries that carries a lot more information than just intensity as a function of time, because the time series is actually reconstructed from images, and the errors depend on the angle at which the source was in the field of view, and on and on.

So, I have a package to represent a event lists and a package of time series, and in the time series package, I have an interface ITimeSeries, AbstractTimeSeries, TimeSeries, and CodedMaskTimeSeries going from the most general to the most specific, each class inheriting from the previous while each implementing the interface. I also have an interface ITimeSeriesFileReader with specific file readers that implement it but that apply to different kinds of input files (fits, ascii, qdp), that the TimeSeriesReader just loops through to try each implemented type when given an input file. Similarly, there are several TimeSeriesFileWriters.

The similarity between an EventList and TimeSeries in terms of its temporal information carrying capacity is used in the factory classes that can handle time data: a TimeSeriesMaker and a PeriodogramMaker. These have methods to deal with event lists and with time series as inputs. For the periodograms, it is again with an AbstractPeriodogram that is used as the base, and then different types that extend it (FFTPeriodogram, RayleighPeriodogram LombScarglePeriodogram, ZPeriodogram, ModifiedRayleighPeriodgram, LikelihoodPeriodogram, as well as AggregatePeriodogram to put together many periodograms and AveragePeriodogram to express the result, which is the only periodogram that has uncertainties simply because they don't make sense in any other one since the power a each frequency in a periodogram is a point estimate.

Lastly, I have been recently re-writing the resampling of time series in order to make is more manageable, because dealing with resampling of a binned time series in a general way is very complicated. My time series are allowed to have both irregular sized bins and irregular spaces in between them, and everything is written to handle that. So, I have been working on decomposing the time series into a collection of intensity bins composed on a intensity object and a bin object each of which carries specific information that defines everything we need to know about it, and can thus be handled simply one bin at a time. This is ongoing and I'm still sorting out difficulties like how to deal with different kinds of intensities (absolute quantity and density, for example) in a general manner.

But in conclusion, I think that firstly, having EventList and TimeSeries objects makes sense, and an EventList is not really a kind of TimeSeries even though its arrival times can be used to make one. And secondly, it is very important to look down the line and consider all the kinds of things that will be done with the EventList and TimeSeries so that it doesn't become impossible to do them when we get there.  Please let me know if you want more details about anything address in this comment. I'd be happy to help further, but unfortunately, I haven't yet started coding in python. Maybe it's the time to start ;)

astrofrog:
@gbelanger - thank you for your insight - it's always really helpful to know that people have gone through similar design questions in the past and see how they have solved them. I think it would be really helpful if you take a look at what we have so far (once ready to try out) and see if you have any suggestions (and it would be great if you were interested in contributing!)

astrofrog:
I've now had a chance to digest all the suggestions above and have now updated the prototype implementation. I've also now put the implementation in a standalone package to make it easy to try out. The repository and docs are here:

:octocat: https://github.com/aperiosoftware/astropy-timeseries
📖 https://astropy-timeseries.readthedocs.io/en/latest/

It would be really helpful if people here can take it for a spin to see if you like the current API. The way I've set up this package is such that once we open a PR to astropy core, all commits will be preserved. Therefore, feel free to open pull requests, or issues!

A few updates based on the above discussion:

* I've decided to stick with separate classes for the binned and sampled time series for now, but calling the sampled one ``TimeSeries`` rather than ``SampledTimeSeries``

* ``TimeSeries`` has a ``time`` attribute, while ``BinnedTimeSeries`` instead has ``time_bin_start``, ``time_bin_center``, ``time_bin_end``, ``time_bin_size`` attributes to make sure everything is explicit.

* I've moved some of the functionality (e.g. the downsampling) out of the base class since the implementation that's there may not be appropriate for all cases. In general I think the base time series classes should be as generic as possible, and we or affiliated packages can then implement sub-classes if needed.

* I don't think we need to take a stance on event lists - if people want to use ``TimeSeries`` for event lists, they are free to do so, but some packages may want to implement their own specialized event list class.

StuartLittlefair:
I'm a little late to this party, but having looked at and tried the prototype I have a suggestion which might make the ```TimeSeries``` class more powerful. Unfortunately I think it would mean a quite different implementation.

If I imagine my use cases for a class, it is to handle collections of **data**, which have a matching time attribute. Typically, data has values, uncertainty, and perhaps a mask indicating data quality - just like the ```NDData``` class.

It would also be incredibly powerful if arithmetic on ```TimeSeries``` objects performed arithmetic on matching "columns", whilst handling the uncertainties and masks automatically. Showing, an example, wouldn't it be great if something like this worked...

```python
flux_1 = NDDataArray(data=y_1, uncertainty=StdDevUncertainty(yerr_1), flags=flags_1) 
flux_2 = NDDataArray(data=y_2, uncertainty=StdDevUncertainty(yerr_2), flags=flags_2) 
ts_1 = TimeSeries(time=t, data={'flux': flux_1})
ts_2 = TimeSeries(time=t, data={'flux': flux_2})
rel_ts = ts_1/ts_2
```
...and the ```rel_ts``` object also had a ```flux``` entry which had the correct values, uncertainty and mask? Users could subclass ```NDDataRef``` to get have the flags combine anyway they'd like.

Such behaviour is compatible with the APE as it stands, but can't be achieved if a ```TimeSeries``` is a subclass of ```Table```, since they don't support having ```NDDataArray``` as a column, as far as I'm aware. One would lose the ability to join, stack etc - but I'd argue that the functionality above is more powerful for many users.

Sorry if this suggestion comes after many people have done a lot of hard work; I hadn't realised this issue until I started to play with the implementation. Feel free to point out that I'm wrong!

astrofrog:
@StuartLittlefair - thanks for the suggestion! I agree that using NDData might work well for some use cases. I'd like to make two suggestions:

* In specutils, we made a spectral mix-in class that can be used to make one of the NDData dimensions be a spectral axis. We could do the same for time, i.e. make a mix-in class that provides an easy way to set up one of the axes of an NDData cube as a time axis. Note that this would then also allow one to set up cubes with two spatial, one spectral, and one time axis. To me, this is complementary to the Table approach - in some cases people want to think of time series as being simple tables, and in others people might want to think of it as an n-dimensional cube (where n>=1) where one axis is time.

* There's no reason we can't make ``NDDataArray`` be used as a mix-in column in tables - in fact I'll open an issue about that now!

I think both of these can be done of course, we don't have to choose.

I'll also mention that for NDData, we changed the API a few times because we ended up making the base class try and be too smart, and some functionality is best left to sub-classes. I think we will need to do the same here - that is, functionality like ``rel_ts = ts_1/ts_2`` makes too many assumptions to belong in a base time series class. Having said that, I think this kind of operation would work well in the context of ``NDDataArray`` with a time axis.

I'll try and work on an NDData mix-in class implementation to see how much work it might be and whether it would suit your use case. I'll also see if I can re-use some of your prototype code.

gully:
Should we merge this APE since astropy.timeseries is merged?

bsipocz:
@gully - I would rather wait until 4.0, and ask everyone (especially folks with non single band optical photometry use cases) here to give the current version a good try and report back issues or missing functionalities.

gully:
Thank you @bsipocz , indeed learning from the existing timeseries implementation should feed back to the long term vision.

bsipocz:
(well, it's not my call to anyway to merge this, but the feedback I heard so far is that the current api is very much kepler type light curve centric. but it's very very difficult to anything about it without actual issues being opened...)

gully:
Interesting, I should try out the timeseries class more to see what feedback I can give. I am curious about multichannel and sparse applications too, despite being Kepler-centric much of the time.  I'm especially interested in use cases that combine sparse ground-based observations with high cadence space based observations.  By "combine", I don't necessarily mean in the same "flat" timeseries object, though that's conceivable.  I just mean making sparse and high cadence interoperate more easily would unlock some good science.  The different bandpasses comes up a lot too.

I re-read the APE governance protocol on the README and appreciate that a lot goes into formalizing the APE, so my use of "merge" wasn't the right choice.  I had thought an APE was a pre-requisite to making new functionality, but I see now that making an experimental implementation further informs use cases and guides the functionality in the longer term.

bsipocz:
@gully - make sure you're playing with the dev version as there are some substantial improvements in the underlying classes that were necessary to be able to extend the functionalies.

Jerry-Ma:
Hi,

I am working on the TolTEC project and I am wring python code for handling the time stream data collected from an array of bolometers that scans in time.

As already pointed out in the previous comments, since each time sample contains tightly packed data (in our case, a vector of size ~4000, matching the total number of detectors in the array), it would make most sense to have a NDData as the storage and have timeaxis mixin attached to one of the axis.

I saw that #16 is still open and I was wondering what is the status of that?

pllim:
So... astropy/astropy#8540 was merged and released in v3.2. Where are we going with this in 2022?

astrofrog:
We should have a "Superseded by events" status for APEs 😅 (jk!)

gully:
So is this APE complete now that `timeseries` exists?  I guess a repeat of the question above.  Would folks say that the aspirations of this APE followed through in the implementation?

pllim:
Would have to ping @astrofrog and @Cadair again

APE9.rst
```diff
@@ -0,0 +1,150 @@
...
+
+author: Stuart Mumford
+
+date-created: 2016 March 24
+
+date-last-revised: 2017 November 10
+
+type: Standard Track
+
+status: Discussion
```
hamogu, line 12:
```suggestion
status: Accepted
```

APE9.rst
```diff
@@ -0,0 +1,150 @@
+A General Timeseries Class for Astropy
+-------------------------------------
+
+author: Stuart Mumford
+
+date-created: 2016 March 24
+
+date-last-revised: 2017 November 10
```
hamogu, line 8:
```suggestion
date-last-revised: 2023 Oct 20
```

APE9.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Alternatives
+------------
+
+Forego the functionality provided by Time and Units and recommend everyone use pandas.
+
+
+Decision rationale
+------------------
+
```
hamogu, line 150:
```suggestion
This APE has long been implemented in https://github.com/astropy/astropy/pull/8540 and somehow the Coco just never merged it.
```

hamogu:
@eteq Please follow through with the Zenodo steps when they are online again (they had trouble minting DOIs the last few days).

PR #13, opened 2016-07-26
Added APE describing roadmap for ending Python 2 support (APE10)

astrofrog:
The purpose of this APE is to set out a roadmap for future releases of the Astropy core package, in particular to identify the process for switching to releases that support only Python 3 and ending support for Python 2.7 in December 2019.

If you would like to leave comments on this, please aim to do so in the next couple of weeks, and by August 19th at the very latest.

**Rendered version:** https://github.com/astrofrog/astropy-APEs/blob/ape-python-3/APE10.rst

*Note: We can settle on whether this is APE 9 or APE 10 once it is ready to merge*

PR #14, opened 2016-07-26
Remaking NDData to support LSST needs (APE11)

perrygreenfield:
This is a first draft of an APE to support the issues raised by LSST and the subsequent email discussions regarding that


MSeifert04:
I'm not sure where the discussion should be.

First: I agree on many points of your APE but I don't really like the proposed implementation.

Let me summarize the current standard:

- `NDDataBase` - Interface definition of "common" attributes
- `NDData` - Container based on `NDDataBase`, the restrictions here are: `data`: numpy-array-like, `meta`: dictionary-like, `uncertainty`: NDUncertainty-like
- `NDDataRef` - `NDData` with specified slicing, arithmetics and io-interface.

So if reworking `NDData` (or `NDArr`) to some "working" implementation would promote the class from a simple container. That's not bad but worth mentioning because it does not only break backwards compatibility (possibly) but also alter the basic data structure.

Additionally the current `NDData` interface would be totally (or mostly) compatible with your suggestions except for the two-layered structure:

- `bitmasks` as `mask`: No problem there. It wouldn't work with "NDArithmeticMixin" (by default) but the current implementation would allow with a minimum of tweaking, it may be as easy as alter one default parameter ("handle_mask"). for "NDData" the mask is designed as property but there is no reason why a property shouldn't return a callable. So there is currently no problem with just using your "mask" class. Besides this there is a flags attribute in NDDataArray which could be reinterpreted as the bitmask without seriously altering mask at all.
- `offset`: Slicing is implemented in "NDSlicingMixin". It's to some extend customizable and adding a "offset" wouldn't be any problem. It doesn't support all possible numpy indexing ways because the `astropy.WCS` doesn't support them. But it also is highly customizable so there is actually no problem instead of slicing the wcs to alter some property (like the proposed offset).
- `variance`: I actually have a working implementation of Variance-uncertainty based on `NDUncertainty` which supports the proposed operations (+, -, /, *). So it's no problem to implement something like that within the current framework.
- ExposureInfo: I think this would be far too specific for `NDData`. You assume that the `NDData` is used for exposures is very strict in my opinion.
- Two-layer-class-structure: Could you clarify _what_ would be seperated, especially _when_ and _why_ this would be needed or not achievable with the current implementation?

Sorry this was so negative. I think a lot of your APE makes sense, for example the bitmask, exposureinfo and variance uncertainty would be really great additions. But I somehow disagree on the proposed implementation. I think these could also be integrated without changing (much) of `NDData` itself. Altough I have to admit that I'm probably biased on these points because I did change a lot in NDData :smile: 

Just to cross-link it here: @crawfordsm already mentioned some or even all of these points in https://groups.google.com/forum/#!topic/astropy-dev/I_co2botSNw

perrygreenfield:
@MSeifert04, I should say that the primary goal of this effort was to see if we couldn't align NDData better with LSST needs and it was deemed as an important goal by the coordinators, so I gave it a shot. I'll try to address some of the comments, but really, it is probably better if someone representing LSST (e.g., @r-owen) responds to the justification for these since I don't really feel I should speak for them in the end.

First, the main goal is not to break backward compatibility, at least for normal uses. This may not  be true for those building subclasses that depend on deeper aspects of NDData and associated classes. Are you in that category? 

Some have questioned the need for the dualism of NDDataBase and NDData (is there any existing compelling use of this split yet?).

**The need for NDArr:** 

I'm guessing the main justification LSST has for this is for an object for which mathematical operations have unambiguous results. This isn't true for your proposed solution of using a subclass since meta or other attributes do not have well accepted semantics in this case. Yes, they could use these with conventions (ignore these attributes, or not set them) but then you have mixed usages of these objects that isn't clear from the class alone. Having subclasses that disable attributes or capabilities in the parent class is generally not a good idea. The proposed NDArr class removes these sorts of problems. It seems much cleaner to me than multiple flavors of NDData. Imho, when there are many different dimensions to how a class will be subclassed, e.g. through mixins of different kinds or other means), one is probably asking for a lot of trouble and confusion down the road. Reducing the number of such variants is a good thing if it can be done flexibly.

**Regarding masks:**

I don't think a property that returns a boolean in some cases and a callable in others is a good solution.

**Regarding offsets**

Yet one more mixin. My previous concern applies

**Regarding ExposureInfo:**

I wasn't clear enough. I don't think such an attribute is justified. It is something that LSST asked for, but I feel that it could be handled on their part by a special class or subclass if they need it in their software with a simple function to extract the needed info from an NDData instance (or other constructors if they don't need an NDArr instance).

Again, I welcome LSST comments and any others that have a better solution to their needs than this APE currently provides.

In the end, it is up to the coordinators to decide whether to accept this or some other variant.

Cheers, Perry

mhvk:
I think it may be most useful to split the discussion along with the two types: a simple data array with mask and uncertainty on which mathematical operations are possible (`NDArr`) is extremely useful and we should support it. The larger one is something that maybe astropy could provide a good base for, but some aspects could also be added by a specific LSST subclass.

For `NDArr`, my guess is that much of it could be made to work from a combination of the various mixin's that we have. It would be lovely to have some specific tests written that would check whether it works correctly, so we can try.

One specific comment: I don't quite see why `NDArr` would not have a `wcs` -- it just has to be one that is limited to carry slicing information. (Or, in other words, it would seem to me that the slicing information should be kept in a wcs-like attribute -- no need to reinvent wheels.)

perrygreenfield:
@mhvk, I don't think WCS as we usually understand it should be part of it since having it makes operations much more fragile or presents many options, none of which are the obvious default. If two NDArr objects with different WCS are combined, what is to be done? Generally resampling is required and that is a complex operation with many issues to decide. 

But it is true the slice aspect with regard to the parent is a kind of a wcs, but it isn't any sort of absolute coordinate system. It is much simpler to handle though. It is something to discuss as to whether that information should affect operations on these arrays.

mhvk:
@perrygreenfield - yes, I agree. It was more that I felt that even an offset in an original frame is best done via a wcs.  But this may just be an implementation detail.

r-owen:
Here is the way LSST views things and what we are trying to emulate with `NDArr`: `offset` is a pair of ints that gives the position of the lower left corner of an image or subimage with respect to the parent image. It is 0,0 for the parent image (or a subimage whose lower left corner matches the parent). Pixel position = index on image + offset and is used almost everywhere that a pixel-based position is wanted, e.g. for pixel position of detected sources and to translate pixel to sky via a WCS.

We are keen to avoid having a WCS in `NDArr`, because as @perrygreenfield says, it makes simple operations such as addition ambiguous (are the arrays aligned via the WCS or via pixels?). Also, `NDArr` is nicely self-contained; many times a WCS is not required (and may not even be known).

For uniformity and expandability, I suggest that WCS stored as part of `meta`, e.g. `meta.wcs`. Other attributes that may prove useful (but should not be required) include exposure date and time, filter name and a PSF model. Having suggested names and implementations for these may provide useful uniformity, even while recognizing that not all will be present for all images.

LSST finds that the metadata is often useful without the pixels, so a single object containing all of it (including the WCS) is a big win. Some use cases:
- Identify reference objects that overlap an image
- Identify images taken with a particular filter and with good seeing that overlap each other, for coaddition.

As @mhvk notes, the offset of the `NDArr`, and also the dimensions (which LSST combines into one object called a bounding box) is needed in order to perform many or most operations with the metadata. I recommend making bounding box a required part of the metadata, and having the slicing operator update it as required (e.g. when obtaining a metadata for a subregion). The bounding box is the only item that requires this treatment. WCS can simply be copied when taking a subregion (since pixel position is defined with respect to the parent image).

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+becomes much more straightforward, even with masks. Propagation of uncertainty 
+is included, though propagation of uncertainty is certainly not universally straightforward. For both masks and uncertainty, it will be possible to 
+create the NDArr object with options not to propagate one or the other or 
+both to forestall complaints from those for which these operations will be 
+considered invalid.
+
+NDArr will support optional units since the propagation of units is
+unambiguous in mathematical operations.
+
+Supported numerical operations for NDData are: (+,-,*,/). While the discussion
```
MSeifert04: Will NDData (or rather NDArr) really support these numerical operations?

perrygreenfield: Ah, that's a typo. It was missed in editing. NDData here should be NDArr.

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+
+The mask attribute now will be an instance of a Mask Object (but to support
+backward compatibilty, the attribute must have a different name than mask since
+the current implementation treats that attribute as a boolean array (or any
+object that mimics one). [An alternative option is to assume that the mask
+attribute of NDArr is a callable whereas the mask attribute of NDData represents
+a boolean array; this would be workable, but also likely confusing]
+This APE proposes that the new attribute have the name: lone_ranger. 
+Well, perhaps dmask.
+
```
crawfordsm: Is there a good reason not to use .flag here as that is already available in NDData as the name?  

(Although I enjoyed the lone ranger joke)

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+3. For all other masks the result is a boolean array combination of the 
+default callable mask for each of the two inputs. In the case of a bit plane 
+combined with any other type of mask, the boolean will be applied to each 
+bitplane.
+
+Uncertainty (the eternal tar pit)
++++++++++++++++++++++++++++++++++
+
+The NDArr object shall have a "variance" attribute (too bad for those that
+desire standard deviation as an option; though for backward compatibility
```
crawfordsm: So this section switches back to talking about NDArr and not NDData.  Are there going to be restrictions on NDData's uncertainty?   Will it only be allowed to be variance?  A previous restriction on uncertainty that has  been removed had already prevented people from using NDData as users want more complex uncertainties than just variance or standard deviation.

crawfordsm:
Thanks @perrygreenfield and @r-owen.   I've made a few comments in line but I've also made a pull request against @perrygreenfield with an alternative solution.  

The reason for the alternative solution is that I really don't like the idea of NDData being required to use NDArr.   I think that limits the flexibility of NDData, which was one of its features, and if it isn't required to use NDArr, then it increases the ambiguity, which makes compatibility more difficult.   

The bounding box/origin just seems like a very restricted case of WCS (where the only transformation is a translation), but I can see where having a shortcut like that could simplify the coding and be very useful so I could see that being included immediately into NDData regardless of this APE.

APE9.rst
perrygreenfield: @crawfordsm I'm not sure what the status of the .flag attribute is. It's not currently part of the existing NDData class definition, is it? I know it has been discussed in the past, but in searching through documentation, emails and code, I don't see any actual implementation. Am I missing something?

APE9.rst
perrygreenfield: @crawfordsm As an internal representation, it would only be variance. But nothing prevents retaining the current stddev interpretation as a property. Given that most computations using it generally use variance, it does make sense to make that the internal representation. Trying to support both internally to me just adds unnecessary  complexity. As for more complex uncertainty models go, I'd argue that these are probably best handled outside NDData, again to keep it from becoming a burden to figure out what kind is actually present.

perrygreenfield:
@crawfordsm could you provide an example or two where an NDData implementation using NDArr reduces its flexibility. I'm not sure I see the point being made.

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+Mask Object
++++++++++++
+
+The mask attribute now will be an instance of a Mask Object (but to support
+backward compatibilty, the attribute must have a different name than mask since
+the current implementation treats that attribute as a boolean array (or any
+object that mimics one). [An alternative option is to assume that the mask
+attribute of NDArr is a callable whereas the mask attribute of NDData represents
+a boolean array; this would be workable, but also likely confusing]
+This APE proposes that the new attribute have the name: lone_ranger.
```
mwcraig: 😆

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
++++++++++++
+
+The mask attribute now will be an instance of a Mask Object (but to support
+backward compatibilty, the attribute must have a different name than mask since
+the current implementation treats that attribute as a boolean array (or any
+object that mimics one). [An alternative option is to assume that the mask
+attribute of NDArr is a callable whereas the mask attribute of NDData represents
+a boolean array; this would be workable, but also likely confusing]
+This APE proposes that the new attribute have the name: lone_ranger. 
+Well, perhaps dmask.
```
mwcraig: Something a little longer/more descriptive might be more helpful, maybe `data_quality`?

crawfordsm:
The problem that arose previously was that if NDData *has* to have a specific type of `uncertainty`.   That reduces the flexibility and it is not clear in your APE whether or not that is required as it would look that for NDArr variance is required and if NDData requires a valid NDArr that would imply that it required a variance.

APE9.rst
crawfordsm: Inverse variance is far more common in spectroscopic pipelines, and meanwhile other people want to use probability functions or other description of the uncertainty.  The goal of NDData is to describe a base object that can be shared between different groups and different groups want to use different uncertainties.   This is why we didn't fix it to one type in NDData.

APE9.rst
MSeifert04: I don't know if it's used - but the flags are implemented in [`FlagCollection`](https://github.com/astropy/astropy/blob/master/astropy/nddata/flag_collection.py) and are possibly usable in [`NDDataArray`](https://github.com/astropy/astropy/blob/master/astropy/nddata/compat.py) and [`ccdproc.CCDData`](https://github.com/astropy/ccdproc/blob/master/ccdproc/ccddata.py).

crawfordsm: @perrygreenfield No, although I thought there was some mention of it as `FlagCollection` is provided to support it and it does show up in `CCDData`, but I don't think there is a fixed definition of it.   This is just what I remember as being one of its use  cases.

eteq:
Quick note: the ordering of APEs has traditionally been based on pull request number.  So at some point this will need to be rebased to rename it from "APE9" to "APE11".  (#12 is APE9 and #13 is APE10)

APE9.rst
mwcraig: IIRC, `FlagCollection` was, conceptually, supposed to be able to hold things like data quality masks, but its original implementation was minimal. In the [last NDData APE](https://github.com/astropy/astropy-APEs/pull/8) not a lot was said about flags or `FlagCollection`. When the code was re-written for that APE I left the `FlagCollection` code intact largely because it seemed to be doing no harm.

What is, I think, clearly missing from the NDData structure now is a consistent way to handle data quality bit planes. This comes up both for LSST and for HST (see astropy/ccdproc#389 by @SaraOgaz). 

I think the addition of that ability somewhere in the class hierarchy is useful.

jehturner:
Sorry, I started replying to the prior email thread during my vacation in July, but then this APE got circulated before I finished my message and I'm just getting back to it after catching up with other things :-(.

The basic requirements here sound completely fine to me, with a few nice new features -- except that I still disagree with NDData (or a sub-class) not supporting mathematical operations. While I don't doubt there is valuable experience behind the objection to arithmetic, I'm thinking in particular of my own usage here; I would like to work with high-level objects describing exposures and all their logically-associated information while retaining the major convenience of being able to operate on them concisely, in a clear and ergonomic sequence. I really don't want bookeeping or conceptual operations that take several steps to perform in my scientific process. These are things I want to manipulate hands-on. If the default behaviour for arithmetic doesn't make sense, it can be altered in a similar way to your dmask.apply_flags() example or the attribute to disable variance propagation. Otherwise I am just going to end up extending it to work anyway (in fact I will have to if I re-base DataFile on it).

I still haven't thought of a great alternative name, but NDArr seems like it could easily be confused with NDArray. What do people think of something like NDPixData? Then again, people might get used to either.

I'm not clear on the memory implications of all these (strong?) references to parent objects, though this is something LSST must have considered carefully.  Obviously in the ideal case, if one were to slice a data array, drop the variance/mask somehow and delete the original reference(s) to the parent, then only the main subarray would survive (might be awkward to implement without shooting oneself in the foot?).

By the way, Gemini has also been (with a bit of encouragement) re-basing a version of the AstroData class used in our pipeline on NDData.

I'll make a couple more comments in-line then have a better look over the discussion...

Thanks (& sorry for being a bit flaky),

James.

jehturner:
BTW: +1 to bit planes (I think we could probably deprecate flags?) and variance.

APE9.rst
jehturner: Flags is still used in NDDataArray. I am using it (instead of mask) to support bit planes (DQ) in ndmapper, but in the design described here I would switch to using the new mask object instead; I'm not aware of any good reason to keep flags in that case, other than general compatibility.

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+to corresponding information in the astropy object. The question is: does
+it make sense to encapsulate that information in a single attribute, and 
+whether all think that these attributes are "astropy-universal”. For the 
+purposes of this, the only standard attribute is most likely the WCS.
+
+Items such as exposure time and filters could be relegated to the meta
+attribute. Making attributes such as these at the top level could be done
+through subclasses, though it may lead to many variants. Perhaps the best way
+to deal with this is for applications or libraries to state their requirements
+for items required to be in meta and leave it at that.
```
jehturner: I would have thought it will be useful to have more functionality than just a meta dictionary and WCS. While it's not at all practical to standardize everything comprehensively, some sort of fairly general meta-data abstraction (like Gemini has in our mostly-internal AstroData class) would help write data reduction routines that can be used with data from different instruments etc. (or at least avoid hard-wiring more assumptions than are necessary). I thought we also came up with some useful attributes to parse into objects besides WCS at the meeting but will have to dig out the photo. One thing I wonder about, off hand, is tracking associations between NDData objects besides parent-subarray (eg. calibrations that go with a science exposure). By the way, is the implied alternative to an ExposureInfo object here just to keep meta in NDData?

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+NDData Slicing and Related
+++++++++++++++++++++++++++
+
+Slicing of NDData is supported (eventually) as well as striding. The corresponding 
+WCS attribute, if present, is updated accordingly. Eventually there may be support
+for slicing based on world coordinates, but this will not be part of the initial
+implementation.
+
+NDData does not support mathematical operations. Utilities will be provided to aid
+in how metadata items are to be propagated when the user merges NDData instances.
```
jehturner: Having utilities to aid in merging meta-data actually lessens the argument for not supporting arithmetic, doesn't it (that's one of the most frequently-cited problems in this context, though I've never found it to be that big a deal in practice)?

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+
+Mask Object
++++++++++++
+
+The mask attribute now will be an instance of a Mask Object (but to support
+backward compatibilty, the attribute must have a different name than mask since
+the current implementation treats that attribute as a boolean array (or any
+object that mimics one). [An alternative option is to assume that the mask
```
jehturner: Some people (Erik T?) state that the current implementation does support bit planes and it is only required to treat them as Boolean in the sense of good=0, bad!=0 (NumPy convention), but there seems to be some ongoing confusion about this.

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+The returned boolean array must be broadcastable to the underlying data array
+(a scalar boolean is permitted).
+
+As an example for bitplane masks:
+
+- ``nda.dmask()`` returns a boolean array
+
+- ``nda.dmask.apply_flags(…)`` sets how the bit planes are to be used to generate a boolean result when ``ndd.dmask()`` is used
+
+- ``nda.dmask(select=nda.dmask.bitplanes['saturated'] + nda.dmask.bitplanes['dead_pixels'])``
```
jehturner: Had you considered something like this, which might be more backwards compatible?
-  `nda.mask` a boolean-array-like object (numpy sub-class?)
- `nda.mask.select(nda.dmask.bitplanes['saturated'] + nda.dmask.bitplanes['dead_pixels'])`  does the same as your 3rd example
I think Perry asked previously for an example of how to implement this and it doesn't seem obvious without sub-classing numpy since we don't know how everyone is currently using mask.

Also, these labels are a good thing (along the lines of the meta-data abstraction I discussed previously) but can a bitmask also be allowed as an alternative? Using a single number is convenient, eg. for users, when dealing with some fixed convention.

APE9.rst
jehturner: I think someone at Gemini has already drafted a VarianceUncertainty class, for what it's worth.

APE9.rst
```diff
@@ -0,0 +1,273 @@
...
+to preclude supporting the older interface to NDData. In effect, all the old
+attributes and methods will continue to work if used in straighforward ways
+(there are always ways to break something with changes of this size; for example
+the new attributes may conflict with existing code that extends NDData, though
+given the level of use of NDData, this is highly unlikely). This backwards 
+compatibility layer may result in some computed items that previously were not
+computed (e.g., computing stddev from variance).
+
+Whether the backward compatibility is temporary or permanent is not yet 
+determined.
```
jehturner: I would perhaps decide this after the fact, depending on demand?

APE9.rst
MSeifert04: I also have a `VarianceUncertainty` class. :smile:

mhvk: And there is a PR with a uncertainty propagating, covariance preserving `Uncertainty` class at https://github.com/astropy/astropy/pull/3715

Side note: I don't understand the preference for `Variance` over standard deviation. Calculations hardly speed up for any operation where it matters, yet one looses the immediacy of the error and value having the same unit and scale.

jehturner: OK, we should probably use that (it's not in astropy dev yet, correct?) since you've already done a lot of work on the arithmetic. There was some talk here of submitting a PR for ours, but I think it needed re-basing first.

jehturner: @mhvk At this point my preference is mostly based on our long-established current practice, rather than the underlying arguments -- not that it should prevent doing whatever is optimal -- I'm mainly saying that variance is fine for me and Gemini, as an interested party.

mhvk: I'm fine with that; just don't quite see why the PEP has to be so exclusive in how it is stated.

MSeifert04: @jehturner I don't know if it's still compatible with the `astropy`-state. :smile: I wanted to do some additional changes to `NDData` for 1.3 but there always seemed something else was more important. If you want I do have a working implementation in a git-repository.

jehturner:
The changes proposed by @crawfordsm are here: https://github.com/perrygreenfield/astropy-APEs/pull/1/files.

Given that there seems to be more consensus on what `NDArr` (or whatever it's called) should look like than `NDData` and `ExposureInfo`, I was also wondering independently whether it would make sense to standardize on that and have alternative higher-level objects (`LSSTData` vs `NDData`) that both have a beak and webbed feet (eg. an `.arr` attribute) but without the former having to flap against its will. Steve goes a bit further than me, however, by proposing that `NDData` shouldn't necessarily have `.arr`.

APE9.rst
jehturner: While not longer or more self-explanatory, ".dq" would be concise and I think is an acronym understood by many people? That's what we've always called it anyway. Just an idea; I suppose data_quality isn't terribly long either and more the Python way.

perrygreenfield: I've been away a while, but I'm back... I would be happy with any sensible consensus (e.g., if the consensus was .dq_aka_data_quality_mask I don't think I would agree :-)

APE9.rst
perrygreenfield: @mhvk the main point is that the internal representation should be consistent. I think the worst situation is to support both. We should pick one. I could live with either, and there are good arguments for one or the other. One possibility is to make one a property that references the other. But is sure would simplify all the internal code to know that it only has to deal with one case when performing operations on this object.

APE9.rst
perrygreenfield: @jehturner Yes, that's my worry. I can see all sorts of unexpected issues arising from such an interface. So I do an arithmetic operation on two NDArr instances that both have bitplanes stored in such an attribute (.mask) what is the result? Or are you saying that the important stuff is in .dmask and that .mask is a property that shows a simple view of .dmask (or did you mean .mask instead of .dmask?). I'm not sure how the above is supposed to work in these cases.

APE9.rst
perrygreenfield: @jehturner I'm not sure how that is the case if the utilities have multiple options for merging (and generally these are needed). As soon as you have more than one way to do it, then simple expressions don't have a way of indicating which way it should be done. For example, for the metadata, do I choose one or try to merge them (and how specifically do I merge them; by averaging, concatenation, by date, by order, etc?), do I add history about the merging, and so forth.

APE9.rst
perrygreenfield: @jehturner The devil's in the details. If there is a broad consensus for extra attributes and their meaning, fine. But one could quickly get into a VO-like morass before long. I'd suggest putting that off and settling the other issues first.

perrygreenfield:
> 
> The basic requirements here sound completely fine to me, with a few nice new features -- except that I still disagree with NDData (or a sub-class) not supporting mathematical operations. While I don't doubt there is valuable experience behind the objection to arithmetic, I'm thinking in particular of my own usage here; I would like to work with high-level objects describing exposures and all their logically-associated information while retaining the major convenience of being able to operate on them concisely, in a clear and ergonomic sequence. I really don't want bookeeping or conceptual operations that take several steps to perform in my scientific process. These are things I want to manipulate hands-on. If the default behaviour for arithmetic doesn't make sense, it can be altered in a similar way to your dmask.apply_flags() example or the attribute to disable variance propagation. Otherwise I am just going to end up extending it to work anyway (in fact I will have to if I re-base DataFile on it).
> 
Nothing prevents this being done in a subclass. I think the main issue was whether there would be an object for which such operations intrinsically had no ambiguities. That’s not true for an NDData subclass unless you force it to be like the proposed NDArr class. Taking away attributes in a subclass is generally a bad thing.

> I still haven't thought of a great alternative name, but NDArr seems like it could easily be confused with NDArray. What do people think of something like NDPixData? Then again, people might get used to either.
> 
Again, I’m happy with any reasonable consensus. 

> I'm not clear on the memory implications of all these (strong?) references to parent objects, though this is something LSST must have considered carefully. Obviously in the ideal case, if one were to slice a data array, drop the variance/mask somehow and delete the original reference(s) to the parent, then only the main subarray would survive (might be awkward to implement without shooting oneself in the foot?).
> 
Not sure what is meant by this. Slicing a data array doesn’t do away with the original array buffer, even if no reference to the original numpy array exists.

jehturner:
I'm just wondering where things stand with this (having been focused on finishing some unrelated work)? Has LSST given up on it? There were obviously some differences of viewpoint, but I would have thought they could be accommodated somehow, with a reasonable degree of interoperability. I think having some idea of the plans for NDData might be useful context in next week's spectroscopy meeting. Thanks!

perrygreenfield:
I’d like to work on it, but haven’t had time. 

Perry

jehturner:
That makes two of us then... It would be interesting to confirm where LSST stands though. The grapevine suggests they are currently not planning to use NDData. Some of these changes may be worth making whether or not that's true, but obviously the constraints and available effort would differ a bit in the two cases.

parejkoj:
@r-owen , Simon K. and I just chatted about this. Sorry we also dropped the ball here. LSST is definitely interested in making progress on it.

I think what we should try to do is get a new design (incorporating comments from above) from astropy, and then we can provide our preferred diff to that and discuss it on a telecon.

mwcraig:
Perry and I will both be at pyastro17...next week I'll propose some times that would fit well with the conference schedule and not be too crazy for remote participation from the US. The goal of that would be to come to consensus from the astropy side on what NDData would look like. When I create the doodle poll I'll send it out to astropy-dev; please feel free to forward on to anyone you might think would be interested.

mwcraig:
First in a series of comments: Summary of current features in nddata/LSST:

| Feature | `NDData` | `NDDataArray` | LSST `Exposure` | LSST `MaskedImage` | `NDArr` (proposed) |
| ------- | -------- | ------------- | --------------- | ------------------ | ------- |
| Sliceable |   No   |    Yes        |     Yes, [but](https://community.lsst.org/t/indexing-afw-image-objects/1872/1) (does not affect the meta)          |      Yes           |  Yes    |
| Has origin |   No  |   No          |      No         |       Yes          |  Yes    |
| Operations? | None | methods like `.add` | None      |   in-place `+=`, etc  |  binary `+`, etc |
| Has metadata (i.e. non-mask, uncertainty, origin) | Yes | Yes | Yes (as `ExposureInfo` attribute) | No | No |

mwcraig:
Proposal for evolution `NDData`, *assuming it will remain a container class*. In other words, it is NOT really a direct equivalent to LSST's `Exposure`. Essentially, classes that implement this interface will be understood by the `@support_nddata` decorator.

## Attributes/methods of `NDData` object (i.e. high-level container)
### Most properties would be read-only (uncertainty has a setter for compatitibility with previous versions)

|       | **astropy v1** | **astropy v2** | **astropy v3** |
| :---: | :-----------:  | :------------: | :------------: |
| Pixel data (slicable) |  `data`    | `data`   | `data`   |
| Binary mask (numpy convention) | `mask` (property) | `mask` (property, deprecated as part of NDData) and `binary_mask()` | `binary_mask()` [1] |
| Bit planes | --  | `bit_mask` (property)  | `bit_mask` (property)  |
| Metadata | `meta` | `meta` | `meta` |
| WCS  |   `wcs`   | `wcs`  | `wcs` |
| Uncertainty (has setter) |  `uncertainty` (should have attribute `uncertainty_type`) |  `uncertainty` (should have attribute `uncertainty_type`) | `uncertainty` (should have attribute `uncertainty_type`) |
| origin (`xy0`) |  Not present | `origin` (defaults to zero for backwards compatibility) | `origin` (defaults to zero) |
| arithmetic operations | No | No | No |
| Sliceable? | No | No | No |

I'm not 100% convinced, now that I think I fully understand the role that origin plays in the LSST stack, that it should be exposed at the `NDData` level. Instead, an astropy-integrated LSST image class needs to properly use the pixel information passed into the WCS object to return a world coordinate and vice versa.

It would be completely valid for this `LSSTData` to handle slicing by updating origin instead of WCS, as long as the interface presented to the user allows her to go from pixels in the subimage to a world coordinates.

mwcraig:
If we want to introduce an `NDArr` class that is like the LSST `MaskedImage` it would look something like this:


## Attributes/methods of `NDArr` object (new class like LSST's `MaskedImage`)

|       | **astropy v1** | **astropy v2** | **astropy v3** |
| :---: | :-----------:  | :------------: | :------------: |
| Pixel data (slicable) |  NA    | `data`   | `data`   |
| Binary mask (numpy convention) | NA | `binary_mask()` | `binary_mask()` |
| Bit planes | NA | `bit_mask` (property)  | `bit_mask` (property)  |
| Metadata | NA | No | No |
| WCS  |   NA   | No | No |
| Uncertainty (has setter) |  NA |  `uncertainty` (should have attribute `uncertainty_type`) | `uncertainty` (should have attribute `uncertainty_type`) |
| origin (`xy0`) |  NA | `origin` (defaults to zero) | `origin` (defaults to zero) |
| arithmetic operations | NA | binary and unary | binary and unary |
| Sliceable? | NA | Yes | Yes |

mwcraig:
Questions:

1. Should the `nddata` module continue to have both a container (currently `NDData`) and a concrete image-like class (currently `NDDataArray`)?
2. How should masking be handled?
    1. `mask` becomes an object which tries to behave both as a property and as a method (or at least a property with methods).
    2. `mask` is deprecated in astropy 2, removed in astropy 3. Method `bit_plane()` is introduced for bit planes (these were once called flags), and `binary_mask` is a property that returns a binary mask. The name `mask` will *not* be used by NDData but could be used by subclasses.
3. Should we switch from a bunch of mixins for implementing arithmetic and slicing or switch to a more LSST-like separation with arithmetic included by composition instead of inheritance  (with a target for implementation in **astropy v3**)?

mwcraig:
A couple questions that @perrygreenfield raised offilne:

> masks:

> Seems to be lacking detail on how much of bitplanes work. E.g., it says bit_plane() returns a bit plane. A single one, a logical combination of bit planes? Perhaps it should say it returns a boolean from a set of bit planes, and there are a number of options for how it does it?

> Should there be a statement about whether both binary_mask and bit_plane can coexist? (Would seem weird and confusing if they did).

The intent (though not what I actually wrote, unfortunately) is that `.bit_plane` gives one access to the bit plane object. It might be better to have `.bit_plane` be a property that presents the bit plane object rather than a method that simply returns an array.

The intent for `binary_mask` is that it is either simply a numpy array of mask values (or at least acts like one) or a property that returns the numpy-style mask given the state of the bit plane object.

In this iteration, if `bit_plane` is not None, then `binary_mask` should return a numpy-style mask instead of None. It would be valid for `binary_mask` to not be None but `bit_plane` is none.

mwcraig:
The other question from @perrygreenfield 

> Uncertainty: 

>How to handle more complex uncertainty models? Should we outline how that would be done (how it would be generalized)?

Good question...haven't thought it through beyond the statement that having an uncertainty implemented as a variance should not restrict the  more general class from having some other kind of uncertainty, though using that implementation would, of course, require that the uncertainty be reasonably easy to transform into a variance.

TallJimbo:
I think implementing a good class for handling multi-bit masks is probably the hardest part of this - hard enough that it may be best to try to work that out independently of all of the other issues.  I'll try to post on other questions later and say a bit more about masks now.

- I think you want something that behaves mostly like an integer array with the same shape as the other image planes - but ideally it'd be implemented as a `uint8` array with an extra dimension; that would let the number of planes be any multiple of 8, not just the number of bits in the standard integer types.  That "dynamic bit dimension" would be a lot harder, though (more on that below), and it might make sense to have the first implementation just based on a standard integer type.  LSST just uses a fixed number of bits right now.

 - The Mask needs to hold some kind of two-way dictionary that manages the mapping between bit numbers and string names (and possibly with some longer-form documentation for each bit).  And I think there needs to be some way to make that dictionary "mostly consistent" between mask instances, when possible (i.e. in a particular project you probably want one bit plane to have a consistent definition).  This is perhaps the trickiest part - LSST's mask plane dictionary object has a singleton component that's shared by *all* mask instance.  Almost always just does what the user wants and expects if they don't think about it too hard, but it can be *really* confusing when it doesn't, so I'm not sure I'd recommend it to anyone else.  Having some shared (but not fully global) dictionary that defines shared mask plane definitions might be a better choice, but even that can make surprising links between objects that users would normally consider distinct.  I/O also needs to be considered here: when should reading a file from disk modify the shared definitions?  When should the shared plane definitions cause bits to be reordered when reading?  I don't have good answers for these questions, and it may be that shared mask plane definitions are just more trouble than they're worth, but they're worth thinking about.

 - You obviously need a way to extract a boolean array according to some expression of which bits you do and do not want.  That's pretty straightforward if you just use a standard scalar integer for the expression, bitwise-AND it with a standard integer array, and then take just the nonzero pixels.  It'd be a lot harder with the dynamic bit dimension approach.  I think you probably want to have a method to do this (rather than expecting the user to just do the bitwise operation on the array itself) to leave the dynamic bit dimension option open in the future.

 - Be careful about providing access to individual bit planes as bool arrays: there will be a lot of temptation for users to get and operate on these, which of course unpacks them and sort of defeats the purpose of packing things into bits.  It'd also be easy for users to think they're getting a view, and being confused by changes to the bool arrays not propagating back to the multi-bit mask.  LSST has gotten by just fine without any interface to do this (besides applying a single-bit mask to the array, which makes it very clear what's happening).  Of course, having a new class that actually does provide a bool-array-like *view* to a bit could be a very nice way to handle this.

 - I personally don't think it makes sense to have a way to get a boolean mask from a many-bit mask without making the user specify which bits they care about; I think the question of which bits are meaningful is too context-dependent to be carried around in the mask object itself.  But I this is a subject on which reasonable people can disagree, and my perspective is probably heavily influenced by the ways in which we use mask planes in LSST.

mwcraig:
@TallJimbo -- sorry for the extremely long radio silence on this. I have afternoons (CDT) open for the next week and would like to iterate on the NDData APE so as to have it polished going into the astropy coordination meeting Sep 18-20. 

I agree with your points about the need for a carefully designed "mask" object (quotes because the astropy and LSST currently use the term in two different ways). I think we can make progress on the `NDData` (or `NDArr`) object without resolving all of the mask issues and agree a separate APE for hashing out the details of a fairly generalized bitplane/mask object would be helpful.

I think (but please do correct me if I'm wrong) that the initial priority is to make sure that data that implements the astropy `NDData` interface can easily move into the LSST pipeline  and that gridded data in the LSST pipeline can easily be fed into astropy functions (or into astropy affiliated packages).

A longer term goal might (or might not) be to provide a class in astropy that uses some of the LSST code for handling masked arrays. My understanding from @parejkoj and @r-owen is that would be a substantial undertaking so I don't think we should start there.

Could please take a look at the three comments in this thread starting [here](https://github.com/astropy/astropy-APEs/pull/14#issuecomment-306299104) and let me know what you (and everyone who is interested at LSST) think?

My sense is that if we can agree on some names then writing a class which allows an LSST image to present itself as an NDData object would be fairly straightforward, as would translating an NDData object into an LSST image.

Once again, sorry for the long delay!

TallJimbo:
I think a concrete `NDArr` object that resembles LSST's `MaskedImage` is much more likely to be a useful connection point between LSST and astropy than a `NDData` container that resembles LSST's `Exposure`.  That's basically because having a useful common interface for the latter would require common interfaces for essentially all of the metadata objects they hold, and I just don't think that's a realistic goal.

The correllary of that, of course, is that then both projects need to make sure that lower-level interfaces that only need an `NDArr` (or `MaskdImage`) are identified and exposed as well as possibly more convenient higher-level interfaces that operate on `NDData` or `Exposure`.

With that in mind, I'll start with @mwcraig's questions from June 6:

> Should the nddata module continue to have both a container (currently NDData) and a concrete image-like class (currently NDDataArray)?

I think so.  I'm skeptical that there's any value to making either of these polymorphic rather than just making them both fully concrete, because:

 - It's useful to have a container-like class that combines the image-like object with structured and unstructured metadata, but I think you want the default implementation to be sufficiently flexible in what you can attach (and optionally not attach) to it that there isn't a whole lot of value to being able to subclass it.  What people will care about subclassing are the structured metadata objects you stuff into it.

 - The image object needs to be expose very low-level array or buffer interfaces, so you can pass it (or the things that comprise it) to C/C++ code, and you can't have the interpretation of those depend much on a Python method call that's overrideable.  I think that means that the uncertainty in particular needs to have a finite, predefined set of possible implementations (this may have been what you intended with `uncertainty_type`) so all consumers of these objects can know what to expect.

> How should masking be handled?

As I mentioned in a previous post, I'm in favor of just having `bit_planes` and no `binary_mask`.  I think both could be regular attributes rather than methods and should unless you need to keep a method for backwards compatibility.

> Should we switch from a bunch of mixins for implementing arithmetic and slicing or switch to a more LSST-like separation with arithmetic included by composition instead of inheritance (with a target for implementation in astropy v3)?

I don't feel as strongly about this as other LSST people, but I do think what really matters is to have the right set of arithmetic operators on the `NDArr` object, and that it probably doesn't make sense to have them on `NDData`.  I suppose I think this is just one more thing that gets massively simpler with no real loss in functionality/flexibility if both concepts are implemented as concrete classes that aren't expected to be subclassed.


I think the only other point I'd like to make is that (unsurprisingly, perhaps), I strongly support adding `origin` to both `NDArr` and `NDData` (even though the latter also has a `wcs`).  While an integer offset is a special kind of `wcs`, it's a sufficiently important and simple one that it merits special consideration, and that includes putting it in a lower-level object like `NDArr` and calling it out as a special component of `NDData` that's distinct from `wcs`.  I'd also strongly recommend defining the `wcs` such that the actual coordinate system of the `NDData` pixels is defined by the composition of the `wcs` transform and the `origin` scaling, so you don't have to modify `wcs` in concert with `origin` every time the latter is changed.  That does imply that the WCS you write to FITS should represent that composition (rather than the attached `wcs` object on its own), so external tools see the same coordinate system Python code would.  Of course, we're also all hoping that someday we'll be using on-disk WCS formats that can represent composition natively.

jehturner:
I (and I believe STScI) already intend to use a `wcs` that's a composition of transforms (via `gwcs`) like @TallJimbo suggests. Here, `origin` would just be a special case of the first co-ordinate transformation in a chain. The scheme doesn't always translate perfectly to FITS, however, so maybe what you want is just to expose some version of FITS `CRPIX` as an `origin` attribute. It has also been proposed that `gwcs` should be able to write the closest FITS approximation of its WCS, if that's useful.

mwcraig:
Reporting back on discussions at the coordination meeting so far...please comment if interested

+ Names for "mask"-related things:
    - `binary_mask` will be added (some users have only a `binary_mask` and/or use numpy-style masking. This may be a simple mask array or a property that calculates a mask from the `bit_plane`, or it may be `None`.
    - `bit_plane` will be added, though details of what this object is are up in the air.
    - `mask` will still be present (a substantial amount of code that depends on astropy.nddata uses the current `mask` property). The name will be dropped from the documentation (except for a notice about transitioning to the new name).
    - I'm thinking about this like the `clobber`/`overwrite` naming issue in `io.fits`. (almost) Everyone would like to switch the name to `overwrite` but too much existing code uses `clobber`, so it will (I think) be deprecated by documentation.
+ `origin` will be added to both `NDData` (container) and `NDArr` (concrete array-like thing).
+ Uncertainty
    - `NDData` (container) will permit any kind of uncertainty (as it does now).
    - `NDArr` will allow as uncertainty anything convertible to standard deviation/variance. What it actually uses internally hasn't been decided (but should be an implementation detail).
+ Still unresolved:
    - Future of `NDDataArray` (could just change internal implementation to use `NDArr`) (could eliminate)
    - mixins - keep? drop?

mwcraig:
ping @MSeifert04

MSeifert04:
I still don't like the part where `mask` is hidden/and or replaced by `binary_mask`. It actually throws away the "duck-typing" part that makes it behave like `MaskedArray`, `MaskedColumn` and similar classes.

While I agree that "bitplanes" is a good addition I don't think we need a `binary_mask`. What is the disadvantage of just using `mask` as if it were `binary_mask` with regards to "This may be a simple mask array or a property that calculates a mask from the bit_plane, or it may be None".

mwcraig:
My understanding is that both astropy and LSST have been use `mask` with completely different meanings.

You raise a good point though; in Python (or at least numpy and scipy) there is already well-established meaning, whether we like it or not, of `mask`: it is boolean, and `True` means the pixel is left out of calculations.

So, for the LSST folks: is it a deal-breaker if we were to use `mask` instead of `binary_mask`?

mwcraig:
@MSeifert04 and @mhvk -- I'd like your feedback on this question: what do you think about switching from the current design, i.e. lots of mixins to add slicing, arithmetic, with uncertainty handled via a separate class to arrive at something concretely useful like `NDDataArray` or `NDDataRef`, to something like the LLST design, where the  `NDArr` class has the methods for arithmetic operations and propagation of uncertainties (in the simple case where the uncertainty is convertible to variance or stddev)?

This is separate from whether the classes `NDDataArray` or `NDDataRef` stay around; clearly those could be made to work with the current API even if the internal implementation changes to something like `NDArr`

mwcraig:
fyi, the current nddata design (except `CCDData` is now in `astropy.nddata`)

current-nddata

mhvk:
@mwcraig - I think our problem so far has been that we have different usages in mind, and we introduced the mixins to accommodate cases were slicing makes no sense or where uncertainties are not known. I think in principle that remains a reasonable path, as long as one has a "standard" implementation using all of them (which now is `NDDataArray`). I think it is definitely fine for that default implementation to deal with uncertainties as well, as long as it remains fairly straightforward to swap that out (e.g., for my propagation that includes covariances, or for @eteq's distributions).

Overall, I don't see a specific reason to give up on the mixins.  But I'm also not quite sure what the problem is that we're trying to solve: is the current implementation a problem for making a suitable LSST class?

MSeifert04:
Good question but I don't fully understand it. Do you mean removing `NDDataBase`, `NDData`, `NDArithmeticMixin`, `NDSlicingMixin` and `NDIOMixin` in favor of a new container and `NDArr`?

mwcraig:
Removing` NDArithmeticMixin`, `NDSlicingMixin`, `NDIOMixin`, keeping `NDData` as a very general and flexible container (basically like it is now). 

Realistically, `NDDataArray` and  `NDDataRef` cannot vanish in 3.0 because too many other people use them. The earliest I'd propose removing them is 4.0 (so deprecate in 3.0 with a very long timeline for removal).

`NDArr` would replace those, with `CCData` modified to use `NDArr`.

I should note I'm not 100% sold on this approach, which is why I'm asking for feedback. `NDArr` seems like a simpler way to go than the current setup, though it is less general.

mhvk:
:+1: to a dynamically derived mask, and I don't thinking a mask being settable has to be a requirement (though it might be trivial to implement with bitplanes, just have an additional one called "user_mask" that always is OR'd in to `mask`).

Also :+1: on the bitplanes having nice, self-explanatory names.  Is there a prototype for `bit_planes`? I think it would be quite useful elsewhere...

crawfordsm:
That definitely sounds great to me and even very useful so 👍  from me as well!

parejkoj:
@mhvk I dont' think we want a "user_mask": `mask` either exists as a bit array of the same shape as `data` (True=bad pixel), or it's None and is computed from `bit_planes` in an as-yet unspecified way, or if both `mask` and `bit_planes` are None, then all pixels are always valid. I'd advocate enforcing that only one of `mask` or `bit_planes` can be not None.

As an example (that I'm *not* advocating we use as the API in this design, but that would probably be the under-the-hood calculation), here's how LSST's `mask` object (the planes of bits) is used to compute the "bad pixel mask" for an operation that should reject pixels marked "bad", "saturated", and "no data":

```
maskVal = mask.getPlaneBitMask(["BAD", "SAT", "NO_DATA"])
badPixels = (mask.getArray() & maskVal) > 0
```

mhvk:
@parejkoj - my specific suggestion would be to include a bit plane "USER", which is the one that gets affected by setting `mask`. But really this is jumping far ahead - I would definitely start by just allowing `mask` to be calculable, and not allowing it to be set.

jehturner:
Deriving a binary mask from bitplanes sounds fine to me. I just need the bitplanes...

I should be able to switch to `NDArr` (or whatever it's called), as and when I find time, as long as the functionality is there. There is already a bit of a proliferation of classes, which might be confusing and offputting for new users.

jehturner:
So it seems like there may be some consensus on `NDArr`, but I'm not clear on where `meta` is now kept, if `NDDataArray` and `NDDataRef` are deprecated? Does `NDData` have a `.arr` and we then sub-class that to implement all the functionality that `NDDataArray` had (or not)?

mwcraig:
Good point; I'd been so focused on `NDArr` I forgot about that. A few thoughts:

+ `NDDataArray`/`NDDataRef` present almost the same public interface as before, but the internal wiring is a little different. 
+ The numerical bits of `NDDataArray`/`NDDataRef` are handled by `NDArr` rather than mixins.
+ The tricky issue then becomes one of slicing and WCS, but that can be resolved in a sensible way I think (and maybe also in a backwards-compatible way).

I'll try to formalize all of the recent suggestions here into an update of the APE over the weekend.

jehturner:
That sounds reasonable. While `NDArr` might be the broadest standard, I think there's also collaborative value in code being able to accept an `NDDataArray` / `NDDataRef`-like object as input -- which is more awkward when there are two possible parents that do similar things. Could we therefore get rid of one of them and keep the other (especially if we're getting rid of mixins and are already resigned to some disruption anyway)?

The most obvious justification for having both classes would seem to be compatibility with old code, but I wonder whether one could be sub-classed from the other, so that it doesn't necessarily break but there is really only one implementation and `isinstance(x, 'whichever_is_the_parent')` will work for both (as well as duck typing, where that's better).

parejkoj:
@mwcraig Just curious what the status of the updated APE is? I think we've had enough LSST-astropy agreement now that there shouldn't be many issues from our end, but we'll all probably have to read through the final document again to remind ourselves of everything that we've discussed over the past year+.

mwcraig:
@parejkoj — I’ve been about 2 days from finishing the APE update for almost 2 weeks. Just wrapped up something that had been consuming a lot of time, so hopeful I’ll have the update ready by the weekend. Then one more broad call for comments, and assuming no major objections, the APE gets merged and the coding starts to implement it!

mwcraig:
An update to the APE language is pending (https://github.com/perrygreenfield/astropy-APEs/pull/3) but in the meantime I'll outline the revisions in the comments here.


## API additions

+ The attribute ``origin`` is added to ``NDDataBase`` and its subclasses.
+ These classes are added to the ``nddata`` package: ``NDArr``,
  ``VarianceUncertainty``, and a bit plane class ``BitPlane``.
+ A `bit_plane` property is added to ``NDDataArray`` and ``CCDData``.

## API changes

+ Only one of ``mask`` and ``bit_plane`` can be set on those object that have a ``bit_plane`` property.
+ If  the ``bit_plane`` property is set, then ``mask`` is calculated from ``bit_plane`` (the user must have indicated in some way which bit planes to use).

## API not changing

Though it is unusual to call out things that are not changing,  it seems appropriate to do so for  ``mask`` given the extensive discussion of the topic. 

+ The meaning of ``mask`` will be the same as in current ``ndddata``: a binary mask that follows the numpy masked array convention of ``True`` means ignore , ``False`` means use.  
+ ``mask`` is still settable as in the past, though it will raise an exception if one sets ``mask`` and then tries to set ``bit_plane``.

## API deprecations in astropy 3 for removal in astropy 4

+ The ``FlagCollection`` class and the ``flags`` attribute of ``NDDataArray``
  and ``CCDData``. This functionality will be replaced by an implementation
  of bit planes.
+ The ``__array__`` and ``__array_prepare__`` methods of ``NDDataArray``
  and ``CCDData``, which allow them to be used as numpy arrays in expressions.

mwcraig:
The diagram below summarizes the new class structure; most of the changes are in ``NDDataArray`` and its subclasses. That is an effort, in part, to ensure backwards compatibility of ``NDData`` (the container class) as much as possible.

I do realize the bit plane class is not on here yet...

new-nddata

mwcraig:
@parejkoj -- this is ready for more feedback. I know there still needs to be some work on the details of the bit plane object, but I agree with @TallJimbo's advice above that users should be presented with a way to specify which planes are included in calculating the mask rather than access to individual planes. 

Not sure I'll have time to implement anything more than a very basic bit plane before feature freeze for 3.0 next Friday, but my intent is to get as much of the stuff in this APE in there as I can....

mwcraig:
@jehturner @crawfordsm @MSeifert04 -- comments welcome!

jehturner:
@mwcraig: Many thanks for the summary and sorry I can't really do this justice before the code freeze (I only became conscious of the latter recently and just couldn't find much time).

I think this sounds quite sensible overall, at least at a glance. Does this mean that the name `NDArr` is about to be baked in? I'm still a bit wary of that, but it can't hold things up indefinitely. I'm OK with deprecating flags, even though it means I'll have to rewrite my code (which I have to finish making Python 3 compatible anyway). Any special reason for deprecating `__array__` and `__array_prepare__`? I think those seem useful off hand, albeit not in the spirit of LSST's usage.

Will `NDArr` contain (or behave like) a `MaskedArray`? I'd think that could help with compatibility, eg. where existing functions elsewhere know what to do with one (such as `LinearLSQFitter` as of today). 

The way you have divided things up for backwards compatibility seems fine, but I'm again thinking that. in principle, there might be some code that could be compatible with either `NDDataArray` or `NDDataRef` (or similar classes, but not `NDDataBase`) and it's a bit messy asking, for example, "is x an `NDData`-like thing with arithmetic?", without a common ancestor. But maybe we just want to move away from supporting `NDDataRef`.

OK, have to go now. Hope that's slightly helpful (& thanks for the work and happy Christmas!).

pllim:
It has been 6 years since the last reply, should we just say this won't go in at this point, or are the interested parties going to revisit this?

parejkoj:
Thanks for the reminder. I think we're still interested in looking at differences in how we handle mask bit fields: we've already adopted CCDData for our alert packets, and submitted a PR to include a PSF cutout to support that. I'll see if we can find some time to discuss this more on the LSST side.

pllim:
Thank you for your response, @parejkoj !

It has been almost another year since and CoCo reached out to VRO again. It appears while there is definitely interest, unfortunately there is no available bandwidth to pursue this in the near future. Therefore, CoCo has decided to reject this APE for now. However, I will open a follow-up issue to revisit should both interest and bandwidth return in the future.

Thank you all again and sorry this did not work out this time around.

PR #17, opened 2016-11-17
Use Cookiecutter for the package-template (APE 12)

Cadair:


adrn:
This week is a little crazy because of travel so I'll have to do a more thorough review in a few days, but my gut reaction is that I'm +1 to the idea. I don't know much about `cookiecutter` but it sounds like it allows for a much nicer solution than copy-pasting directions from the package template setup.

APE12.rst
```diff
@@ -0,0 +1,188 @@
...
+recommended to affiliated package authors in the
+`Astropy documentation <http://docs.astropy.org/en/latest/development/affiliated-packages.html>`_.
+
+The second advantage to using a templating system like Jinja2 is that it makes
+it possible for other projects (*e.g.* SunPy) to maintain a fork of the package
+template with defaults that specific to that project. Without having to make
+extensive changes to the actual package template, thereby making it easy to pull
+in changes from the Astropy template without lots of complex merging.
+
+There are also secondary advantages to using a Cookiecutter project, such as
```
adrn: Two "such as" clauses in the same sentence

APE12.rst
```diff
@@ -0,0 +1,188 @@
...
+status: Discussion
+
+
+Abstract
+--------
+
+This APE proposes migrating the `package-template
+<https://github.com/astropy/package-template/>`_ to use the `Cookiecutter
+<http://cookiecutter.readthedocs.io/>`_ project to provide templating facilities
+to the package template to ease use of the package template and to make it easy
```
astrofrog: repeated 'package template' - maybe reword. Somewhere in abstract you need to spell out more clearly that creating new projects will be much easier for people (ease use doesn't quite cover it). Maybe have a separate sentence for the fact it's easier for people to use, and to 'subclass'. You said it better in the detailed bit. Can you try and improve it here too?

APE12.rst
```diff
@@ -0,0 +1,188 @@
...
+`Astropy documentation <http://docs.astropy.org/en/latest/development/affiliated-packages.html>`_.
+
+The second advantage to using a templating system like Jinja2 is that it makes
+it possible for other projects (*e.g.* SunPy) to maintain a fork of the package
+template with defaults that specific to that project. Without having to make
+extensive changes to the actual package template, thereby making it easy to pull
+in changes from the Astropy template without lots of complex merging.
+
+There are also secondary advantages to using a Cookiecutter project, such as
+being able to include code, such as tests, in the ``package-template``
```
astrofrog: use a link instead of double backquotes?

APE12.rst
```diff
@@ -0,0 +1,188 @@
...
+Alternatives
+------------
+
+Keeping the current solution. This prevents any trouble for anyone using the
+'git update' method at the moment, but it does not provide any of the templating
+benefits presented above.
+
+Cookiecutter is not the only project templating system, others such as
+`diecutter <https://diecutter.readthedocs.io/>`_ could be used. Cookiecutter was
+chosen due to wide adoption and implementation in Python.
```
astrofrog: Another alternative is to write our own setup script. Might be worth mentioning at least.

Cadair: done

dkirkby:
I used my own cookiecutter template for new python projects until I discovered and switched to the astropy affiliated template because it was much more comprehensive in its support for RTD, travis, etc, and "just works".  This proposal sounds like it would significantly simply the initial setup (although I found it useful to go through the steps manually, at least the first time, so I knew what was under the hood).  

How do bug fixes and maintenance updates (e.g., version bumps) to the template propagate to affiliated packages in this proposal?  I guess this part of the proposal is relevant but I wasn't sure how the automatically generated template repo would solve the problem:

> As implemented in Cadair/package-template users who have followed the instructions where they use the package template as a git remote will no longer be able to pull updates from the package_template git repository. This could be mitigated by automatically generating a rendered version of the package template in another git repository or in a different branch.

Cadair:
Thanks for the feedback!

The idea is that for people who want to continue a git-remote based update workflow (where they start a project with the git history of the template) could pull form the automatically generated repo.

My thoughts on people who start with cookiecutter would be that you could render any updates to a new folder and then diff the updated template to your project and see any changes and pull them in as needed.

adrn:
@Cadair Hmm, that sounds quite a bit more tedious than the current workflow though...

Cadair:
@adrn really because isn't that what you are doing in a mature project at the moment? Some files have changed in upstream and you have to manually merge them?

adrn:
@Cadair Sorry, I meant the 2nd case above -- rending updates to a new folder and diffing.

Cadair:
Yeah I did as well, although I went a little to "git" in my terminology. If you are not tracking a git remote, you have the rendered template in a `package_template` folder on your disk and your project, how are you merging it and managing updates at the moment?

bsipocz:
What about adding a working solution to https://github.com/astropy/package-template/issues/197 to this APE? I agree with the others that having an easy way to update an existing package that uses the template is just as important (actually in my view it is more important) as having an easy way to start up something .

dkirkby:
I suspect most users don't care about inheriting the template git history (and it has burned me a few times when I forgot to delete the template repo's tags).  The ideal update workflow for a new user is something like:
```
% ./update_template
Your template is up to date.
% ./update_template
Your template needs updating but you have modified files that must be
committed / stashed / reset first.
% ./update_template
...
The following template files have been updated:
  setup.py, travis.yml
Use "git diff" to review the changes then "git add" and "git commit" to accept them.
%
```

Cadair:
While I agree that having a nice update mechanism is important, I am not sure I want to conflate that with this PR. (Ignoring the git remote case) I can't see how this make things *worse* and because cookiecutter will remember the settings you used to render the template the first time it is possible to get your customised version of the template for diffing, which could be scripted out later.

dkirkby:
I agree with @bsipocz that straightforward updates are at least as important as a more convenient initial install.  The main benefit of the template is that it allows developers to adopt the best practices of the astropy community (RTD, travis, etc) without necessarily understanding all of the details of how they are implemented (which was true for me when I started using the template).  However, a lower barrier to adopting best practices also requires a simple and robust mechanism for keeping the relevant files up to date.

mhvk:
Another agreement with the importance of updating: I have one project for which I use the astropy template, and I found setting it up via the current instructions not particularly onerous, but I find updating it painful (mostly because it needs to be done so infrequently that I have no memory any more about how it is to be done; indeed, I have no clue whether I'm currently up-to-date).

That said, I do like the idea of providing a more automated way of getting a useful environment.

astrofrog:
One idea I've had for a while would be that repositories which would like to see updates could include a configuration file (e.g. ``.astropy_template.yml``) which could indicate which files should be updated automatically, and a bot could open a PR to do this. But that's a separate issue from the initial creation.

hamogu:
Agreements that updating is important. I currently have 4 or 5 packages that use the template, most of them are for my own personal use (although two or three might be advertised to a wider community in the future and will hopefully grow into affiliated packages).

Suggestion: Poll the astropy template users list to find how many people use git for updating vs. a more manual approach.

mhvk:
@hamogu - thanks for mentioning the template users list -- I didn't even know it existed... (for others like me: https://groups.google.com/forum/#!forum/astropy-affiliated-maintainers)

cdeil:
@Cadair - Thanks!

I'm +1 to introduce cookiecutter for this.

Along the lines of what several others have stated, I think it would be important to think through (or even exercise through) how updating existing packages will work. That's been the painful and time wasted part for me.

Currently you have this which is pretty vague and thin IMO:

> As implemented in Cadair/package-template users who have followed the instructions where they use the package template as a git remote will no longer be able to pull updates from the package_template git repository. This could be mitigated by automatically generating a rendered version of the package template in another git repository or in a different branch.

I currently have the following thoughts / questions, which I think would be useful to address in the APE (not just the GH discussion here):

- Should https://github.com/astropy/package-template/pull/202 be merged? Or would it better to leave the existing repo as-is and start a new one. In the discussion you say "maybe there will be a rendered version in a second repo" ... that needs to be decided so that the dicussion can be more concrete and move on, no?
- If we merge https://github.com/astropy/package-template/pull/202, then the "git update" method will not work any more for anyone (existing or future repos). Too much changes and it's practically a new repo. Personally I think "git update" never really worked well, I tried it a few times and there were always nasty git merge conflicts, it made the affiliated project history and contributor lists messy ... so good riddance to "git update"! But make this more clear in the APE, please.
- Should this change be taken as an opportunity to re-think what's in the astropy-helpers and package-template files and which files in astropy-helpers and package-template must be version-coupled? I mean, is it really necessary to have so many boilerplate files in the package-template. Can't some of them be removed completely (e.g. ez_setup, can't we just require setuptools and pip which has been the standard for years) and some others moved to astropy-helpers? A lot of the complexity comes from each affiliated package having two somewhat linked, somewhat independent connected other sets of files: astropy-helpers and package-template.
- I think a little tooling (i.e. a script) to help with updating a package to new astropy-helpers or package-template versions would be awesome. It's probably tricky to maintain that script, so that it works for older and newer projects. Maybe a section in the APE discussing if it's possible would be helpful?

bsipocz:
> I think a little tooling (i.e. a script) to help with updating a package to new astropy-helpers or package-template versions would be awesome. It's probably tricky to maintain that script, so that it works for older and newer projects. Maybe a section in the APE discussing if it's possible would be helpful?

nit-picking, but the ``astropy-helpers`` update in affiliated packages (and other packages that opt in) is already working more or less like that through a script (thanks to @astrofrog). Whenever a new release is made, the person releasing the helpers can run the script that generated automated PRs to the packages as part of the release process. Obviously this could be upgraded to do by a bot or whatever, but the current way is already much better than it used to be to do the updates and puts basically no burden on the package maintainers.

With the rest I agree.

astrofrog:
And I think we can easily adapt the astropy-helpers update script to also update e.g. ``ez_setup.py``, ``_astropy_init.py``, ``setup.py``, and so on.

bsipocz:
Not ``setup.py`` though, as many packages had it customised.

cdeil:
Where is this update script?
Is it advertised where people might look for it?
Should it be mentioned in the APE?

---

Mostly, but not fully unrelated question to this APE: where do we have / should have / will have documentation for `astropy-helpers`, `package-template` and the cookiecutting and update process?
At the moment it's a bit of a mess, especially `astropy-helpers` doesn't have docs) and is really hard to understand / contribute to, no?

---

I'll stop sidetracking this now. @Cadair - Take whatever aspects you think in scope for the APE and feel free to ignore the rest!

astrofrog:
@cdeil - https://github.com/astropy/astropy-tools/tree/master/update-affiliated

astrofrog:
@bsipocz - I think ``setup.py`` should now be in a state that it can be package-independent because all the info is in ``setup.cfg``. But yeah, I agree we can't assume, so that's why I was mentioning the idea that there could be a config file in the repo that indicates which files to auto-update.

dkirkby:
The `update_astropy_helpers.py` script seems geared towards packages already within the astropy release ecosystem:

> This script is not intended to be run by affiliated package maintainers.

However, I suspect that most packages that started from the astropy template are not in this category, i.e., they use the template but are not (yet) formally affiliated.

The [affiliated package docs](http://docs.astropy.org/en/latest/development/affiliated-packages.html#updating-to-the-latest-template-files) do provide a user recipe for updating astropy-helpers but is it still valid?

bsipocz:
@dkirkby - The way you've linked is still the recommended one to update the helpers. Alternatively, upon request, we can add packages to the list that are getting the auto updates of the helpers (they don't need to be affiliated).

dkirkby:
@bsipocz Thanks!  I just tried it and it works (but the version should be bumped from v0.4.3 to v1.3).

mhvk:
Wow, I'm certainly learning from this thread -- so let me echo the "we need a central place to document this" comments...

bsipocz:
While some more docs wouldn't hurt, there are two documents out there summarizing what the average user may ever want to know about the helpers and about how to use it: One is the docs to set up and maintain an affiliated package here: http://docs.astropy.org/en/latest/development/affiliated-packages.html#updating-to-the-latest-template-files, while the other is the astropy-helpers APE: https://github.com/astropy/astropy-APEs/blob/master/APE4.rst

cdeil:
@bsipocz - The problem with https://github.com/astropy/astropy-APEs/blob/master/APE4.rst and other APEs is that they are mostly fixed. This is bad if they are the docs for things that change over the years.

It's not even clear if APEs are supposed to be fixed / extended as things change.
See e.g. https://github.com/astropy/astropy-APEs/pull/16 which contains a small fix to the ECSV APE, it was never merged.

What if I want to contribute to `astropy-helpers` docs? Should I go edit the APE?
Probably better to transform the APE content into proper RST Sphinx versioned docs and maintain it there for the coming years, no?

Cadair:
On the documentation front, a transition to cookiecutter allows us to maintain files, i.e. tests and docs in the package template repo that do not get given to people when they run the cookiecutter command. So we could have a full set of docs in the package template repo rather than scattered around.

cdeil:
> So we could have a full set of docs in the package template repo rather than scattered around.

You mean joint docs for astropy-helpers and package-template?

I think that's a good idea...

Cadair:
yes it could mean that. That would obviously be up to the maintainers, but as mentioned in the text is one of the advantages of this APE.

bsipocz:
@cdeil - I agree that the APE is not ideal as a docs, but it's more than no docs, and actually answer to the general purpose question: "What is astropy-helpers and why we need it." It's definitely not adequate enough if someone considering contributing to the helpers itself.

> What if I want to contribute to astropy-helpers docs? Should I go edit the APE?
Probably better to transform the APE content into proper RST Sphinx versioned docs and maintain it there for the coming years, no?

If you are willing to do that, then I think the best would be to start up a docs directory at the helpers repo and as the first approximation copy the text of the APE over to there and update where needed.

cdeil:
> If you are willing to do that, then I think the best would be to start up a docs directory at the helpers repo and as the first approximation copy the text of the APE over to there and update where needed.

Yes, I would be willing to spend some time on docs.

But let's wait for this discussion to conclude and give others a change to chime in on how to handle astropy-helpers and package-template docs before taking action.

bsipocz:
> But let's wait for this discussion to conclude and give others a change to chime in on how to handle astropy-helpers and package-template docs before taking action.

For one, I'm 👎 on mixing the two under package template. Both the infrastructure (one is a git submodule, the other is a pick and choose list of files/features) and release cycle is different, and while the package-template would benefit of a detailed how-to use user docs that may not be true to the helpers. Anyway if the docs will be mixed, it may better go into astropy core, under the affiliated dev docs there.

Cadair:
The general consensus seems to be positive so far, with some comments about updating and documentation. On the documentation front I don't feel we need to decide anything in this APE, beyond acknowledging that this change would allow the maintainers to create a set of sphinx documentation in the `package-template` repo.

On the updating side, I also wan to minimise the decisions made in the APE, I feel there are ways to maintain parity with the current (less than perfect) workflow. I do not think that @bsipocz's suggestion of including a update mechanism in this APE is needed, or something I have time to work on.

I do think, however, that this APE makes the future development of such a process easier, as it could rely on the templating of the package template by cookiecutter to reduce the number of conflicts by re-rendering the template with settings used on setup (and potentially stored in the repo for prosperity).

As far as the git remote update path goes, without taking further steps merging astropy/package-template#202 would kill that completely. So far on this thread the general opinion seems to be that this would not be any great loss. (I posted to the affiliated package list [here](https://groups.google.com/forum/#!topic/astropy-affiliated-maintainers/_60tYfbjLgk)).

For the sake of completeness, we have options to maintain this workflow if desired. They both involve having a cookiecutter repo and one repo where cookiecutter is automatically run with default settings to create a repo that looks like the current `package-template` repo. This could be done in two ways,

1. as @cdeil suggested make the cookiecutter repo a new repo and update `package-template` automatically.
2. As I allude to in the APE merge astropy/package-template#202 and make a new repo for the rendered version.

Having said all that, I am at the moment going to suggest that unless we hear any dissenting opinions that I modify the APE to remove the option of the git remote update path completely, which would leave us with the requirement to document a cookiecutter based manual update workflow until we automate it.

APE12.rst
Cadair: reworded

dkirkby:
Breaking one of the existing update mechanisms potentially affects a lot of repos. There are currently ~200 public repos on github that contain a copy of .gitmodules from astropy/package-template.

Cadair:
That's got to include forks right?! 

Also having `.gitmodules` from the template does not imply that they are using the git remote method (it means they are using astropy_helpers). Even having the history of the package template is not indicative that they still are using that update method and the same the other way around (they could have squashed the history).

cdeil:
The relevant question is  "how many people / projects are happily using the git update method and would mind if it went away?", no?

Personally I have used it in the past, found it to be very painful, stopped using it and now think it's good if it goes away.

Cadair:
Indeed. @mwcraig replied on the package maintainers list that he believes ccdproc is using that method.

adrn:
I agree with @cdeil and @mhvk -- I usually default either manually updating the files or just let them be out of date...more often than not, the latter. I wouldn't mind if the git update method went away

dkirkby:
@Cadair It looks like mostly unique names but I didn't go through them all:
https://github.com/search?q=filename%3A.gitmodules+astropy_helpers&ref=searchresults&type=Code&utf8=✓

I maintain 5 packages that use the git update method, but I have only used it a few times on each package and agree that it is painful.  Updates without git sound like they could be slightly more painful and error prone, but that could certainly be offset with some clear instructions or (even better) an automation script.

Cadair:
yeah that's all sources not forks. Damn that's a lot of astropy_helpers.

I will work on some prototype docs for a non-git update method so people can see what it would be like. Also if we want to keep it around we can as discussed and I will modify the APE accordingly. At the moment it seems to be roughly landing on the side of removal.

bsipocz:
> That's got to include forks right?!

Cadair: 200 sounds about right. I've updated the helpers in ~30 packages after the latest release, and that number includes only the official affiliates and a small number of additional packages where the maintainers asked for it. Having ~4-5times that many out there using astropy-helpers seems believable.

dkirkby:
The template + helpers have been a great success!   I even use them for packages that don't use any other astropy since its the easiest way to quickly get started with all the "best practices" baked in.

mhvk:
Bit late: I agree with @bsipocz that we should not mix `astropy-helpers` and the template -- though would stress that the helpers are more in need of documentation than the template: the bus factor is now uncomfortably low. 

Of course, the above is separate from the discussion here, except that I do worry slightly about yet another layer of compatibility (with cookiecutter) to maintain. So, good documentation of what happens would seem essential. That said, the actual PR looks like this is not a particularly difficult process to understand.

p.s. Agreed that the template + helpers is very, very useful. Just not having to look up how to build docs and do tests is invaluable (in other words, my packages that did not use the template + helpers also do not have good docs or working tests...).

astrofrog:
@bsipocz - maybe we should adjust the script to actually open PRs to all repositories where we can find astropy-helpers? (rather than restricting ourselves to a limited set)

bsipocz:
@astrofrog - We can try that, and then maintain a list of opt-outers if any.

mwcraig:
I *think* we've been using the git update method is astropy but have found it to be painful, and the tags are a frequent source of irritation to me because I forget to delete them, push to github, then spend forever trying to get them deleted. Someone else (either @bsipocz or @MSeifert04) handled the last template upgrade so I'm not actually positive how it was done 😀 .

Removing that path is not the end of the world, but the fact that 200 people are using the template makes me a little reluctant about removing the method without trying to notify them. 

One minor reason towards moving towards cookiecutter is that people often forget to trim out the parts of the template they don't need. In several affiliated packages people hand entry points defined, for example, that were straight from the template and not actually implemented.

bsipocz:
@mwcraig - Indeed it was me, and looking at the PR I must have copied over the files manually following @astrofrog latest suggestions, and must have dealt with the conflicts by hand. In retrospect I should have tried doing the git update way as either way one needs to deal with the conflicts and I feel the latter is a more proper way (easy to say when never have tried 😀). Messy business.  It would great to have a way to address bugs/updates in connection with the template (I'm thinking e.g. about the ConfigParser issue), so doing https://github.com/astropy/astropy-APEs/pull/17#issuecomment-271377422 will be a good pyastro hack to work on, assuming we'll be there ;).

Cadair:
> Removing that path is not the end of the world, but the fact that 200 people are using the template makes me a little reluctant about removing the method without trying to notify them.

I would like to point out that a decent number of those 200 projects are probably using `astropy_helpers` and not the package template. i.e sunpy and chiantipy are both not inheriting from the package template.
However, I agree with you that we should not just rip the rug out from people without warning. What methods, beyond the various mailing lists could we use to notify people about this change?

Cadair:
> One minor reason towards moving towards cookiecutter is that people often forget to trim out the parts of the template they don't need.

Yes, I think this is quite valuable in a number of ways, one of the reasons I don't use the package template sometimes is for a small hack there is way too many use cases covered by it, which means it's too much stuff to sift through to get to what I need. Being able to opt in and out of different parts should make it more useful as well as more approachable to people who don't even know about the features let alone need to use them.

mwcraig:
> However, I agree with you that we should not just rip the rug out from people without warning. What methods, beyond the various mailing lists could we use to notify people about this change?

Since we have a list of the repositories from @dkirkby's search, it would be relatively straightforward to script opening an issue in each of those repositories. Not sure if that would be seen as helpful or as spam...

adrn:
I think that would be helpful. But I do have one request: can we, at minimum, do `and user not in ['adrn']` because apparently [15 of those repositories are mine](https://github.com/search?utf8=%E2%9C%93&q=filename%3A.gitmodules+astropy_helpers+user%3Aadrn&type=Code&ref=searchresults) :)

Cadair:
I wonder if there is a way to more accurately search for the package template rather than just astropy helpers. I will give it some thought. I wonder if @astrofrog or @bsipocz have an idea of a file or two that is in the template that's likely to be in all the projects that use it. Maybe `_astropy_init.py`??

astrofrog:
Yes, ``_astropy_init.py`` is the right way to go

dkirkby:
There are [202 packages](https://github.com/search?utf8=✓&q=filename%3A_astropy_init.py&type=Code&ref=searchresults) containing `_astropy_init.py`.  If you limit the search to some text added in the most recent (May 2015) commit, that brings it down to [170 packages](https://github.com/search?utf8=✓&q=filename%3A_astropy_init.py+pypi%2Fpsutil&type=Code&ref=searchresults), which suggests that there are 32 packages that have not updated their template files since then.

It would probably be useful to do this more carefully using the github API (e.g., to filter out clones or dormant repos).

Cadair:
I have updated the wording around the update issues. Mainly saying we will provide a new repository to function as a git remote to maintain compatibility with the current git method. Also, a sentence to highlight future tooling around automatic updates.

APE12.rst
```diff
@@ -0,0 +1,203 @@
...
+As implemented in
+`Cadair/package-template <https://github.com/Cadair/package-template/tree/cookiecutter>`_
+users who have followed the instructions where they use the package template as
+a git remote will no longer be able to pull updates from the
+``package_template`` git repository. To mitigate this lack of backwards
+compatibility a new repository will be created which holds an automatically
+rendered version of the template with the default settings. This will result in
+a repository that looks like the current package-template repository. This could
+then be used as a git remote for people who wish to continue using this update
+method.
```
astrofrog: Maybe mention it will share the history up to the point where cookiecutter was added. Also maybe just a new branch in the repo rather than a separate repo?

APE12.rst
```diff
@@ -0,0 +1,203 @@
...
+Keeping the current solution. This removes any need for changes for anyone using
+the 'git update' method at the moment, but it does not provide any of the
+templating benefits presented above.
+
+Cookiecutter is not the only project templating system, others such as
+`diecutter <https://diecutter.readthedocs.io/>`_ could be used. Cookiecutter was
+chosen due to wide adoption and implementation in Python.
+
+Finally, a custom templating system could be developed and maintained for the
+Astropy project.
```
astrofrog: maybe add 'but this would likely require a lot of effort'

APE12.rst
Cadair: could do, not sure which one is better, I suppose a new branch does not require remote change on the local git side.

APE12.rst
Cadair: explicit is better than implicit I suppose :p

APE12.rst
bsipocz: I would be +:100: for having a new branch rather than a new repo for the current template (unless the cookiecutter version goes into the new repo)

eteq:
@Cadair - I like this idea, but have a couple small bits that should be stated expressly in the APE:

1. While initially worried about the fact that this breaks the auto-update, the suggestion of having a secondary branch to support that is a good one.  But one more thing that should be added: this auto-generation should be *automatic* e.g. via a travis deploy script that runs on package-template and automatically pushes a new rendered template to the special branch. I can help with setting that up, as I've done this before in other contexts.  But it should be specified in the APE that this automatic approach will be done, because it won't be kept up to date if it isn't automatic.
2. I see in your trial implementation that there's a set of instructions build into the template repo.  Right now the instructions for the template are in the astropy repo (http://docs.astropy.org/en/stable/development/affiliated-packages.html) - is your intent here to change that so that the authoritative instructions are now in the template repo again?  If so, that should be specified in the APE, if not, you may want to note that the instructions will still remain in the Astropy docs.  

(Also have some thoughts on implementation details, but they don't affect the APE so I will put them in astropy/package-template#202)

Cadair:
@eteq I did mention "automatically" in the text of the APE, I have added Travis to be explicit.

As for the documentation, the move to cookiecutter means that it is *possible* to have the canonical documentation for using the template to be in the template repo. I defer to group opinion on if this is something we want to do. Personally I like it, you would only need one PR to update the docs and the template, easy to version the docs with the template etc.

bsipocz:
I very much like the branching suggestion by @astrofrog.

Cadair:
@astrofrog it being the default repo branch means it is the one that's displayed by default by GitHub right? If so I am fine with it.

astrofrog:
@Cadair - yes, that's correct

astrofrog:
@Cadair - can you update the APE accordingly? (the bit about making the new branch for the rendered version)

eteq:
@Cadair - alright, I'm satisfied with this - thanks!

On the documentation, after a bit further investigation, I think the reason we made it a separate doc in the first place was because it enabled the fancy javascript magic of filling in the package name using a form.  But now that's not needed due to cookiecutter, so indeed it does make sense to move it back into the package-template!

Cadair:
@astrofrog @eteq I pushed the re-word.

APE12.rst
```diff
@@ -0,0 +1,209 @@
...
+<https://github.com/astropy/package-template/>`_ repository while not including
+it in the code that the person setting up a new project receives. This is
+because only the templated directory in the repository gets processed by
+Cookiecutter.
+
+An example of using cookiecutter with the package template is below. This
+example uses the `Cadair/pacakage-template <https://github.com/Cadair/package-template/tree/cookiecutter>`_
+``cookiecutter`` branch, if this work was to be merged the command would be
+``cookiecutter gh:astropy/package-template`` as the ``-c`` option specifies a
+non-master branch.
```
astrofrog: @Cadair - very minor before I approve and merge, we should keep the -c cookiecutter since we're going to merge this into a cookiecutter branch. So can you update it here and in the first command of the code block below?

Cadair: I have just tested this and cookiecutter uses the default git branch rather than "master".

Cadair:
Wooop 🎆

embray:
Great to see this finally happen!

cdeil:
I have a package I'd like to set up in the next days using the package template.

@Cadair - Would you recommend doing it via cookiecutter version you have made now?
(I'm a bit familiar with the process, and didn't like the old scheme with the many manual changes much, so if at all possible I'd like to use the new scheme and wouldn't mind if there's still some issues or rough edges.)

PR #19, opened 2017-01-15
Include a table of APE #, title and date in README.rst

### LINKED ISSUE #10: Add an APE index
If there is one already it would be good if it were linked to in the readme.  If there isn't, there should be (perhaps in the readme, or somewhere else).

taldcroft:
Closes #10.

This is an alternate idea to put the list of APEs directly into the README (for better visibility) instead of a new APE-0.  I don't think the list of APEs will ever get unmanageably long so we can't just put it in the README.

I couldn't figure out how to make relative links to the APE RST docs that will render correctly on GitHub.  If someone knows how to do that then please let me know.  Otherwise I propose just putting this in because as it stands there is no other table of APEs available anywhere AFAIK.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,19 +12,29 @@ accepted, rejected, or modified.
 Accepted APEs
 ^^^^^^^^^^^^^
 
-== ================================================= ===========
+=== =================================================== ===========
 #     Title                                             Date
-== ================================================= ===========
-1  APE Purpose and Process                           2013-Nov-08
-2  Astropy Release Cycle and Version Numbering       2013-Dec-11
-3  Configuration                                     2013-Dec-10
-4  Astropy Setup Helpers                             2014-Jun-28
-5  Coordinates Subpackage Plan                       2014-Jan-22
-6  Enhanced Character Separated Values table format  2015-Jan-26
-7  NDData Plan                                       2014-Dec-17
-8  Astropy Community Code of Conduct                 2015-May-04
-10 Roadmap for Python 3-only support                 2016-Aug-22
-== ================================================= ===========
+=== =================================================== ===========
+1   `APE Purpose and Process`_                          2013-Nov-08
+2   `Astropy Release Cycle and Version Numbering`_      2013-Dec-11
+3   `Configuration`_                                    2013-Dec-10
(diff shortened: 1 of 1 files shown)


astrofrog:
What about just putting absolute links?

taldcroft:
Yeah, that would mostly be right.  But if you open a different branch and view on GitHub then the links will be wrong.  Maybe not a big deal.

eteq:
👍  to putting in the absolute links. I think that's fine here because there really is an "authoritative" version, so there's at least an argument that that's the right thing regardless of the technical limitations.

eteq:
Actually, an alternative struck me: have the index be auto-generated in javascript.  Something like this:

http://eteq.github.io/astropy-APEs/ (which has the real script [here](https://github.com/eteq/astropy-APEs/blob/gh-pages/scripts.js))

Of course in its current state that's rather ugly, but I could easily clean that up if we want to go that route?

taldcroft:
Well that is pretty fancy...  I was hoping to have the table be visible front-and-center within the README.rst.  Can you do this with your idea?  I tried doing raw HTML earlier in the RST and GitHub did not render it.  I personally think this javascript is overkill for a really teeny thing that is easily maintained each time a new APE is accepted (twice a year on average).

eteq:
Hmm, I don't think it's possible to put it in the README due to security concerns (executing javascript in arbitrary files and all... that's what [this SO post](http://stackoverflow.com/questions/19314567/embed-bookmarklet-in-github-readme-md) suggests, anyway).  We could of course put a link that goes straight to the index, though.

I guess I'm less convinced that we'll remember to keep the README up-to-date, particularly if the outcome of #20 is that APE's can be changed... But in general I have a nearly-pathological aversion to manual list-updating of this sort, so I might not be the best opinion here ;)

taldcroft:
It's really no different than CHANGES.rst, but you may have been permanently scarred by being astropy release manager.  But really the activity here is *so* low.

bsipocz:
What about setting up astrobot to check on PRs here, too but rather than checking the changelog it checks the readme?

eteq:
OK, I see your point, @taldcroft - I'll just go ahead and merge this.  We can always re-visit the topic if things start to get disconnected.

PR #22, opened 2017-02-13
Migrate README.rst to README.md

taldcroft:
This solves the issue of relative links in the APEs table by migrating the README from RST to markdown.  I also think markdown is nicer for something like this.


astrofrog:
I'm going to close this as there are some big conflicts so this would need to be redone.

PR #23, opened 2017-04-18
Vision for Astropy Spectroscopic Tools

crawfordsm:
This is a pull request for an APE for development of spectroscopic tools for astropy.  This APE is being submitted ahead of the meeting on this scheduled for 20/21 April and we expect this to change with the outcome of the meeting, but this pull request hopefully provides a space for comments on this topic.

### DIFF
--- a/APE_specutils.rst
+++ b/APE_specutils.rst
@@ -1,0 +1,64 @@
+<APE title here>
+----------------
+
+author: <your name>
+
+date-created: 2013 November 5 <replace with the date you submit the APE>
+
+date-last-revised: 2013 December 17 <keep this up to date anytime something changes>
+
+type: <one of these three: Standard Track, Informational, Process>
+
+status: Discussion
+
+
+Abstract
+--------
+
+The abstract should be a short description of what the APE will achieve.
+
+
+Detailed description
+--------------------
+
+This section describes the need for the APE.  It should describe the existing
+problem that it is trying to solve and why this APE makes the situation better.
+It should include examples of how the new functionality would be used and
+perhaps some use cases.
+
+
+Branches and pull requests
+--------------------------
+
+Any pull requests or development branches containing work on this APE should be
+linked to from here.  (An APE does not need to be implemented in a single pull
+request if it makes sense to implement it in discrete phases). If no code is yet
+implemented, just put "N/A"
+
+
+Implementation
+--------------
+
+This section lists the major steps required to implement the APE.  Where
+possible, it should be noted where one step is dependent on another, and which
+steps may be optionally omitted.  Where it makes sense, each  step should
+include a link related pull requests as the implementation progresses.
+
+
+Backward compatibility
+----------------------
+
+This section describes the ways in which the APE breaks backward compatibility.
+
+
+Alternatives
+------------
+
+If there were any alternative solutions to solving the same problem, they should
+be discussed here, along with a justification for the chosen approach.
+
+
+Decision rationale
+------------------
+
+<To be filled in by the coordinating committee when the APE is accepted or rejected>
--- a/APE_specutils.rst
+++ b/APE_specutils.rst
@@ -31,7 +31,7 @@ Background
 
 A spectroscopic package has been a primary element of the Astropy Project from its inception.  This was borne out by the large number of different spectroscopic packages that were in development in 2011 (when Astropy was started).  Many of these projects had similar, but not quite compatible interfaces (i.e. differences in formats, arguments, and style) with overlapping functionality.   At the same time, some commonly used functionality was not implemented at all or only implemented as scripts (e.g., heliocentric corrections).  At the same time, different users had slightly different priorities and requirements while some basic functionality did not exist in the Python universe.   
 
-The proposed first step of this was to develop a common spectral object and a collection of tasks that has resulted in the [specutils](https://github.com/astropy/specutils) package.  This package contains a `Spectrum1D` object and has the functionality to read and write in a number of common formats for spectroscopic data. However, this package has not been widely adopted, primarily due to limited coordination/communication with other efforts and slower-than-expected progress in developing a generalized WCS that can be used in specutils.  This motivates this APE, as a way to set a clearer path forward for the specutils effort.
+The proposed first step of this was to develop a common spectral object and a collection of tasks that has resulted in the `specutils <https://github.com/astropy/specutils>`_ package.  This package contains a `Spectrum1D` object and has the functionality to read and write in a number of common formats for spectroscopic data. However, this package has not been widely adopted, primarily due to limited coordination/communication with other efforts and slower-than-expected progress in developing a generalized WCS that can be used in specutils.  This motivates this APE, as a way to set a clearer path forward for the specutils effort.
 
 Currently, the Spectrum1DRef object operates ostensibly as a wrapper around specutils’s core Spectrum1D object. It is, in a basic sense, a workaround to avoid the highly specialized implementation details of earlier attempts at a broadly applicable spectroscopy package. In this case, a custom WCS framework had been developed and maintained within the specutils package. The Spectrum1DRef was a compromise between maintaining package compatibility, and re-introducing the more general Astropy-related behaviors (IO, arithmetic, WCS, etc).
 
@@ -52,7 +52,8 @@ This APE proposes a coordinated suite of packages for doing astronomical spectro
 The packages will be developed separately but are connected by a common interface object: `Spectrum1D`. This object will be in a common package - for the near-term, the common package will be `specutils`, but once the interface has stabilized, these interface classes may be moved to the `astropy` core package. Packages should use this common `Spectrum1D` object (described in more detail below) for representing spectra, although they are welcome to sub-class it to add package-specific functionality.  Also note that, in the future, more interface objects might be called for - e.g. `Spectrum2D` or `SpectralCube` or similar.  In previous discussions at Astropy coordination meetings, there has also been suggestions to eventually create domain-specific subclasses like `RadioSpectrum1D` or `XRaySpectrum1D` (with e.g. different default units appropriate for that field).  The intent is for `Spectrum1D` to provide a starting point for such efforts, which will then develop in a similar manner as they are called for by the community.
 
 The foundational tier of packages will consist of three packages: 
-* `[specutils](https://github.com/astropy/specutils)`.  This will provide the basic interface classes, as well as *basic* analysis of spectroscopic data. “Basic” here simply means highly standard pieces like measuring equivalent widths, fluxes, Gaussian/Lorentzian/Voigt fitting, RV corrections, etc.
+
+* `specutils <https://github.com/astropy/specutils)>`_.  This will provide the basic interface classes, as well as *basic* analysis of spectroscopic data. “Basic” here simply means highly standard pieces like measuring equivalent widths, fluxes, Gaussian/Lorentzian/Voigt fitting, RV corrections, etc.
 * `specreduce`.  This will provide a standard toolbox for reducing spectra, roughly akin to the IRAF “specred” functionality.  Note that this should *not* contain any instrument-specific code.  Observatories or instruments are encouraged to use and contribute back functionality to `specreduce`, but should provide local functionality in their own dedicated packages or guides.  This, ideally, would provide the codes to quickly be able to build specific instrument pipelines. 
 * `specviz`. This will provide 1D spectrum visualization capabilities, but is scoped to be focused primarily on that, and not arbitrarily complex spectrum visualization (e.g. data cubes). Note that, unlike the above two, this package should *not* be considered a toolbox that other tools should be derived from.  While other packages are welcome to use it, it is meant to provide a critical piece of functionality, while acknowledging that visualization, GUI code, and even personal taste are variable enough that there is room for a few other visualization packages in the ecosystem.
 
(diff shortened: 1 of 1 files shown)


weaverba137:
It looks like this document is written in MarkDown, but GitHub is trying to render reStructuredText because of the .rst suffix.  This makes the document hard to read on GitHub.  Can you change the suffix or re-write in reStructuredText?

crawfordsm:
If you go to the [view](https://github.com/crawfordsm/astropy-APEs/blob/27064a66093ea4c5a55ef3465472ecc87a597b96/APE_specutils.rst) version of it, does it render better for you?    

though I just noticed there are some formatting errors that showed up that need to be fixed.

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+
+Because of the diverse set of analysis requirements and reduction techniques for astronomical spectroscopy, we do not focus on specific requirements for the reduction or analysis of spectra, but the aspects that are similar regardless of the type of spectroscopy. This functionality will be included in `specutils` so that it does not need to be duplicated in various domain-specific tools.   Some of the common functionality that is required for spectroscopic packages include:
+
+* Identification of spectral lines or other features
+* Measuring features in the spectra including fitting models, calculating line centroids/shapes, and fluxes
+* Transforming the spectra to different unit systems (e.g., converting between different velocity frames, frequency vs wavelength, flux transformations)
+* Measuring velocities for the spectra including cross-correlation
+* deblending of lines
+* Interpolation, either onto different wavelengths (with flux-conserving at least an option) or over bad data points
+* Continuum subtraction or normalization
```
ehsteve: Continuum or background subtraction. I would not consider them always necessary the same as you may want a measurement of the properties of the continuum but the background is usually just removed.

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+
+Spectrum1D will define a specific interpretation of the meaning of a pixel coordinate value (i.e., is it the mean? Left side? Median? In what units).  This sentence is a placeholder; we plan to have a breakout session on April 21 to identify whether there is a standard we can select and finalize this choice.
+
+
+Specifications of Other Things to be included in specutils
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+Because of the diverse set of analysis requirements and reduction techniques for astronomical spectroscopy, we do not focus on specific requirements for the reduction or analysis of spectra, but the aspects that are similar regardless of the type of spectroscopy. This functionality will be included in `specutils` so that it does not need to be duplicated in various domain-specific tools.   Some of the common functionality that is required for spectroscopic packages include:
+
+* Identification of spectral lines or other features
```
ehsteve: Identification of spectral lines could be related to associated region of interests (ROI) which might be objects themselves. A procedure might generate them automatically or they could be defined by hand. An ROI could also be used to identify a suitable region to generate a background model.

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+* Measuring features in the spectra including fitting models, calculating line centroids/shapes, and fluxes
```
ehsteve: In order to fit a model correctly , generally a detector response model is necessary and therefore needs to be associated with the spectrum object.

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+* Transforming the spectra to different unit systems (e.g., converting between different velocity frames, frequency vs wavelength, flux transformations)
+* Measuring velocities for the spectra including cross-correlation
+* deblending of lines
+* Interpolation, either onto different wavelengths (with flux-conserving at least an option) or over bad data points
```
ehsteve: Should add rebinning to this list.

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+* Continuum subtraction or normalization
+
```
ehsteve: Likely want the ability to add spectra together or subtract them. For example many observatories have multiple detectors observing the same object.

astrofrog: Just a heads-up, this proved to be a topic of disagreement/difficulty in the past, since dealing with errors is not trivial etc. Not saying it can't be done, but probably shouldn't be in the base class.

APE_specutils.rst
astrofrog: This is probably something that is wavelength-regime specific - rebinning a radio spectrum is different from rebinning an X-ray spectrum.

weaverba137:
> If you go to the view version of it, does it render better for you?

That's how I was viewing it.  The bulleted lists were not lists, the links were not links.  It's as I said, the document was written in MarkDown, but GitHub is trying to render it as reStructuredText.

bsipocz:
@weaverba137 - maybe try another browser? The view works perfectly for me on firefox, including links, lists, etc.

weaverba137:
The document changed subsequent to my original comment.

bsipocz:
And I have replied to your comment made an hour ago. Steve mentioned above that he fixed a few things.

crawfordsm:
Yup, @weaverba137  I was using the wrong syntax, but hopefully everything should look right now!  Comments/fixes even on structure are still very welcome!

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+type:  Process, Standard Track
+
+status: Discussion
+
+
+Abstract
+--------
+
+Spectroscopy lies at the core of much of astronomy.  It is the goal of this APE to coordinate and plan the development of spectroscopic software within the astropy project.   The requirements for spectroscopic tools include handling and reducing observations, different types of analysis, and visualization of spectra.  The goal of this APE is to provide a framework that different astronomy packages can use to coordinate the development of spectroscopic tools.   The idea is a set of foundational packages that can then be extended to build other tools. As part of this coordination effort, this APE also provides specifications for `specutils` that includes a `Spectrum1D` class, a standard interface for 1D spectra, providing the basis for interoperability, and general purpose spectroscopic tools.  .
```
astrofrog: In general, can you wrap the text to make it easier to do in-line comments?

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+* Dispersion
+* Flux 
+* Errors
+* Units
+* System (ie. heliocentric)
+* meta information about the provenance of the spectra
+
+To ease development, the Spectrum1D object could have energy, wavelength, and frequency properties that convert the dispersion to the appropriate dimensions.  In addition to this information, software should provide a method for reading and writing formats regularly used for spectroscopy.   Common formats include iraf, SDSS, *please list others*, midas, FITS tables, and ascii.   Reading and writing into a 1D spectrum object should use the astropy io register.   In addition, methods should exist for arithmetic on spectra1d objects including how to handle objects with different dispersion sampling.  
+
+Spectrum1D will define a specific interpretation of the meaning of a pixel coordinate value (i.e., is it the mean? Left side? Median? In what units).  This sentence is a placeholder; we plan to have a breakout session on April 21 to identify whether there is a standard we can select and finalize this choice.
```
astrofrog: Just to be clear, Spectrum1D will provide a way to define the meaning of pixel coordinate, or it will actually adopt one?

dkirkby: I don't see any explicit mention of a WCS associated with each `Spectrum1D` object.  Is the idea that no WCS is required?  (which I am in favor of).

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+The `Spectrum1D` class 
+++++++++++++++++++++++++++++++
+
+A spectroscopy package should provide a representation of the data, tools for reducing spectroscopic observations, analysis tools, and methods for interacting and visualization of the spectra.    In this section, we describe some of the functionality that will be required for a common spectroscopic package.   While equally important, we leave the specification of multi-dimensional spectral objects to future work.  At the same time, we should consider how these tools will be compatible with multi-dimensional data in the future. 
+
+A spectrum describes how the flux changes as a function of the energy.   There are a number of other parameters that can be used to describe a spectrum, and a 1D Spectrum object should have the following properties:
+
+* Dispersion
+* Flux 
+* Errors
```
dkirkby: Errors should be specified for both the flux measurement and the dispersion measurement (e.g., wavelength resolution).  In practice, a (sparse) resolution matrix can dominate the size of a spectrum object.

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+* Units
+* System (ie. heliocentric)
+* meta information about the provenance of the spectra
```
dkirkby: Also: per-pixel bitmasks to summarize processing status, e.g. [SDSS SPPIXMASK](http://www.sdss3.org/dr9/algorithms/bitmask_sppixmask.php).

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+
+In both tiers, Packages should aim for the same level of testing and documentation found in the `astropy` package, and should follow the standard Astropy affiliated package process to do this.  This includes, in particular, considering overlap with existing packages - this should be considered and merged when appropriate.  The Spectroscopic Coordinating Committee will assist the Astropy Coordinating Committee with this process.  Developers are encouraged to build or update existing tools to meet their needs rather than creating a new package.  
+
+The `Spectrum1D` class 
+++++++++++++++++++++++++++++++
+
+A spectroscopy package should provide a representation of the data, tools for reducing spectroscopic observations, analysis tools, and methods for interacting and visualization of the spectra.    In this section, we describe some of the functionality that will be required for a common spectroscopic package.   While equally important, we leave the specification of multi-dimensional spectral objects to future work.  At the same time, we should consider how these tools will be compatible with multi-dimensional data in the future.
```
dkirkby: An important use-case for large spectroscopic surveys (SDSS, eBOSS, DESI) is a container of a large number (500-5K) of 1D spectra, each with the same wavelength grid.  In this use case, it is important to keep flux, ivar, etc, in dense numpy arrays (or equivalent) and to support efficient in-place operations and automatic broadcasting.

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+
+The foundational tier of packages will consist of three packages: 
+
+* `specutils <https://github.com/astropy/specutils)>`_.  This will provide the basic interface classes, as well as *basic* analysis of spectroscopic data. “Basic” here simply means highly standard pieces like measuring equivalent widths, fluxes, Gaussian/Lorentzian/Voigt fitting, RV corrections, etc.
```
astrofrog: I think it might be worth saying from the start that the aim is that this should eventually be moved to the core package? (in my opinion this will be needed for widespread adoption)

APE_specutils.rst
```diff
@@ -0,0 +1,147 @@
...
+
+1. Do nothing.  Pro:  This requires no additional work or collaboration.   Con:   We will continue to redevelop the same set of existing tools and waste limited resources in the community.  There will also be in clear, established way of teaching how to develop spectral objects in python. 
+2. Have the Spectrum1D object be a very limited base class and have development for different  discipline/wavelenth ranges happen independently (i.e.  RadioSpectrum1D, OpticalSpectrum1D, XRaySpectrum1D would have their own packages).   Pro:  This is more descriptive and then more familiar for users of a field. Con:  This will likely lead to redevelopment of some tools, and likely lead to less sharing of  tools developed for different domains. 
+3. Developing functional code with no Spectrum1D object. Pro:  Faster development with less overhead.  Con:  Increase difficulty in maintaining common namespace (wave vs. wavelength vs. something else) that can lead to conflicts in collaborating. Also is quite different from other Astropy efforts and Python standards (e.g. PEP8).
```
dkirkby: What parts of PEP8 are you thinking about here?

APE_specutils.rst
dkirkby: Some other items for the list of common functionality that specutils could provide:
* applying a redshift transform from z1 to z2 (need to use correct exponents for wavelength, flux, ivar, ...)
* downsampling by an integral factor (use mean / sum / sqrt(sum**2) appropriately for each column)
* convolution with a filter response to calculate a magnitude or zero point.
* calculate signal-to-noise ratio per unit dispersion over some band.
* co-add exposures taken at different times and/or covering overlapping wavelength regions.

APE_specutils.rst
bsipocz: I guess it refers to the astropy coding style guide (that is very broad): http://docs.astropy.org/en/latest/development/codeguide.html#coding-style-conventions

APE_specutils.rst
crawfordsm: will do on the next version

APE_specutils.rst
crawfordsm: I'm not sure if everything in specutils should be moved to the core, but I would agree that the Spectrum1D object should move into the core at the very least (and probably be done sooner rather than later)

crawfordsm:
Comments on here will be added to the google doc or check the slack channel for the link to directly edit it

APE_specutils.rst
ehsteve: Well I think that some form of rebinning or resampling should likely be provided in the base class but maybe it can be overwritten by subclasses?

adrn:
Something to consider (possibly by vote, possibly to reject outright) is to rename `specutils` to `specanalyze`. 

Some arguments for:
* `utils` doesn't mean anything
* The package might eventually be split into astropy core and a new `specanalyze` package, so why not start with that name
* now is the time to do it if we are going to

Some arguments against:
* There are already some dependencies (?)
* The name is similar to `photutils`

eblur:
I like ``utils`` because it implies that it is a library of utilities and basic tools.  Meanwhile ``analyze`` implies that you're going to be fitting the spectra or extracting some higher level data products.  And that will very much depend on the data you're working with (radio vs optical vs x-ray).  I want basic spectrum class, but for my own science I would need to write a separate library for analyzing the spectra.

eblur:
What about ``spectools`` ?

adrn:
@eblur But aside from the class, the other items in the description of what is in `specutils` are "things needed for basic analysis"? So if the class gets moved into `astropy` core, all that is left are analysis tools...

eblur:
Ah, you're right.  I just have to point out that NONE of the basic analysis tools suggested for ``specutils`` is something we would be able to use for x-ray spectra.

eblur:
(Mainly due to limitations of our instruments, not specutils itself... so not a criticism of specutils!)

crawfordsm:
Closing this pull request in favor of  #28

PR #24, opened 2017-04-20
Adding cookiecutter APE to the list

bsipocz:


eteq:
Ah, good catch.  Merging!

PR #25, opened 2017-05-03
Update APE10 to have higher minimum python 3 versions

mhvk:
Following discussion in https://github.com/astropy/astropy/issues/6000: change the minimum version for python 3 such that the minimum python version will be supported for (most of) the lifetime of any astropy release.

### DIFF
--- a/APE10.rst
+++ b/APE10.rst
@@ -90,12 +90,12 @@ versions:
 ==========    ===============================
 Version       Minimum required Python version
 ==========    ===============================
-v2.0 - LTS    Python 2.7 or Python 3.3+
-v3.0          Python 3.4+
-v3.1          Python 3.4+
-v3.2          Python 3.4+
-v3.3          Python 3.5+
-v4.0 - LTS    Python 3.5+
+v2.0 - LTS    Python 2.7 or Python 3.4+
+v3.0          Python 3.5+
+v3.1          Python 3.5+
+v3.2          Python 3.5+
+v3.3          Python 3.6+
+v4.0 - LTS    Python 3.6+
 ==========    ===============================
 
 Implementation
@@ -210,8 +210,8 @@ There are several benefits to following the plan proposed above:
 
 * We will be able to start using Python 3-only features internally, including
   for example function annotations (e.g., for units), matrix multiplication
-  (e.g., for coordinates; note that this will only be possible once we support
-  only Python 3.5+)
+  (e.g., for coordinates; note that this will only be possible with Python 3.5+,
+  which in part drives that as minimum version for astropy 3.0)
 
 * Since we will need to keep adding Python 3.x releases to the continuous
   integration over the coming years, we will at least be able to remove Python
@@ -256,3 +256,11 @@ feedback (some of which led to minor adjustments particularly regarding the PyPI
 issues).  Because of this apparent support by the community and agreement from
 the coordination committee with the goals of this APE, it was accepted Aug 22,
 2016.
+
+Updates after acceptance of the APE
+-----------------------------------
+In further discussions (https://github.com/astropy/astropy/issues/6000), it was
+noted that one should not aim to support versions of python that will reach end
+of life during the time an astropy version is supported. In particular, python
+3.3 would reach end-of-life only 3 months after the nominal astropy 2.0 release.
+Hence, all minimum versions of python 3 were increased by one minor version.
--- a/APE10.rst
+++ b/APE10.rst
@@ -5,7 +5,9 @@ author: Tom Robitaille
 
 date-created: 2016 July 13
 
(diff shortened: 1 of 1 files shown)


astrofrog:
I'm not sure if we ever made a decision on updating APEs, but I feel like this is a valid case

APE10.rst
```diff
@@ -256,3 +256,11 @@ feedback (some of which led to minor adjustments particularly regarding the PyPI
 issues).  Because of this apparent support by the community and agreement from
 the coordination committee with the goals of this APE, it was accepted Aug 22,
 2016.
+
```
kelle: this explanation is very clear.

mhvk:
@astrofrog, @eteq - shall we merge this change in minimum python version addendum? I made the PR to implement removing python3.3 tests in master -- https://github.com/astropy/astropy/pull/6021

MSeifert04:
Just a minor point: astropy 3.0 isn't a LTS and 3.4 wouldn't reach end of life till 2019 (expected: 2019-03-16). So following the logic in the APE, including the new update, python 3.4 should be supported until astropy 3.2 or 3.3.

That doesn't mean I'm opposed to dropping python 3.4 support earlier (I'm definetly for it!). It's just that nothing in the APE (including update) justifies that.

bsipocz:
As v3.0 is already a python3 only version, we will have tons of code removal for it. If we refactor anyway, being able to use py3.5+ features would be really nice, and py3.7 will come out just before v3.1.

Can we do a new survey anytime soon to estimate how large percentage of our users use py3.4 as their main version?

kelle:
I'd be happy to put up a poll on the FB group. I think there's enough people there to give a robust sampling. I'll discuss with folks today F2F at PyAstro17.

astrofrog:
We can do a poll on twitter too with the following options:

Python 2.7
Python 3.0-3.2
Python 3.3-3.4
Python 3.5+

(since limited to 4 options)

saimn:
As long as Python 2 is supported there is no real gain to drop support for Python 3.3 (except saving CI time), as it is not possible to use the new Python 3 features anyway. So for Astropy 2.0, even if there is no official support for Python 3.3, it should still be usable.

It will be more interesting for the following versions, when Python 2 is no more supported ! Dropping Python 3.4 for Astropy v3.0 seems reasonable, since this version is probably not widely used. I use py3.4 for our unit tests with the current Debian stable, but a new version of Debian should come in a few months with Python 3.5. And the latest Ubuntu LTS also comes with Python 3.5.

bsipocz:
@saimn - The main argument for dropping py3.3 for v2.0 is that it will become unsupported during the lifetime of the release and thus we better don't say that we support it (while it may of course still work).

saimn:
@bsipocz - Sure, and that's a good motivation.

MSeifert04:
> As long as Python 2 is supported there is no real gain to drop support for Python 3.3 (except saving CI time), as it is not possible to use the new Python 3 features anyway.

That's not the whole motivation. 

Python 2.7 still receives bugfixes, while 3.3 does **not** (except for security related ones, ordinary bugfixes have been stopped since **March 9th, 2014**, 3 years ago!!!). So by dropping 3.3 we can remove workarounds for bugs fixed in 2.7 and 3.4+ but not in 3.3.

mhvk:
A poll sounds like a good idea: always good to have some real data!

eteq:
Realized it was easier to just implement my comment, so all is good!

saimn:
BTW Numpy 1.13 will support Python 2.7 and 3.4 - 3.6:
https://github.com/numpy/numpy/releases/tag/v1.13.0rc1

eteq:
@saimn - just to be clear: are you happy with this given https://github.com/astropy/astropy-APEs/pull/25#issuecomment-300113342 / https://github.com/astropy/astropy-APEs/pull/25#issuecomment-300115849

saimn:
@eteq - Yes that's fine for me (sorry if my initial comment was not clear)

eteq:
OK, great. @astrofrog, if this looks good to you, I think this is set to go.

APE10.rst
```diff
@@ -256,3 +258,11 @@ feedback (some of which led to minor adjustments particularly regarding the PyPI
...
+Updates after acceptance of the APE
+-----------------------------------
+In further discussions (https://github.com/astropy/astropy/issues/6000), it was
+noted that one should not aim to support versions of python that will reach end
```
taldcroft: Instances of `python` should be capitalized.

mhvk: Done.

mhvk:
It seems all of you approved the changes to APE10 - so, ready to merge!?

eteq:
All good. Merging, and thanks @mhvk!

PR #28, opened 2017-10-02
Specutils APE

crawfordsm:
This is an updated version of the APE for `specutils`.     We have started a new pull request for a fresh discussion after incorporating  suggestions  from the specutils coordination meeting and comments from the previous pull request, #23.  We believe we have addressed many of the comments that were made, and  incorporated many of the ideas shared.  

Hopefully this meets many of the requirements that people were interested in, but we would be very happy to receive further comments on this version.   

If you have commented, participated in the meeting, or would like to co-sign this APE, please feel free to leave your name in the comments or add it via a pull request. 

Thank you to everyone for your comments and your patience as we created the updated document.


sosey:
Can we add "ASDF" to the "Creating Spectrum1D objects" section as a common format? Many jwst supporting files will be serialized with ASDF and it's used to represent the complex WCS models in JWST output images. 

Also, not specifically mentioned are slitless spectra, though perhaps more suited for detailing in the `specreduce` package, since they require some more specialized handling of the WCS and modeling on the 2D image. Once reduced to a 1D extraction the data should be compatible with `specutils` 1D methods. Whether or not the committee has decided on handling grism spectra entirely as an associated package, it might be nice to mention that here for completeness.

APE_specutil.rst
```diff
@@ -0,0 +1,482 @@
...
+Attributes of Spectrum1D
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+In this context, a spectrum describes how the flux changes as a function of the
+energy of the radiation from some object.  There are a number of other
+parameters that can be used to describe a spectrum, and a ``Spectrum1D`` object
+will have the following properties:
+
+* ``spectral_axis``
+* ``flux``
```
Cadair: Apart from the obvious physical naming, what was the logic behind going for 'flux' and not 'data' here?

We have adopted NDData for a lot of multi-dimensional spectral (and not spectral) work around SunPy so far [1], I would like to see some level of API consistency between `Spectrum1D` and `NDData`, which there appears to be (i.e. uncertainty, meta).

What would people think of aliasing `.data` to `.flux`? 

[1] We did a lot of stuff on 2D+ spectral data (slit spectrograph) this summer see these repos if you are curious: https://github.com/sunpy/ndcube/ https://github.com/sunpy/irispy

crawfordsm: Thanks for the feedback @Cadair -- yes, `Spectrum1D` will inherit from `NDData` so `.data` and `.flux` will be aliased.   I'll add that to the APE.

crawfordsm:
@sosey Thanks for the feedback.   I'll add ASDF to the methods to incorporate and also mention slitless spectroscopy under `specreduce`.   It very well might be better to develop it as an affiliated package but still probably closely developed with `specreduce`

ntejos:
Hi all,

First of all, thanks for developing this. May you clarify how you envision the WCS to work in the context of IFU data. Even though with IFUs you get cubes, you can still get 1-D spectra out of it...

profxj:
Here are my comments and suggested edits to the draft, in order
of importance:

1. Wording that states usage of the WCS to access the spectral_axis
will be fast, efficient, or however you wish to word it.
1.  The Spectrum1D will be able to hold multiple 1D spectra, either with
an identical spectral_axis or multiple (as the user desires).  Here is an example of what I mean: https://github.com/profxj/specutils/blob/strip_wcs/docs/nb/Fiddling_about.ipynb
1.  The uncertainty property should be optional

jehturner:
Thanks for writing this. I think this looks great in general! My main reservation is that I'll probably need to work with 2D spectra at least as soon as 1D, making it difficult to buy into this scheme 100%. In some sense, 2D spectra are usually a precursor to the (more processed) 1D case, making this a bit of a cart-before-the-horse solution for someone writing a DR package.

I note in the text "This APE is not concerned with representations of data cubes, times series of spectra, or similar objects with multiple spectral axes in the same object". However, a typical data cube will *not* have multiple spectral axes; rather, it is a collection of 1D spectra (as opposed to, say, a collection of image slices prior to resampling). To me, the *main* case not covered by the APE is mixed spatial and spectral co-ordinates on the detector in 2D, such as a slit spectrum. I am not 100% convinced that this needs a separate object from Spectrum1D -- but supporting a set of 1D spectra is certainly a good start.

Some of the specreduce functionality is also going to apply to 2D as-is, eg.  tracing a spectrum but not integrating over its cross-section. The assumption of being able to work with 1D spectra is generally going to impose resampling, which may not be appropriate or viable for some datasets.

I am wondering why there is no `mask` attribute in this proposal, especially when the text discusses interpolating over bad data points? Is this an NDData sub-class like CCDImage, BTW, and if not why not?

I was initially concerned that there isn't a unique way of evaluating the wavelength as a function of pixel index (with spec.spectral_axis.value and spec.wcs.pixel_to_world competing for this role), but after considering your examples, I think this scheme could work well if documented & used appropriately. I wonder whether assigning a spectral axis should also generate a simple interpolated-table WCS, so the .wcs interface is always working when the necessary information is available? I appreciate that people probably wanted to avoid making assumptions, but any WCS will be limited by the accuracy of the available information and having no .wcs may lead to interoperability pitfalls, as well as a level of inelegance similar to having no .spectral_axis and using the WCS, eg. with np.arange(), in its place.

The text says that "at the minimum, the user will have to provide a flux ... and the spectral_axis" but then example 3 seems to contradict this by not specifying spectral_axis; presumably it means "will have to provide enough information to specify the spectral_axis".

Could the name spectral_axis be a bit confusing? I tend to assume it's a number telling you which axis is the spectral one (where that makes sense for 1D).  Also, is there a danger that some grad student might think spec.velocity measures a velocity from the data?? :-). Maybe not if they had to provide it, though a qualifying noun might avoid any doubt.

I hope that's somewhat helpful. Thanks again.

crawfordsm:
@profxj --  I've added the following:

(1) In the wcs section, I've added:The ``wcs`` property should
provide seamless and efficient access to the `spectral_axis``. 

(2) We did try to state this in the `[Dimensionality](https://github.com/crawfordsm/astropy-APEs/blob/specutils_take_2/APE_specutil.rst#dimensionality)` section.    As long as they share the same spectral_axis then the Flux parameter can hold multiple spectra.  

(3) I've made it explicit that `Uncertainty` is optional.

crawfordsm:
@ntejos IFUs and data cubes are outside the scope of this APE.   Nonetheless, we hope that this APE would help inform people writing their code to provide an interface between the different data types.    Although in cases with a common `spectral_axis`, we do hope the `Spectrum1D` object will be useful.

There is work on WCS and datacubes, and it would likely be helpful to those groups if you are interested in participating to develop those requirements.   If you are interested, I think @eteq can provide some more information.

crawfordsm:
@jehturner Thanks for the good points.  We decided to limit the scope for this APE to a `Spectrum1D`.  It sounds like you have a lot of good ideas about what is needed for the data reduction package and I'd highly encourage you to start another APE outlining those in detail.   It will hopefully be helpful to start filling up the [specreduce](https://github.com/astropy/specreduce) repository.

As for`` masking``, due to the current ``NDData`` discussions we did not end up including it.  However, I've now added a line indicated that as such under 'Meta and other attributes' 

It is expected this will inherit from NDData thought that design detail is not explicitly mentioned. 

I've clarified the line to indicate either a `spectral_axis` or a `wcs` can be provided to initiate it.  

I think you raise some good points that it will be important to make sure that the functionality is clearly documented so as to limit any confusion.    Any suggestions would be welcomed,  of course.

eteq:
@profxj - on your 2 from https://github.com/astropy/astropy-APEs/pull/28#issuecomment-335244426 : can you say a bit more why you think it's critical to have an option of *different* spectral axes in a single spectrum 1D? At that point it's not really "1D". It also makes it harder for users of the class to write code that supports it, because in practice they'll have to account for a lot of cases.

That said, I'm open to the idea but think we'll need to add some words justifying/explaining it if we're going to do that.

profxj:
The motivation is two-fold:  

1. convenience -- I assure you that it has been extremely convenient to hold one, two, 100, 1000 spectra in a single object
1. performance -- I shudder at the thought of instantiating 1000 astropy.Spectrum1D objects!

And as I think more deeply on your question, I see that we find ourselves back
at the WCS vs. data array 'argument' (i.e. Issue #159, https://github.com/astropy/specutils/issues/159).

It has come to me now, however, that there may be a good compromise:

- Spectrum1D, mainly as it is defined in the current draft (yes, with WCS)
- Spectra1D -- A lite-weight object to contain *infinite* numbers of 1D spectra designed in a fashion similar to [XSpectrum1D](https://github.com/linetools/linetools/blob/master/linetools/spectra/xspectrum1d.py) or @dkirkby 'speclite'   

The two would be tightly coupled.  Thoughts?
The APE would need to be modified accordingly, of course.

astronomeara:
(long time listener, first time caller)
I would second the need for the ability (and utility) to hold multiple spectra in one object, so I would definitely support the compromise position of having Spectra1D be the 'box' for individual Spectrum1D.  Could then be used in a variety of ways, including the IFU question from ntejos.

astrofrog:
> Spectra1D -- A lite-weight object to contain infinite numbers of 1D spectra designed in a fashion similar to XSpectrum1D or @dkirkby 'speclite'

I might advocate for ```Spectrum1DCollection`` as a name - but in any case, based on some initial experiments with implementation, it may be that the spectral axis functionality would come from a mixin class that can be added to a dataset of any dimensionality, and that would make it easy to implement a class representing a collection of spectra.

eteq:
@profxj - I like your compromise suggestion!  I think I also favor @astrofrog's name suggestion, but I think that's probably a detail that's not critical to finalize right now.

The challenge here is that down the road we will likely also want an `IFUSpectrum` or similar that supports more complicated WCS, and it'll be a challenge to try to keep this functionality from overlapping in somewhat confusing ways.  But I see @profxj's point that there is a need *now* for the "collection" form, so we can't ignore that.

crawfordsm:
I think I might suggest `SpectraArray`, but that idea does seem a reasonable compromise.   

I could imagine it isn't too much of a step to generalize `spectral_axis` to have multiple dimensions to match the dimensions in the flux array.

profxj:
Crisis averted.  ;)

ivvv:
Late arrival... We did similar implementation from scratch for Herschel data, but in Java. Some 20 developers worked on this project at the most active development phase of the Herschel Common Software System. I see a lot of synergies , so maybe checking our [spectrum API](http://herschel.esac.esa.int/hcss-doc-15.0/load/hcss_drm/api/herschel/ia/dataset/spectrum/package-summary.html) would be helpful and could bring some new ideas.

We have Spectrum1d class, SpectrumContainer, even Spectrum2d... and lots of methods. 

Note that with Herschel we had 3 spectrometers: one IFU (PACS), one heterodyne spectrometer (HIFI) and one Fourier-Transform Spectrometer (SPIRE). All of then had different data structures for spectra, so we had to put them somehow within the same "Common Software System". 

I am happy to contribute to the efforts for specutils API, as much as I can.

profxj:
You lost me at 'Java'.  ;)

ivvv:
@profxj Well, it was a managerial decision many years ago, and more java developers were available at that time.  I can also see another point of frustration, using _javadoc_ for API documentation... it took me years to get used to it :-)

profxj:
Let's just say that you are going to need to provide your input
directly, ie. I doubt many of us are going to scan through the
code (or docs) to glean insight..

astrofrog:
@profxj - actually the link that @ivvv provided (thanks!) is a pretty readable API doc:

http://herschel.esac.esa.int/hcss-doc-15.0/load/hcss_drm/api/herschel/ia/dataset/spectrum/package-summary.html

I do think it would make sense for the APE authors here to take a look at this. Clearly a lot of thought went into that.

lemmatum:
This looks quite interesting and should have substantial overlaps with our project [SpectroscoPy](https://github.com/PlasmaPy/SpectroscoPy). Though I think we will be more dependent on 2D data with either spatial or temporal information, I would be happy to collaborate and trade around useful functions and whatnot.

Some questions/comments that came up when looking through the APE_specutils.rst:
* Will the HDF5 format be included for reading/storing spectral data?
* I noticed mention of wavelength and flux calibration, does this include calibration with respect to a known whitelight source? (in the lab we use tungsten lamps cross-calibrated by NIST, I assume in astronomy you use a star or some other point-like object with a "known" blackbody temperature?)
* There was mention of continuum subtraction, does this include baseline estimation for absorption spectra or are you focusing on emission spectra?
* You mention identification of spectral lines, one way to achieve this would be through using wavelet transforms for peak identification (not sure how common this is in astronomy).
* I agree with the comment about including a detector response model. Would you include instrument function convolution and deconvolution within such a model? It would be useful to place dispersion and whitelight calibration tools under the same model. You could add to the model with things like detector quantum efficiency, pixel size, lines/mm on a grating, to create approximate calibrations in lieu of having a measured whitelight and dispersion calibration.
* Another common data reduction requirement you might add is transforming a spectrum between different models. For example, transforming between transmission and opacity as the y-axis by using a baseline estimate or an independently measured backlighter spectrum. Another transformation would be flipping an absorption spectrum to overlap and compare to an emission measurement, especially important for comparing broadening of lines.

eteq:
@ivvv - at risk of over-generalizing, I'm actually happy to see how similar the `Spectrum1D` interface in this APE looks to be to the Herschel API.  With that in mind, if you have some *specific* feedback on the `Spectrum1D` in this APE and some lessons you may have learned in Herschel that can be applied to that, please let us know.

Some background you might not know: the reason for this APE is because there were efforts to build a "do it all" API like the Herschel one (but even *broader*, since it's at least focused on one mission), and part of the point of this APE is to bring the scope down to something we all know is practically needed by a wide swath of the community (a 1D spectrum and associated standard analysis tools). So the intent is expressly *not* to build a "do it all" API right now because we want to have something to build from.

But having looked at the Herschel API, there's definitely some great ideas there that we might want to use as the `specutils` API expands beyond the `Spectrum1D` realm. Unless I'm mis-understanding something, I don't think there's anything of that architecture which can't by added on post-hoc once the `Spectrum1D` is up-and-running.

eteq:
@lemmatum 
> Will the HDF5 format be included for reading/storing spectral data?

This APE is meant to be agnostic about on-disk format.  The idea is to use the astropy unified i/o machinery to allow anyone to register a reader/writer from a `Spectrum1D` to whatever there favorite format is.  There are unfortunately a *lot* of "standards" in the form of fits (or in a few cases hdf5 files) that store their metadata in different ways, so the goal is to support people writing readers/writers as needed, but not call a specific one "correct" (at least right now).

> I noticed mention of wavelength and flux calibration, does this include calibration with respect to a known whitelight source? (in the lab we use tungsten lamps cross-calibrated by NIST, I assume in astronomy you use a star or some other point-like object with a "known" blackbody temperature?)

Wavelength calibration is usually done with arc lamps in the spectrograph, while flux calibration is done with "known" celestial sources like stars. There are standard procedures in astro for this that the goal is to replicate (i.e. approaches taken by IRAF or the SDSS), with more cutting-edge algorithms to potentially come later.

> There was mention of continuum subtraction, does this include baseline estimation for absorption spectra or are you focusing on emission spectra?

Both, although again `specutils` is more meant to be a "toolbox" for this than an "application" (there are about as many opinions on the "right" way to remove continuum as there are astronomers doing spectroscopy...)

> You mention identification of spectral lines, one way to achieve this would be through using wavelet transforms for peak identification (not sure how common this is in astronomy).

Some basic peak-identification algorithms are definitely in-scope, although we haven't talked a ton about exactly which. Wavelet transforms are not particularly common, which makes it out-of-scope for the baseline tools that the APE is focused on. *But* it would be a welcome contribution!

> I agree with the comment about including a detector response model. Would you include instrument function convolution and deconvolution within such a model? It would be useful to place dispersion and whitelight calibration tools under the same model. You could add to the model with things like detector quantum efficiency, pixel size, lines/mm on a grating, to create approximate calibrations in lieu of having a measured whitelight and dispersion calibration.

Yes, this is all in scope for `specreduce`, which is *intentionally* not outlined in detail in this APE.  But definitely there needs to be some architecting in specreduce to cover these things, as they are standard things done by typical data reduction pipleines.

> Another common data reduction requirement you might add is transforming a spectrum between different models. For example, transforming between transmission and opacity as the y-axis by using a baseline estimate or an independently measured backlighter spectrum. Another transformation would be flipping an absorption spectrum to overlap and compare to an emission measurement, especially important for comparing broadening of lines.

This is a great *use case* that should be enabled, so if you think something in teh `Spectrum1D` architecture prevents it, that would be important to know!  But as I see it that's more, again, "application" (or "user") level, so we should get the low-level pieces done first. Then once we have some people actually using it we'll be better positioned to know what high-level tasks should be provided.  Make sense?

lemmatum:
@eteq thanks for the thorough response and clarifying where you are distinguishing between toolbox/framework and applications.

> Wavelength calibration is usually done with arc lamps in the spectrograph, while flux calibration is done with "known" celestial sources like stars. 

This should be readily translatable to Hg/Ar pen lamps and tungsten whitelight sources for optical spectra. X-ray spectroscopy calibrations may be a bit more complex. Do you need to account for nonlinear detector responses to large fluxes in astronomy?

> there are about as many opinions on the "right" way to remove continuum as there are astronomers doing spectroscopy...

Preach! And there will always be that one physicist in the back that doesn't like your results because you didn't use their favorite continuum estimation method.

> This is a great use case that should be enabled, so if you think something in teh Spectrum1D architecture prevents it, that would be important to know!

I don't see any obvious conflicts... the basics of taking 2 spectra, dividing them and applying some simple math functions should be easy to do within the framework you've outlined. More sophisticated analysis to correct for experiment/observation specific effects would have to be on the user end. Though this does bring up another question - will error propagation be built into the framework? (propagation of calibration errors is not so fun)

crawfordsm:
@profxj I've finally took a moment to add a section on multi-dimensional data that I think would cover your use case.  If you aren't happy with it, please suggest some text and I'll replace what is there

crawfordsm:
@lemmatum this would handle error propagation as long as that was type of propagation was implemented.  For right now, standard deviation error propagation has already been implemented in NDData, but anything more complicated would  still have to be written.

ivvv:
Thanks @eteq, I've read the current APE_specutils.rst and the current ideas for the `Spectrum1d` class seem good with few comments from my side:
* I see  masking in `Spectrum1d` is mentioned in the metadata and attributes. In _Herschel_ we have a `flag` column with the same size as the `spectral_axis`.  I guess `NDData` allows similar functionality via masking, but my feeling is that this needs to be specified more clearly. Also, masking is binary 0/1, so maybe an optional `flag` (or `quality`) column allowing for encoding some additional information rather than 0/1 could be incorporated. E.g. `flag=0` good channel, `flag=1` noisy channel, `flag=2` telluric line region, etc...
* I do not see a mention if the `spectral_axis` only allows unique values, i.e. no repeated values. In this sense, an optional spectral `segment` column could be useful, i.e. `spectral_axis` should contain unique values for the same spectral segment. This also could cover the usecase of merging/stitching two or more `Spectrum1d` spectra into one `Spectrum1d`. It's to be defined how the WCS (ra,dec) of the two segments would be propagated to the final `Spectrum1d`.
* On `uncertainty` column, one of the _Herschel_ instruments (HIFI) use `weight` instead of `error` or `uncertainty`. Usually we assume `weight = 1/error` or `1/error^2`, but this may not be the case for some spectral data. I guess we should leave this to the user to convert their `weights` to `uncertainties`, as otherwise propagating the `weights` is different and needs additional coding.

I hope these comments are not too much off-topic.

profxj:
@crawfordsm -- This looks fine although maybe it shouldn't
state the multi-dimensional object will have the *same* attributes.
Implicitly, I read it as having *all* of the same attributes.

Anyways, this is a minor quibble..

Thanks for this.

crawfordsm:
@ivvv Thanks definitely relevant comments

* We left masking and flag a little bit vague as there is another ongoing discussion about how to implement them exactly in NDData.   But that is meant to indicate they will be there.

* This is interesting and will have to think/read more about.   Would like to see more details if you have a link to something describing it in more details

* Yes, the `uncertainty` class is there so that whatever is passed and used, it is clear what is being passed.

APE_specutil.rst
```diff
@@ -0,0 +1,496 @@
...
+it may seem to conflict with the “1D” in the class name, this name scheme is
+meant to communicate the presence of a single common spectral axis.  For more
+on this, see the section on data cubes below.
+
+Metadata and other attributes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``meta`` attribute is meant for metadata, e.g., information from
+a FITS header.   The ``Spectrum1D`` object will also have a method for
+masking spectrum that will be defined in a similar manner as in  ``NDData``.
```
crawfordsm: add flags here

eteq:
Procedural note: this has been open for a while with lots of great feedback that @crawfordsm has now integrated into the APE.  We're going to set a deadline for the end of this week (Dec 15) for feedback, after which we will ask the Coordination Committee (sans me as a co-author) to rule on this one.

kelle:
Consider adding this key point to the Abstract:  The “1D” is meant to communicate the presence of a single common spectral axis. It can be collections of spectra that share the same spectral_axis.

eteq:
Good point, @kelle! Just pushed up a commit with that clarification.

PR #29, opened 2017-10-20
Add zenodo upload instructions

eteq:
This adds an additional step for "accepting" in the readme, to register the APEs with Zenodo.  

cc @bsipocz @astrofrog


README.rst
```diff
@@ -75,3 +75,4 @@ or rejection.  One of the committee members should then:
 3. If the APE is accepted, add a commit to the APE which puts the APE into the
    "Accepted APEs" table of the repository's README.
 4. Merge the PR with the above changes.
+5. Upload the APE to Zenodo to give it a DOI. Go to https://zenodo.org/deposit/new, upload the .md file, Set the title as "APE ##: xxx", paste the abstract in for the "Description", enter in the authors of the APE (with ORCIDs if possible), leave the License as CC-Attribution, and under "Related/alternate identifiers" paste in the github link to the APE file *at the specific merge commit* (e.g. https://github.com/astropy/astropy-APEs/blob/42951733ac42c0ea178d8df30705274a43c93091/APE1.rst) as "is supplemented by this upload".
```
astrofrog: github -> GitHub

astrofrog: also upload the .rst file? (not md?)

bsipocz: If there is an annex added to the APE, that should receive a separate DOI, or I guess it's enough to only do it once for the first merge?

eteq: Actually Zenodo now supports "versioned" DOIs, so @astrofrog was thinking we use that for changes.  Should add something about that now that you mentioned it, though.

README.rst
```diff
@@ -75,3 +75,22 @@ or rejection.  One of the committee members should then:
...
+
+Updating APEs
+^^^^^^^^^^^^^
+
+In the cases where an updated APE requires updating (e.g. references to a  new
+APE that supercedes it, clarifying information that emerges after the APE is
+accepted, etc.), changes can be made directly via PR, but the
+"date-last-revised" should be updated in the APE. Additionally, the Zenodo entry
+will need to be updated with a new version of the APE (*not* a completely new
+Zenodo entry).
```
astrofrog: So the way this works is that there should be a button 'add new version of' or something like that. We should upload an APE then see if there is something like that.

astrofrog:
@eteq - one last comment: set the publication date to accepted date?

README.rst
eteq: There is one - I saw this already on something else I Zenodo-uploaded. Will use the exact wording so it's clear what that looks like.

eteq:
> one last comment: set the publication date to accepted date?

I'd say it's the last modified date - for revisions that's when that *revision* was published, and presumably that's the same as the accepted date for non-revised APEs.  But definitely should add that.

eteq:
@astrofrog - OK, I think I've addressed your comments. If it looks good to you now feel free to merge!

astrofrog:
Looks good! I'm going to upload one as a test now.

PR #31, opened 2017-11-07
WIP: A shared Python interface for World Coordinate Systems

astrofrog:
This is a WIP and not yet ready for review

Human-readable version: https://github.com/astrofrog/astropy-APEs/blob/ape-wcs/APE13.rst

### DIFF
--- a/APE13.rst
+++ b/APE13.rst
@@ -140,9 +140,11 @@ low-level API recommends:
         def world_axis_physical_types(self):
             """
             Returns an iterable of strings describing the physical type for each
-            world axis. This should follow the VO names from the UCD1+ controlled
+            world axis. They should be names from the VO UCD1+ controlled
             Vocabulary (http://www.ivoa.net/documents/latest/UCDlist.html).
-            If no matching VO type exists, this can return None for any or all axes.
+            If no matching UCD type exists, this can instead be "custom:XXX",
+            where XXX is an arbitrary string.  Alternatively, if the physical
+            type is unknown/undefined, an element can be `None`.
             """
 
         @property
@@ -293,8 +295,8 @@ dimension encoding time-of-observation.
 Common UCD1+ names for physical types
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-As outlined above, the ``world_physical_type`` attribute should include straings
-that fullow the VO UCD1+ vocabulary for defining physical types. The
+As outlined above, the ``world_axis_physical_types`` attribute should include
+strings that fullow the VO UCD1+ vocabulary for defining physical types. The
 `full UCD1+ vocabulary <http://www.ivoa.net/documents/latest/UCDlist.html>`_
 includes a large number of options, but here we summarize some of the most
 common ones that will likely be used::
@@ -327,10 +329,10 @@ common ones that will likely be used::
 
 The full UCD1+ vocabulary does not include all possible type names that would be
 needed to represent **all** WCSes (for example, there are no keywords for
-helioprojective coordinates). In this case, ``world_axis_physical_types`` should
-either return ``None`` for those coordinates, or return a string prefixed with
-``custom:``. This should also be taken as a call to work with the International
-Virtual Observatory Alliance (IVOA) to implement
+helioprojective coordinates). In this case, the element of
+``world_axis_physical_types`` for those coordinates should be a string prefixed
+with ``custom:``. This should also be taken as a call to work with the
+International Virtual Observatory Alliance (IVOA) to implement
 new type names, which the Astropy Project will facilitate as needed. If a
 ``custom:`` type name is needed, we recommend that these be coordinated and
 agreed as much as possible between different packages to make sure that these
--- a/APE13.rst
+++ b/APE13.rst
@@ -79,7 +79,7 @@ and should be *compatible* with this APE’s interface, an API for these
 operations is out-of-scope for this APE. This APE also does not address how to
 *create* such WCSs. Rather it is aimed specifically at an API for the simpler
 problem of actually performing the “end-to-end” transformation from pixel to
-some world space. Overview of the proposed WCS Interface
+some world space.
 
 Our proposal in this APE is the following:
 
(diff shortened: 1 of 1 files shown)


APE13.rst
```diff
@@ -0,0 +1,409 @@
...
+    wcs.axis_correlation_matrix = [[True]]
+    wcs.world_axis_units = [‘pixel’]
+    wcs.world_axis_physical_type = [‘instr.pixel’]
+    wcs.world_axis_object_components = [(‘spec’, 0)]
+    wcs.world_axis_object_classes  = {‘spec’:(‘astropy.units.pixel’: {})}
+
+Common UCD1+ names
+^^^^^^^^^^^^^^^^^^
+
+The full UCD1+ vocabulary (used for the ``world_physical_types`` attribute)
```
Cadair: How can you represent spatial coordinate systems not explicitly in the UCD1+ list? Like any of the solar coordinate systems and quite a few others.

APE13.rst
```diff
@@ -0,0 +1,409 @@
...
+High-level Astropy Object
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Unlike the low-level API, the ‘high-level’ interface described here will be a
+single Astropy-developed class since it interfaces with various Astropy objects.
+This high-level API would provide the ability for example to get ``SkyCoord``,
+``Time`` etc. objects back from a pixel to world conversion, and conversely to
+be able to convert ``SkyCoord``, ``Time`` etc. to pixel values.
+
+The high-level object would not inherit from the low-level classes but instead
```
Cadair: Should the high-level object drop the ``pixel_to_world_values`` and ``world_to_pixel_values`` methods?

astrofrog: The high-level object won't have the low-level methods - instead these will be available as ``w.low_level_wcs. pixel_to_world_values`` for example (see below for a mention of ``pixel_to_world_values``

APE13.rst
astrofrog: This is a valid question - if we decide to go ahead with UCD1+ then we can either just work with the IVOA to add other names as needed and just return None in the mean time, or we can come up with a list of additional acceptable attributes. While frustrating because things might be slow, I think the former would be better.

A complete alternative would be to use much more general terms such as celestial, spectral, etc. and just come up with the list ourselves (basically with terms that align with the STC standard)

@eteq @keflavich - any thoughts?

APE13.rst
Cadair: Ok, so that's not how I read this section. Is this section then saying the only methods required on high level are `world_to_pixel` and `pixel_to_world`?

astrofrog: Yes - bear in mind the high level object is not an API specification but an actual object. We could definitely add more to it over time if we want. This is just the minimum.

Cadair: ok, I would say that needs clarifying a little.

astrofrog: @Cadair - can you check my latest changes in e31fd0f9d63f71a4f8e0281a978f4e11309d0d3a

Cadair: looks good

APE13.rst
astrofrog: @Cadair @eteq @keflavich - see 362b28d0f9dd6c269ac1ab45ce2002fed45f97e5

APE13.rst
```diff
@@ -0,0 +1,431 @@
...
+    wcs.world_axis_physical_type = [‘instr.pixel’]
+    wcs.world_axis_object_components = [(‘spec’, 0)]
+    wcs.world_axis_object_classes  = {‘spec’:(‘astropy.units.pixel’: {})}
+
+Common UCD1+ names
+^^^^^^^^^^^^^^^^^^
+
+The full UCD1+ vocabulary (used for the ``world_physical_types`` attribute)
+includes a large number of options, but here we summarize some of the most
+common ones that will likely be used::
```
eteq: I think you should link to the official UCD1+ list somewhere here (http://www.ivoa.net/documents/latest/UCDlist.html) - it's really hard to find without a direct link!

APE13.rst
```diff
@@ -0,0 +1,431 @@
...
+Detailed description
+--------------------
+
+Background/Terminology
+^^^^^^^^^^^^^^^^^^^^^^
+
+One of the key basic operations necessary for any astronomical data analysis is
+the association of “pixels” with their “real-world” coordinates (and vice
+versa). While this concept has existed essentially for as long as computers
+have been used on astronomical data, the FITS WCS standard (Calabretta &
```
eteq: Maybe link this to the paper? (http://dx.doi.org/10.1051/0004-6361:20021327)

APE13.rst
eteq: I think that new change looks good.  But maybe you should add this to the list of "alternative", @astrofrog?  Also I have some minor word-smithing changes but can just do that myself.

APE13.rst
astrofrog: I added a link in the docstring higher up but yeah we need a link here too!

eteq:
@astrofrog - in astrofrog/astropy-APEs#11 I think I've implemented all of my comments here.  If that looks good I think I'm happy here aside from one question:

Maybe it should be allowable to use a custom string instead of only `None` for non-UCD keywords?  It might be we should mandate a prefix like `nonUCD:XXX`, but still, it seems like there should be a way for a user to say "this has a known physical type, just not one the IVOA thought of".

astrofrog:
@eteq - I could live with ``nonucd:`` (all lowercase since all UCDs are lowercase too)

Cadair:
Maybe something like `custom:<ucd like grammar>` rather than `nonucd`?

astrofrog:
Yeah I was almost going to suggest that, so ``custom:`` sounds good to me. @eteq @keflavich?

eteq:
`custom:` is good to me too.  I think there needs to be a bit more clarity on `None` vs `'custom:'` though, and went ahead and created astrofrog/astropy-APEs#12 for that.  If that looks good to you @astrofrog, I'm happy to have this move forward with that change.

PR #33, opened 2017-11-08
Make dates consistent between APEs

astrofrog:
Some APEs were missing accepted dates. I would like to suggest that changes such as this don't warrant a revision update though.


eteq:
LGTM. I think the policy is basically that an "update" requires that the text be modified, not just the metadata

PR #34, opened 2017-11-15
APE: A shared Python interface for World Coordinate Systems

astrofrog:
**Abstract:** Astronomical data are often provided with information about the “real-world” coordinates that correspond to pixel coordinates. This mapping is the essence of the “World Coordinate System (WCS)” concept. In addition to FITS WCS (the origin of the “WCS” term), other WCS standards/representations are becoming necessary for new missions and observatories (e.g. the James Webb Space Telescope or the Large Synoptic Survey Telescope). In order to build functionality in Astropy and affiliated packages that may require use of WCS, having to deal with different kinds of WCS objects with different APIs will be essentially impossible. Thus, the purpose of this APE is to define a standardized interface to WCS objects based on simple objects (strings, scalars, and arrays). Packages implementing WCS objects will then be able to either modify their classes to conform to the API recommended here, or build thin wrappers that conform to the API. The APE also provides a recommendation for a high-level object to be implemented in Astropy that understands familiar Python objects such as ``SkyCoord``, ``Time``, and other Astropy classes.

**Comments requested** by December 4th 2017


cdeil:
One long-standing debate is whether the data size (array shape) is part of WCS or not. In the Astropy issue tracker it's an open question: https://github.com/astropy/astropy/issues/4669

From a quick look you're not mentioning the data size (array shape) here. Would it make sense to say something? E.g. "decided it's not part of the WCS object / interface we define here" or "is a separate issue, not discussed / decided here"?

On the Gammapy side, we have some interest in this. @woodmd developed `gammapy.maps` with the goal to support N-dim maps with two sky axes, and an arbitrary number of extra axes, e.g. energy, time, but also other things. For now we have [gammapy.maps.WcsGeom](http://docs.gammapy.org/en/latest/api/gammapy.maps.WcsGeom.html) which uses `astropy.wcs.WCS` for the two spatial axes, then has the data size (array shape) added, and then a list of [gammapy.maps.MapAxis](http://docs.gammapy.org/en/latest/api/gammapy.maps.MapAxis.html) for the non-spatial axes. The code is very recent and the design and code isn't complete; specifically we're still discussing how to interact with coordinates, partly looking at http://xarray.pydata.org/ as an example and maybe trying to get to named dimensions in the interface, i.e. "lon", "lat", "energy" instead of integers. At the moment the `WcsGeom` interface does support `SkyCoord`, but not `Time` objects.

I hope this comment / example what we do was a little useful and not too much off-topic. I'll try to find time to read the APE and understand it better and come back to you with more detailed feedback / questions. @woodmd - maybe you have time, too?

APE13.rst
```diff
@@ -0,0 +1,450 @@
...
+    wcs.world_axis_object_classes  = {‘spec’:(‘astropy.units.Wavelength’:
+                                              {‘airorvacwl’: ‘air’})}
+
+**Simple 2D image mapping** where the axis are just RA and Dec (in FITS-WCS this would be CAR)
+
+.. code-block:: python
+
+    wcs.axis_correlation_matrix = [[True, False], [False, True]]
+    wcs.world_axis_units = [‘deg’, ‘deg’]
+    wcs.world_axis_physical_type = [‘pos.eq.ra’, ‘pos.ra.dec’]
```
timj: Typo here: `pos.ra.dec` shpould be `pos.eq.dec`.

APE13.rst
```diff
@@ -0,0 +1,450 @@
...
+Branches and pull requests
+--------------------------
+
+N/A
+
+Implementation
+--------------
+
+The low-level API could be implemented specifically for FITS-WCS into the
+Astropy core package, along with the more general high-level Astropy object.
```
cdeil: Could the low-level API be added to the existing `astropy.wcs.WCS` class? Or probably a new class wrapping it would be needed? I.e. then one would end up with this, no?
```
wcs_high_level.wcs_low_level.wcs_legacy
```
Suggest to clarify that point.

Jerry-Ma:
Referring to #2839, are we to reimplement the WCS module, e.g, based on `gwcs` in the entirety and get rid of wcslib dependency?

cdeil:
So the coordinates you interact with e.g. in `pixel_to_world_values` are tuples or lists and rely on the caller getting axis order right always? There's no interface to interact with coordinates using dimension names like [here](http://xarray.pydata.org/en/stable/indexing.html#indexing-with-labeled-dimensions)? Personally I like named dimensions and using the names in the API when taking or returning coordinates a lot. But I also realise it's an extra and non-trivial to support in addition to or instead of the tuple-based interface. Worth discussing?

astrofrog:
@Jerry-Ma - for now there are no plans to merge gWCS into the core package and re-write astropy.wcs. If this did happen it would be quite a long-term thing. The point of the APE here is to make it so that people developing functionality that relies on WCS doesn't need to worry about whether it's astropy.wcs or gWCS, and if we did refactor astropy.wcs in future we could still provide the same API (described by the APE here).

APE13.rst
astrofrog: I think we didn't really decide in this APE how this will be done for astropy.wcs specifically, we left it pretty general. Since we have control over the astropy.wcs code we could easily just add this interface to the existing WCS object to avoid too many nested levels.

APE13.rst
astrofrog: Thanks!

Cadair:
@cdeil The objective here is that for people using this interface in Python they would be using the high level API, so we designed the low level one to provide enough information to construct other objects that can provide more information like names. In the case of spatial coordinates this would be `SkyCoord` in astropy, but could be anything. Adding names to the `pixel_to_world_values` would, in my opinon, be tricky for minimal gain, when really other objects should be contstructed and passed to the user in the high level wrapper.

mhvk:
General nitpick: you call things "Python arrays". Do you mean `np.ndarray` or truly the python array type?

astrofrog:
@mhvk - we meant the true Python array type

mhvk:
@astrofrog - pure python array so that one doesn't have to worry about strides?  It might be useful to clarify why one wouldn't just allow numpy array (or anything with an `__array_interface__`?

APE14.rst
```diff
@@ -0,0 +1,450 @@
...
+
+Low-level API
+^^^^^^^^^^^^^
+
+The proposed low-level API could be implemented as an abstract base class but we
+would not require all implementations to inherit from it (if they do not inherit
+from it, then they should still register with the ABC to allow code to check
+isinstance(object, LowLevelWCSAPI)). Existing WCS classes are then free to
+either implement that interface or provide thin wrappers providing this
+interface. A guiding philosophy of this interface is use relatively primitive
```
r-owen: "is use" seems strange here. I suggest "is to use"

APE14.rst
```diff
@@ -0,0 +1,450 @@
...
+            A list with n_dim_world elements, where each element is a tuple with
+            two items. The first is a name for the world object this world array
+            corresponds to, which *must* match the string names used in
+            ``world_axis_object_metadata``. Note that names might appear twice
+            because two world arrays might correspond to a single world object
+            (e.g. a celestial coordinate might have both “ra” and “dec” arrays,
+            which correspond to a single sky coordinate object). The second
+            element is  either a string keyword argument name or a positional
+            index for the corresponding class from
+            ``world_axis_object_metadata``.
```
r-owen: Two suggestions for making this clearer:
- Use a bullet list for the elements.
- Reference the example below (e.g. "see example below").

APE14.rst
```diff
@@ -0,0 +1,450 @@
...
+        """
+        Convert pixel coordinates to world coordinates (represented by Astropy
+        objects).
+        """
+
+    def world_to_pixel(self, *world_objects):
+        """
+        Convert world coordinates (represented by Astropy objects) to pixel
+        coordinates
+        """
```
r-owen: I suggest that this method support named arguments for the arrays instead of forcing the user to get the argument order correct. The code will be easier to read and more robust.

Also consider returning a dict or a struct with named fields (to simplify element access, e.g. `world.sky` instead of `world["sky"]`), or at least a named tuple (which I dislike because it allows -- even encourages -- access by index).

APE14.rst
```diff
@@ -0,0 +1,450 @@
...
+more complex logic for converting the wcs outputs into composite Python objects.
+The structure outlined in this API would make that unnecessary by instead having
+a single interface that user code can write against, and only needs to consider
+other details when creating or modifying wcs.
+
+On a more specific note, the primary reason for using a string as the key for
+the dictionary for ``world_axis_object_classes`` (and the corresponding names in
+world_axis_object_components) is because there might be multiple world axes that
+need to use the same class. Otherwise a simpler solution would have been to use
+the class object *itself* as the key.
```
r-owen: What is the problem if you use the class object itself in the case that multiple world axes use the same class?

astrofrog:
@mhvk - oops I got it wrong before, we actually mean anything with the array *interface* not just the Python array type (just realized that was a separate thing). Will clarify in the APE text.

APE14.rst
astrofrog: Fixed

APE14.rst
astrofrog: Fixed

astrofrog:
@cdeil - we decided to add a recommendation for what the shape should be called if it is present (but we don't require it) - just thought I'd mention it since you asked

astrofrog:
@cdeil @r-owen - you both asked about named arguments. In this basic API I think we want to keep things simple and have the low-level API do position-only arguments (note that the high-level API described is much more general). Anyone implementing a low-level API object can of course always support named keywords too as long as passing positional arguments also works. This API is not to say that you can't accept keyword arguments, but a minimal set of functionality that should definitely work.

APE14.rst
```diff
@@ -0,0 +1,484 @@
+A shared Python interface for World Coordinate Systems
+------------------------------------------------------
+
+author: Thomas Robitaille, Erik Tollerud, Stuart Mumford, Adam Ginsburg
+
+date-created: 2017 November 7
+
+date-last-revised: 2017 November 7
```
eteq: Presumably this needs updating... but I suppose it has to be on accept/reject, anyway.

r-owen:
@astrofrog I agree that the low-level API should probably avoid named arguments. I was intending to suggest named arguments for the array inputs to the high-level version of pixel->world and world->pixel, in order to avoid confusion about order.

astrofrog:
@r-owen - ah right sorry, you were indeed commenting on the high-level API. We could easily add this kind of functionality later on the high-level class if needed, since there will only be one main implementation and we can improve it as needed along the way (the thing to get right now is the low-level API since that is the actual API specification for interoperability)

APE14.rst
```diff
@@ -0,0 +1,484 @@
...
+
+* To develop a high-level Astropy WCS object that can wrap any WCS object
+  providing the low-level API and be smarter about the kinds of objects to
+  accept or return to the user (including for example SkyCoord or Time objects).
+
+The idea of this two-tiered approach as opposed to a single-tier approach is
+that if we asked different WCS objects to provide a high-level API, this would
+(1) cause a lot of duplication of logic of constructing the appropriate Astropy
+objects, and (2) force those objects to return specifically Astropy objects,
+whereas having a package-independent API would be better.
```
keflavich: Can we add something like "better because it allows other packages, even non-python packages, to follow a well-defined API for the sake of interoperability."

Or something else saying specifically why it would be better?  My suggestion is a mouthfull...

APE14.rst
```diff
@@ -0,0 +1,484 @@
...
+**Simple 2D image mapping** where the pixel axes are lined up with RA and Dec
+(in FITS-WCS this would be CAR)
+
+.. code-block:: python
+
+    wcs.axis_correlation_matrix = [[True, False], [False, True]]
+    wcs.world_axis_units = [‘deg’, ‘deg’]
+    wcs.world_axis_physical_type = [‘pos.eq.ra’, ‘pos.eq.dec’]
+    wcs.world_axis_object_components = [((‘sc’, ‘ra’), (‘sc’, ‘dec’)]
+    wcs.world_object_data  = {‘sc’:(‘astropy.coordinates.SkyCoord’,
```
nden: typo - `wcs.world_axis_object_classes` ?

APE14.rst
```diff
@@ -0,0 +1,484 @@
...
+        def n_dim_world(self):
+            """
+            The number of axes in the world coordinate system
+            """
+
+        @property
+        def shape_pixel(self):
+            """
+            The shape of the data that the WCS applies to as a tuple of
+            length ``n_dim_pixel`` (optional).
```
nden: Does this silently imply start of the data at 0?
I'm wondering about cases when, for some reason (overscan, reference pixels), the real data does not start from 0.  In this case, for exmaple, a distortion should be applied to ((xstart, xend), (ystart, yend)) and a tuple like this with xstart/ystart defaulting to 0 and xend/yend to NAXIS would be better.

nden: Are there use cases when the actual data size is needed?

APE14.rst
astrofrog: Fixed

APE14.rst
astrofrog: @nden - the shape can be useful for e.g. slicing of WCS objects, but i agree the bounds are useful too - can you review https://github.com/astrofrog/astropy-APEs/pull/15 and let me know if you are happy with those changes?

APE14.rst
```diff
@@ -0,0 +1,484 @@
...
+Low-level API examples
+^^^^^^^^^^^^^^^^^^^^^^
+
+**Simple 1D spectrum** - a 1D mapping from pixel to wavelength:
+
+.. code-block:: python
+
+    wcs.axis_correlation_matrix = [[True]]
+    wcs.world_axis_units = ['angstrom']
+    wcs.world_axis_physical_type = ['spectral']
```
sfgraves: Should 'spectral' be one of the ucd types -- e.g. em.wl ?

APE14.rst
```diff
@@ -0,0 +1,484 @@
...
+dimension encoding time-of-observation.
+
+.. code-block:: python
+
+    wcs.axis_correlation_matrix = [[True, True, False],
+                                   [True, True, False],
+                                   [True, True, False],
+                                   [False, False, True]]
+    wcs.world_axis_units = ['deg', 'deg', 'angstrom', 'day']
+    wcs.world_axis_physical_type = ['pos.galactic.lon', 'pos.galactic.lat', 'em.freq', 'time']
```
sfgraves: Tiny typo: I think 'em.freq' should be 'em.wl' here? as you use angstrom and astropy.units.Wavelength.

taldcroft:
This is approved by the Astropy Coordination committee (Kelle and Tom A; Tom R and Erik were recused).

Cadair:
:fireworks:

eteq:
@taldcroft - Can you clarify which steps of https://github.com/astropy/astropy-apes#accepting-apes you did here? I don't see it in the readme (step 4) - I can fix that of course, but I'm wondering if you did step 6 (7 is sort of optional, although since there's a thread on astropy-dev, it's probably worth doing)?

PR #35, opened 2017-11-17
Removing 3rd 't' in cookiecutter command

karllark:
Extremely minor fix.
Currently, cutting and pasting the recommended command to run cookiecutter does not work as the command has too many ts.


bsipocz:
Clearly a typo, so I'm merging this now. Thanks @karllark!

PR #41, opened 2018-03-19
update APE10 for new release calendar

eteq:
This is a small update to APE10 that accounts for the new release calendar after the +3 month change to avoid the holidays.  The coordination committee approved this change in a committee meeting this week.

One thing to bear in mind: this makes the v4.0 release (the first LTS that's py3-only) in 2020 instead of the end of 2019.  If I take this literally, that means there's a 3 month LTS gap.  @astrofrog (as APE10's author) should clarify: should I also update the "period of support" for v2.0 to go to then?  I think this is not really a big deal because the Python devs have been a bit ambiguous about exactly when py2 support ends (i.e. exactly when in "2020" the ending is) - so we can play it by ear a bit if necessary.

### DIFF
--- a/APE10.rst
+++ b/APE10.rst
@@ -5,7 +5,7 @@ author: Tom Robitaille
 
 date-created: 2016 July 13
 
-date-last-revised: 2017 May 30
+date-last-revised: 2018 Mar 19
 
 date-accepted: 2016 August 22
 
@@ -63,10 +63,10 @@ Version      Planned release date     Notes
 v1.3         December 2016
 v2.0 - LTS   June 2017                Supported with bug fixes until end of 2019
 v3.0         December 2017            First release to support only Python 3+
-v3.1         June 2018
-v3.2         December 2018
-v3.3         June 2019
-v4.0 - LTS   December 2019            Supported with bug fixes for two years
+v3.1         October 2018
+v3.2         April 2019
+v3.3         October 2019
+v4.0 - LTS   April 20120              Supported with bug fixes for two years
 ==========   ======================   ==========================================
 
 Note that v1.0, v2.0, and v4.0 are marked as long-term support releases (LTS),
@@ -266,3 +266,6 @@ noted that one should not aim to support versions of Python that will reach end
 of life during the time an astropy version is supported. In particular, Python
 3.3 would reach end-of-life only 3 months after the nominal astropy 2.0 release.
 Hence, all minimum versions of Python 3 were increased by one minor version.
+
+After v3.0, a decision was made to adjust the release calendar by +3 months.  The
+table above has been updated appropriately.
--- a/APE10.rst
+++ b/APE10.rst
@@ -66,7 +66,7 @@ v3.0         December 2017            First release to support only Python 3+
 v3.1         October 2018
 v3.2         April 2019
 v3.3         October 2019
-v4.0 - LTS   April 20120              Supported with bug fixes for two years
+v4.0 - LTS   April 2020               Supported with bug fixes for two years
 ==========   ======================   ==========================================
 
 Note that v1.0, v2.0, and v4.0 are marked as long-term support releases (LTS),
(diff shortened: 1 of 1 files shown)


APE10.rst
```diff
@@ -63,10 +63,10 @@ Version      Planned release date     Notes
...
 v2.0 - LTS   June 2017                Supported with bug fixes until end of 2019
 v3.0         December 2017            First release to support only Python 3+
-v3.1         June 2018
-v3.2         December 2018
-v3.3         June 2019
-v4.0 - LTS   December 2019            Supported with bug fixes for two years
+v3.1         October 2018
+v3.2         April 2019
+v3.3         October 2019
+v4.0 - LTS   April 20120              Supported with bug fixes for two years
```
dpshelio: That's a lot to wait for v4.0!!! :wink:

saimn:
Quote from Guido a few days ago:

> Let's not play games with semantics. The way I see the situation for 2.7 is
that EOL is January 1st, 2020, and there will be no updates, not even
source-only security patches, after that date.

https://mail.python.org/pipermail/python-dev/2018-March/152348.html

bsipocz:
> One thing to bear in mind: this makes the v4.0 release (the first LTS that's py3-only) in 2020 instead of the end of 2019. 

what about not having a 3.3 release, but 4.0 instead? That case with the late October feature freeze, 4.0 would be released sometimes 2019 Nov, and we can easily manage to have 2 LTS parallel for a month after that. I assume anyway that we start dropping many of the backports in 2019 given they will be more and more complicated by that time.

APE10.rst
eteq: Darn, you caught my scheme to make it so we don't have to work so hard on support!  (thanks! 😉 )

eteq:
I think I'm 👍to @bsipocz's plan. @astrofrog, sound good to you?  That makes it less of a "minor perturbation", but I think this is fairly close to the original spirit of the plan...

This also raises the question as to whether the "month" listed is for release or for feature freeze.  Probably I should switch them to the *following* month if it's release, as realistically, a freeze on the 20-somethingth will not be release until the next month.  If you agree that's the right interpretation, @astrofrog, then I'll advance the months by one.

bsipocz:
> This also raises the question as to whether the "month" listed is for release or for feature freeze.

my understanding on the telecon was to have freeze on the fix date, and then get the release out asap afterwards (~in a month). If we release in Oct, then we are back at square 1 as then the freeze should be on the equinox.

eteq:
@bsipocz - sorry, I wasn't clear there.  I meant I was wondering what  *the APE10 table* meant.  That is, are the dates there release dates or freeze dates?  What I have in this PR is the freeze dates.  I'm pretty sure the table is supposed to be release dates which I agree is 1 month later than what is currently in this PR.  I'll update it once I get confirmation from @astrofrog that I'm interpreting the text correctly (since he's the APE10 author).

bsipocz:
Oh, and I haven't realized this is about changing them on an accepted APE. Is that really necessary? I mean, is the APE really used for these dates or was just a guideline? I always assumed that the "official" dates are in the release calendar wiki page.

astrofrog:
I think the equinox + 1 was meant to be feature freeze?

> I think I'm 👍to @bsipocz's plan. @astrofrog, sound good to you? That makes it less of a "minor perturbation", but I think this is fairly close to the original spirit of the plan...

👍  from me!

taldcroft:
My understanding was that the table here is the *release* date, and so should be May (April 21-ish + few weeks) and November (Oct 21-ish + few weeks).

:+1: on @bsipocz's plan.

eteq:
@bsipocz - re: this being an accepted APE.  The idea is that some people will find the dates here instead of the release calendar, so it makes sense to make them consistent.  Also, especially if we are changing the version numbers, it's important the APE also gets revised because otherwise the APE is in conflict with the reality.  (But there's a note at the end indicating the history... and `git`!)

Ok, will update the PR momentarily to reflect @bsipocz's proposal.

eteq:
Should be updated now and I think ready to go. (FWIW, as I was editing I saw that the heading is pretty clearly "Planned Release Date" - so that whole discussion above re: release vs. feature freeze was poor reading comprehension on my part...)

PR #45, opened 2018-05-23
Revisions for APE14

astrofrog:
As described in https://docs.google.com/document/d/13bJ-tHOuI6pVmB8KtthYB_fBFIYNDF2cApJwt92G6w8/edit, APE14 has significant issues that prevent implementation. Thus, this PR makes revisions to help with the implementation.

@eteq @Cadair @keflavich @nden @parejkoj - please review and let me know what you think! See the google doc above for more justification about the changes.

**Note:** I believe we should keep this open until at least one implementation has been completed to avoid more revisions

### DIFF
--- a/APE14.rst
+++ b/APE14.rst
@@ -180,23 +180,7 @@ low-level API:
             """
 
         @property
-        def pixel_shape(self):
-            """
-            The shape of the data that the WCS applies to as a tuple of
-            length ``pixel_n_dim`` (optional).
-
-            If the WCS is valid in the context of a dataset with a particular
-            shape, then this property can be used to store the shape of the
-            data. This can be used for example if implementing slicing of WCS
-            objects. This is an optional property, and it should return `None`
-            if a shape is not known or relevant.
-
-            The shape should be given in ``(x, y)`` order. For a shape that can
-            be used for Numpy arrays, see ``numpy_shape``.
-            """
-
-        @property
-        def numpy_shape(self):
+        def array_shape(self):
             """
             The shape of the data that the WCS applies to as a tuple of
             length ``pixel_n_dim`` (optional).
@@ -208,8 +192,7 @@ low-level API:
             if a shape is not known or relevant.
 
             The shape should be given in ``(row, column)`` order (the convention
-            for Numpy arrays). For a shape in ``(x, y)`` order, see
-            ``pixel_shape``.
+            for arrays in Python).
             """
 
         @property
@@ -272,7 +255,7 @@ low-level API:
             ``y`` is the vertical coordinate.
             """
 
-        def numpy_index_to_world_values(self, *index_arrays):
+        def array_index_to_world_values(self, *index_arrays):
             """
             Convert array indices to world coordinates. This is the same as
             ``pixel_to_world_values`` except that the indices should be given
@@ -293,7 +276,7 @@ low-level API:
             ``y`` is the vertical coordinate.
             """
 
-        def world_to_numpy_index_values(self, *world_arrays):
+        def world_to_array_index_values(self, *world_arrays):
             """
             Convert world coordinates to array indices. This is the same as
             ``world_to_pixel_values`` except that the indices should be returned
@@ -470,9 +453,10 @@ The API above provides a way to distinguish between pixel coordinates defined
 using the standard Cartesian ordering (x, y) and array indices defined using the
 row-major ordering (i.e. row, column). For example, values returned from
 ``world_to_pixel_values`` would be in the correct order to use for plotting
-using e.g. Matplotlib, while values returned from ``world_to_numpy_index_values``
-would be in the correct order to use for indexing a Numpy array. Both are valid
-in different contexts and we therefore provide two methods for each transformation.
+using e.g. Matplotlib, while values returned from ``world_to_array_index_values``
+would be in the correct order to use for indexing e.g. a Numpy array. Both are
+valid in different contexts and we therefore provide two methods for each
+transformation.
 
 We do not mandate a specific order for the world coordinates. While it might be
 tempting to assume the 'same' order as for pixel coordinates, this only makes
@@ -516,13 +500,13 @@ reorederd world coordinates - this would involve changing the order of
 ``world_axis_physical_types``, ``world_axis_units``, and
 ``world_axis_object_components``, changing the order of
 ``axis_correlation_matrix`` along the first dimension, and changing the order
-of the inputs of the ``world_to_pixel/numpy_index_values`` methods and the
-order of the outputs of the ``pixel/numpy_index_to_world_values`` methods. Thus,
+of the inputs of the ``world_to_pixel/array_index_values`` methods and the
+order of the outputs of the ``pixel/array_index_to_world_values`` methods. Thus,
 implementations of the low or high-level API could provide convenience methods
 to reorder or sort the world axes.
 
 This flexibility does not however extend to pixel coordinates. For example for a
-given array with an associated WCS, the output of ``world_to_numpy_index_values`` has
+given array with an associated WCS, the output of ``world_to_array_index_values`` has
 to consistently return the values in the order that can be used to index the
 array, so the indices/pixel coordinates of the WCS cannot be re-ordered if the
 data is left unchanged.
@@ -600,10 +584,10 @@ API includes the following four methods:
         conventions.
         """
 
-    def numpy_index_to_world(self, *index_arrays):
+    def array_index_to_world(self, *index_arrays):
         """
         Convert array indices to world coordinates (represented by Astropy
-        objects). See ``numpy_index_to_world_values`` for array indexing and ordering
+        objects). See ``array_index_to_world_values`` for array indexing and ordering
         conventions.
         """
 
@@ -614,10 +598,10 @@ API includes the following four methods:
         ordering conventions.
         """
 
-    def world_to_numpy_index(self, *world_objects):
+    def world_to_array_index(self, *world_objects):
         """
         Convert world coordinates (represented by Astropy objects) to array
-        indices. See ``world_to_numpy_index_values`` for array indexing and ordering
+        indices. See ``world_to_array_index_values`` for array indexing and ordering
         conventions. The indices should be returned as rounded integers.
         """
 
--- a/APE14.rst
+++ b/APE14.rst
@@ -82,10 +82,11 @@ and should be *compatible* with this APE's interface, an API for these
 operations is out-of-scope for this APE. This APE also does not address how to
 *create* such WCSs. Rather it is aimed specifically at an API for the simpler
 problem of actually performing the “end-to-end” transformation from pixel to
-some world space. We note that while we have made efforts to ensure that the API
-described here is as close as possible to the final implemented API, the
-authoritative version of the API will be given by the base classes that live in
-the core astropy package.
+some world space.
+
+We note that while we have made efforts to ensure that the API described here is
+as close as possible to the final implemented API, the authoritative version of
+the API will be given by the base classes that live in the core astropy package.
 
 Overview of the proposed WCS Interface
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -115,7 +116,7 @@ interoperability.
 
 
 Pixel Conventions
-"""""""""""""""""
+^^^^^^^^^^^^^^^^^
 
 The exact choice of which value represents the "edge" of a pixel is arbitrary
 and could potentially vary across WCS implementations. Indeed, the
@@ -182,17 +183,30 @@ low-level API:
         @property
         def array_shape(self):
             """
-            The shape of the data that the WCS applies to as a tuple of
-            length ``pixel_n_dim`` (optional).
+            The shape of the data that the WCS applies to as a tuple of length
+            ``pixel_n_dim`` in ``(row, column)`` order (the convention for
+            arrays in Python) (optional).
 
             If the WCS is valid in the context of a dataset with a particular
             shape, then this property can be used to store the shape of the
             data. This can be used for example if implementing slicing of WCS
             objects. This is an optional property, and it should return `None`
             if a shape is not known or relevant.
+            """
 
-            The shape should be given in ``(row, column)`` order (the convention
-            for arrays in Python).
+        @property
+        def pixel_shape(self):
+            """
+            The shape of the data that the WCS applies to as a tuple of length
+            ``pixel_n_dim`` in ``(x, y)`` order (where for an image, ``x`` is
+            the horizontal coordinate and ``y`` is the vertical coordinate)
+            (optional).
+
+            If the WCS is valid in the context of a dataset with a particular
+            shape, then this property can be used to store the shape of the
+            data. This can be used for example if implementing slicing of WCS
+            objects. This is an optional property, and it should return `None`
+            if a shape is not known or relevant.
             """
 
         @property
--- a/APE14.rst
+++ b/APE14.rst
@@ -581,10 +581,15 @@ set of custom type names).
 High-level API
 ^^^^^^^^^^^^^^
 
-The high-level API will provide the ability for example to get ``SkyCoord``,
-``Time`` etc. objects back from a pixel to world conversion, and conversely to
-be able to convert ``SkyCoord``, ``Time`` etc. to pixel values. This API
-includes the following four methods:
+The high-level API's primary purpose is to provide an interface to obtain
+fully-featured Python objects for the world coordinates - for example to get
+``SkyCoord``, ``Time`` etc. objects back from a pixel to world conversion, and
+conversely to be able to convert ``SkyCoord``, ``Time`` etc. to pixel values. It
+is distinct from the actual implementation provided in the core astropy package
+(discussed below), so that developers providing their own WCS objects can
+implement the high-level API on their own. In this sense the astropy-provided
+implementation can be thought of as a reference implementation. The high-level
+API includes the following four methods:
 
 .. code-block:: python
 
--- a/APE14.rst
+++ b/APE14.rst
@@ -732,3 +732,6 @@ Decision rationale
 The content of this APE was discussed and accepted by multiple community stakeholders
 who have technical knowledge, practical experience, and project-level interest in WCS.
 The APE was accepted on Feb 28, 2018.
+
+This APE was subsequently revised on Nov 14, 2018 due to changes motivated by
+implementations in the astropy and gwcs packages.
--- a/APE14.rst
+++ b/APE14.rst
@@ -303,6 +303,14 @@ low-level API:
             """
 
         @property
+        def serialized_classes(self):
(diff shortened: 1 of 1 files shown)


APE14.rst
```diff
@@ -242,13 +248,18 @@ low-level API recommends:
             n_world scalars or arrays as input in units given by ``world_axis_units``.
             Returns n_pixel scalars or arrays. Note that pixel coordinates are
             assumed to be 0 at the center of the first pixel in each dimension.
+            to be 0 at the center of the first pixel in each dimension. If a
+            world coordinate does not have a matching pixel coordinate, NaN can
+            be returned.  The coordinates should be returned in the ``(x, y)``
```
Cadair: I am understanding this correctly that this is *not* numpy order (numpy being rows, columns for 2D right?)? If so I am infinitely :-1: on this, this is a Python API and should be doing things in C order everywhere.

nden: I think of this as a WCS API and as such it is Cartesian coordinates order, i.e. (x, y, z). I believe this is what is common for other packages as well so there's some expectation of this order.

In practical terms, since `astropy.wcs` and `gwcs` use Cartesian order too (not sure about LSST's software), this means the implementation does not have to rearrange the order of the inputs. It would be awkward for the high level interface and the low level object to work with a different order of inputs.
Although I see the appeal of using numpy order.

Cadair: Does gwcs actually enforce this? I have been constructing all my gwcses in numpy order to avoid this disparity.

I really really dislike the fact that astropy WCS axes order does not match the order of the numpy array I think it leads to a really bad user experience, given this API is Python all the way down I really believe that the best thing to do is be consistent with the ordering of the arrays you are presumably working with when you make your WCS.

For just a little bit of context this is what the `repr` for a `NDCube` looks like:
```
NDCube
---------------------
WCS Keywords

Number of WCS axes: 3
CTYPE : 'HPLN-TAN'  'HPLT-TAN'  'TIME'  
CRVAL : 0.0  0.0  0.0  
CRPIX : 2060.610107  2040.26001  1.0  
PC1_1 PC1_2 PC1_3  : 0.99999711341582  0.0024027401094365  0.0  
PC2_1 PC2_2 PC2_3  : -0.0024027401094365  0.99999711341582  0.0  
PC3_1 PC3_2 PC3_3  : 0.0  0.0  1.0  
CDELT : 0.00016669694444444  0.00016669694444444  12.00000000000188  
NAXIS : 0  0
---------------------
Length of NDCube: [1703. 4096. 4096.] pix
Axis Types of NDCube: ('time', 'custom:pos.helioprojective.lat', 'custom:pos.helioprojective.lon')
```

I really do not like the way the dimensions of the array and the WCS have a mismatch. I have been telling @danryanirsh and everyone else who will listen that gWCS and the APE 14 API will remove this glaring inconsistency, and I am very disappointed that it might not!

astrofrog: As discussed on slack, we could agree to change it, I think this clearly just needs more discussion. But the APE as-is was under-specific about this.

APE14.rst
```diff
@@ -269,21 +284,42 @@ low-level API recommends:
...
+            a requirement for the implementation of the high-level interface.
+
+            The second tuple element for each value of this dictionary can in
+            turn contain either instances of classes, or if necessary can contain
+            serialized versions that should take the same form as the main
+            classes described above (a tuple with three elements with the
+            fully qualified name of the class, then the positional arguments
+            and the keyword arguments). For low-level API objects implemented
+            in Python, we recommend simply returning the actual objects (not
+            the serialized form) for optimal performance.
```
astrofrog: One open question - how do we know whether a tuple is a serialized version or the actual Python object?

APE14.rst
```diff
@@ -269,21 +284,42 @@ low-level API recommends:
             """
             A dictionary with each key being a string key from
             ``world_axis_object_components``, and each value being a tuple with
-            two elements:
+            two or three elements:
```
Cadair: Why not just mandate this be three long, and just specify the second argument as an empty tuple? It should make all the logic much easier?

APE14.rst
Cadair: Why not just mandate realised python classes here? What is this addition trying to solve?

astrofrog: The API was designed so that if one wanted to implement it in C with the Python API, that could be done. For that to work, we have to allow objects to be serialized. However, we then realized this would have a performance hit when it's not needed, so this addition was to say that it's ok to just use Python classes. Maybe this should be rephrased to say that this is the standard, and then to give the serialization as an option.

APE14.rst
astrofrog: I agree - someone (maybe @eteq) said they found the empty tuple to not be elegant - but I agree it's additional complication for not much.

APE14.rst
Cadair: > For that to work, we have to allow objects to be serialized.

I mean *have to* is strong here, it's perfectly possible to create Python objects from C. I think we need to walk this line carefully.

Also is it likely that a tuple will actually ever be the desired Python object?

APE14.rst
eteq: I don't remember if it was me or not at the time... But me *right bow* agrees with @cadair :wink:. I think in implementation it should be a `NamedTuple` but that doesn't need to be said in the APE.

Cadair: I was about to suggest namedtuple until I remembered the C-API thing, you would have to import and construct the namedtuple instance from where it was created.

APE14.rst
```diff
@@ -406,6 +426,74 @@ dimension encoding time-of-observation.
...
+
+This flexibility does not however extend to pixel coordinates. For example for a
+given array with an associated WCS, the output of ``world_to_index_values`` has
+to consistently return the values in the order that can be used to index the
+array, so the indices/pixel coordinates of the WCS cannot be re-ordered if the
+data is left unchanged.
+
+For consistency with existing WCS libraries, we recommend that implementations
+based on FITS-WCS choose to order ``world_axis_physical_types`` in the same
+order as the ``CDELT`` values, but we do not require this.
```
nden: Did you mean `CTYPE` here?

astrofrog: woops yes

APE14.rst
```diff
@@ -242,6 +245,14 @@ low-level API recommends:
             ``y`` is the vertical coordinate.
             """
 
+        def index_to_world_values(self, *index_arrays):
+            """
+            Convert array indices to world coordinates. This is the same as
+            ``pixel_to_world_values`` except that the indices should be given
```
eteq: Here (and the same below in ``world_to_index_values``: I wonder if this needs to be clear about something like "the native order of the data array" instead of specifying the row-major way of describing it?  That is, if at some point in the future, a new array package comes out that is column-major instead of row-major, presumably we want this to work that such that indexing is correct, right?

This could just be a single small addition - e.g., ``in `(i, j)` order`` -> ``in the appropriate index order for the data - for numpy, `(i, j)` order``.  But then in the added discussion below, there should be mention that the real "contract" for the ``index`` methods is that they *match* the data, regardless as to whether that's row- or column-major

APE14.rst
```diff
@@ -406,6 +426,74 @@ dimension encoding time-of-observation.
     wcs.world_axis_object_components = [('spec', 0, 'value')]
     wcs.world_axis_object_classes  = {'spec':('astropy.units.pixel': {})}
 
+Pixel and world coordinate ordering
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The API above provides a way to distinguish between pixel coordinates defined
+using the standard Cartesian ordering (x, y) and array indices defined using the
+(row, column) ordering. For example, values returned from
```
eteq: (minor): ``(row, column) ordering`` -> ``row-major ordering (i.e. row, column)``?

APE14.rst
```diff
@@ -406,6 +426,74 @@ dimension encoding time-of-observation.
...
+``world_to_pixel_values`` would be in the correct order to use for plotting
+using e.g. Matplotlib, while values returned from ``world_to_index_values``
+would be in the correct order to use for indexing a Numpy array. Both are valid
```
eteq: ``Both are valid`` -> ``Both are valid in different contexts`` ?

APE14.rst
astrofrog: In principle I agree but there is an additional complication here - it's possible to define a WCS not attached to a specific array - and in fact the data is not a property of the WCS. So a single WCS could apply to both data in Numpy form and in another form. So for that reason I actually feel like we'd be better off just renaming the methods to say ``numpy_index`` instead of just ``index`` and then one day if another popular package comes along, we can add new methods for that.

APE14.rst
astrofrog: Agree

APE14.rst
astrofrog: Agree

APE14.rst
astrofrog: Fixed in https://github.com/astropy/astropy-APEs/pull/45/commits/e88f4b37f22084021a93b96f61b59c45316f14a3

APE14.rst
astrofrog: Fixed in 5c66cedf12db349f306f50f3b823212427be5a94

APE14.rst
astrofrog: Fixed in 0b34b6ccb10618c2ebdb2c24e416d2df0bb5014f

APE14.rst
astrofrog: Fixed in 0b34b6ccb10618c2ebdb2c24e416d2df0bb5014f

APE14.rst
astrofrog: See the change I suggest in ef717cb0972f312c90a9ecb927e9f6d4ffbf4566

APE14.rst
astrofrog: This was addressed in https://github.com/astropy/astropy-APEs/pull/45/commits/8196354329cc2f7c671007dc94781392e3f097cb following a discussion on a telecon.

astrofrog:
Ok so I think we are almost there (for real). The one outstanding question was how we distinguish tuples which represent serialized python objects compared to actual tuples which may be valid arguments for some classes. I think I have a solution - we simply make it so that low level API objects either have all their objects serialized or none, and we use a class-level attribute to indicate which (since the implementer knows). For instance we could call this ``serialized_classes=True/False``.

@eteq @Cadair @keflavich @nden @parejkoj - thoughts?

I'm working on updating the implementations in the Astropy PRs.

APE14.rst
```diff
@@ -233,7 +238,19 @@ low-level API recommends:
...
             ``world_axis_units``. Note that pixel coordinates are assumed
-            to be 0 at the center of the first pixel in each dimension.
+            to be 0 at the center of the first pixel in each dimension. If a
+            pixel is in a region where the WCS is not defined, NaN can be
+            returned. The coordinates should be specified in the ``(x, y)``
+            order, where for an image, ``x`` is the horizontal coordinate and
+            ``y`` is the vertical coordinate.
+            """
+
+        def numpy_index_to_world_values(self, *index_arrays):
```
Cadair: I am still unsure on putting `numpy` in here, would just 'array' not be more appropriate?

astrofrog: Well this was in response to @eteq who pointed out that in future another array package might exist that uses the opposite convention...

Cadair: I find that to be **extremely** unlikely given that the Python language spec is C-order. Also other array packages do current exist that are in C-order. I would rather be inclusive of what current exists rather than worry about things that wont exist beyond midday on April 1st.

astrofrog: @eteq - would you be happy to just switch to ``array_index``?

nden:
I think there're two other underspecified properties - the order of the tuples in `pixel_shape` and `pixel_bounds` - Cartesian vs numpy ordering. The definition of `pixel_bounds` directly matches `gWCS.bounding_box` so I have a preference for this to be in Cartesian order. 
None of these currently have equivalent properties in astropy.wcs.

astrofrog:
@nden - I've now added a commit to clarify this. I think to be consistent, ``pixel_*`` properties should always be in cartesian order, and if we want things consistent with Numpy they can be called ``numpy_*``. So I've added ``numpy_shape``. Does this seem ok to you?

eteq:
@astrofrog - Instead of `numpy_shape`, perhaps `array_shape`? Or `idx_shape`? I'm a bit cautious about tying it specifically to numpy in, say, a future world where everyone is using `xarray` (that is, they all follow numpy conventional order and probably will in the future but it's even now possible to do this with other numpy-alikes)

APE14.rst
eteq: Yes, I guess so.  I came to this conclusion totally on my own, not realizing that I'd said the other thing.  So sure, lets go with `array_index`.

APE14.rst
```diff
@@ -412,40 +578,43 @@ agreed as much as possible between different packages to make sure that these
...
 
-High-level Astropy Object
-^^^^^^^^^^^^^^^^^^^^^^^^^
+High-level API
+^^^^^^^^^^^^^^
 
-Unlike the low-level API, the 'high-level' interface described here will be a
-single Astropy-developed class since it interfaces with various Astropy objects.
-This high-level API would provide the ability for example to get ``SkyCoord``,
+The high-level API will provide the ability for example to get ``SkyCoord``,
```
eteq: Add just a bit more context here.  Maybe "The high-level API's primary purpose is to provide fully-featured python objects for the world coordinates.  For example, to get ...<text already there>.  It is distinct from the actual implementation (discussed below) so that other objects can implement it on their own.  In this sence the astropy-provided implementation can be thought of as a reference implementation. "

APE14.rst
```diff
@@ -467,6 +638,33 @@ for ``pixel_to_world``, but on the other hand provided there is no ambiguity,
...
+the low- and high-level base classes. The high-level base class will be
+implemented with methods that will by default use the information in
+``world_axis_object_components`` and ``world_axis_object_classes`` to work -
+thus, simply inheriting from the high-level base class should be sufficient to
+expose the high-level API.
+
+High-level Astropy Object
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Finally, we will develop a class in Astropy that inherits from the high-level
```
eteq: "Astropy" -> \`astropy\` (i.e., the core package).

APE14.rst
astrofrog: Fixed

APE14.rst
astrofrog: Fixed

APE14.rst
astrofrog: Changed. I've also removed ``pixel_shape`` and changed ``numpy_shape`` to ``array_shape``.

astrofrog:
I'm not going to focus on finishing up the pull requests with the implementations. I suggest merging this PR only once the implementations are merged and the GWCS implementation is also done, in case any other issues come up.

mcara:
As mentioned in https://github.com/spacetelescope/drizzlepac/issues/186#issuecomment-433897429 (and I agree with @philhodge), `pixel_shape` is suggestive of properties of **pixels themselves**. Since already there is a quantity, non-informatively named `numpy_shape` (as if `numpy` has a shape), is there a need for `pixel_shape` as well? At least `numpy_shape` appears to be less confusing.

APE14.rst
```diff
@@ -182,6 +190,9 @@ low-level API recommends:
             data. This can be used for example if implementing slicing of WCS
             objects. This is an optional property, and it should return `None`
             if a shape is not known or relevant.
```
mcara: Shouldn't _"...if a shape is not known or relevant."_ be something like _"... if a shape is neither known nor relevant."_? The way this is currently written is confusing to me (but it just could be my English): does **not** apply to **known** and not to **relevant** or does **not** apply to _both_ **known** _and_ **relevant**?

Same goes for `numpy_shape`, `pixel_shape`.

APE14.rst
```diff
@@ -190,10 +201,11 @@ low-level API recommends:
...
 
-            WCS solutions are sometimes only guaranteed to be accurate within a
-            certain range of pixel values, for example when definining a WCS
-            that includes fitted distortions. This is an optional property, and
-            it should return `None` if a shape is not known or relevant.
+            The bounds should be given in ``[(xmin, xmax), (ymin, ymax)]``
+            order. WCS solutions are sometimes only guaranteed to be accurate
+            within a certain range of pixel values, for example when defining a
+            WCS that includes fitted distortions. This is an optional property,
+            and it should return `None` if a shape is not known or relevant.
```
mcara: Shouldn't _"...if a shape is not known or relevant."_ be something like _"... if a shape is neither known nor relevant."_? The way this is currently written is confusing to me (but it just could be my English): does **not** apply to **known** and not to **relevant** or does **not** apply to _both_ **known** _and_ **relevant**?

Same goes for `numpy_shape`, `pixel_shape`.

astrofrog:
@mcara - can you take a look at the version that's been merged into astropy? Specifically http://docs.astropy.org/en/latest/wcs/wcsapi.html and the associated API docs. There we use ``array_shape`` and ``pixel_shape`` and have tried to clarify what these mean in the docstring (and in the narrative docs)

mcara:
@astrofrog I do not see any kind of documentation for `pixel_shape` and `array_shape` (which, IMO, is better than `numpy_shape`) in the link provided by you. The only reference that I can see is:

> As mentioned in [Pixel conventions and definitions](http://docs.astropy.org/en/latest/wcs/wcsapi.html#pixel-conventions-and-definitions), what would normally be considered the ‘y-axis’ of the image (when looking at it visually) is the first dimension, while the ‘x-axis’ of the image is the second dimension. Thus [`array_shape`](http://docs.astropy.org/en/latest/api/astropy.wcs.WCS.html#astropy.wcs.WCS.array_shape) returns the shape in the opposite order to the NAXIS keywords in the FITS header (in the case of FITS-WCS). If you are interested in the data shape in the reverse order (which would match the NAXIS order in the case of FITS-WCS), then you can use [`pixel_shape`](http://docs.astropy.org/en/latest/api/astropy.wcs.WCS.html#astropy.wcs.WCS.pixel_shape).

Both links (for `pixel_shape` and `array_shape`) lead to nothing:

array_shape_documentation

pixel_shape_documentation

Am I missing something?

APE14.rst
astrofrog: Fixed

astrofrog:
I propose that we now consider merging this PR - the APE has been implemented in the core package and GWCS. We can still make future changes in the implementation if we like, since we now include the clause:

*We note that while we have made efforts to ensure that the API described here is
as close as possible to the final implemented API, the authoritative version of
the API will be given by the base classes that live in the core astropy package.*

If anyone who comes across this in future has any suggestions for improving docstrings or the API, please open issues or pull requests in the core package since that is now the authoritative version of the actual API.

astrofrog:
Since @eteq and @nden 👍 'd my last comment, and @keflavich ok-ed it in Slack, I'm going to go ahead and merge these revisions

PR #46, opened 2018-11-06
Initial Draft of APE16: CPU Threading

jamienoss:
This still needs a lot of work. I've stalled on it a little and rather it sat out here than not at all.

Signed-off-by: James Noss <jnoss@stsci.edu>

### DIFF
--- a/APE16.rst
+++ b/APE16.rst
@@ -1,0 +1,159 @@
+CPU Threading [DRAFT]
+----------------
+
+author: Jamie Noss
+
+date-created: 2018 October 26
+
+date-last-revised: 2018 October 26
+
+date-accepted: TBD
+
+type: Standard Track
+
+status: Discussion
+
+
+Abstract
+--------
+
+As the datasets utilized in modern day astronomy become ever increasing and
+continue to cross over into the realm of Big Data, so does the computational burden to process
+them. At present, there is only one hardware approach to solving this demand and
+that is parallel processing.
+
+This proposal is only intended to address CPU threading and not, e.g., GPU threading.
+It is also only intended to address the *allowance* of utilizing said technology and *not*
+necessarily a broader directive to explicitly incorporate said technology
+everywhere applicable throughout the Astropy code base.
+It is intended to address at which level(s) the Astropy package interfaces
+with a given threading package/library. Whether at the Python level and/or at the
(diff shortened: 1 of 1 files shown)


jamienoss:
In relation to the following comment https://github.com/astropy/astropy/issues/7971#issuecomment-433176324

> As far as I understand the special interpolating functionality that astropy.convolution adds (this is the paper cited in the Astropy paper) can be done as a pre-processing step once during data reduction

It's not really a question of _can_ vs _can't_ but a question of whether it needs to. The reason for this is that the interpolation scheme is itself a convolution. It's convolution with a normalized kernel where the normalization is a function of the presence of NaN values. So ok, it can technically be done pre-convolution, however, it would be equivalent to doing the convolution.

So semantically, ``convolve(nan_treatment='interpolate')`` isn't convolution that deals with NaN values by interpolating them, it is an interpolation scheme for NaN values by convolving with a value dependent normalized kernel. The only caveat to this is thatit may  not be functionally complete because if a region of NaN values is greater or equal to the kernel size, the centroid won't be interpolated thus persisting those centroid NaN values. So some care in its use is required to make sure the kernel is large enough to cover potential NaN regions. 

At least that's my take on things.

cdeil:
FYI, in case you want to continue the APE: I asked for Scipy, and they don't use OpenMP because it's not considered portable enough. They use very little multi-core overall: mostly in the linear algebra packages they call into (don't know what those use), and very few functions use Python multiprocessing from the Python standard library, and even there because of issues on some platforms they are considering using something like https://github.com/joblib/joblib that's apparently more robust. I do think a short APE that is a summary of available options in Python and what other projects use would be very useful, as a basis for discussions what (if anything) to do in Astropy to improve performance in the future. Not sure if an APE would be the appropriate format for that, it could also be a wiki page, or a poster / proceeding at e.g. a Scipy conference "Options for multi-core CPU processing for scientific Python packages in 2019".

eteq:
@jamienoss - are you still interested in pursuing this APE?  If you don't think you have the time/desire, it may be we need to consider this APE withdrawn.  We can then close this PR, but then you can leave your branch in place so that if anyone wants to they can pick up this topic (but include you as a co-author of course if they start from that text!)

PR #52, opened 2019-03-07
A roadmap for package infrastructure without astropy-helpers

astrofrog:
This is an APE describing a roadmap for transitioning away from astropy-helpers in favor of more standardize tools in the Python ecosystem. The short version is that:

* Pure-Python packages that use astropy-helpers currently can already be refactored to not rely on astropy-helpers, and in fact several packages have already done so (see links in the implementation section).

* Packages with C extensions can start to make some of the changes here but should wait a little before making the more significant changes related in particular to build-time dependencies since they require a recent version of pip (19.0 or later) - but we wanted to start the discussion now to start planning the changes.

A fraction of astropy-helpers would continue to exist as a more general extension-helpers package specifically geared towards packages with compiled extensions, and this could be used outside the astropy project, and would no longer be included as a sub-module.

Even though we can't make some of these changes in the core package yet, we wanted to start the discussion now so as to be able to make sure there are no issues in the current Python packaging ecosystem that would prevent these changes.

This is just a draft at this point, and we welcome detailed comments! If some points prove to be controversial, we'll try and set up a call so that we can discuss things and see if we can reach a consensus! Please also say if any aspects of this is confusing - some of us have spent far too much time thinking about this, and there may be some points that could be explained better.

### DIFF
--- a/APE17.rst
+++ b/APE17.rst
@@ -5,11 +5,11 @@ author: Stuart Mumford, Thomas Robitaille
 
 date-created: 2019 March 7
 
-date-last-revised: 2019 March 7
+date-last-revised: 2019 December 12
 
 type: Process
 
-status: Discussion
+status: Accepted
 
 .. _APE 4: https://zenodo.org/record/1043892
 .. _PEP 508: https://www.python.org/dev/peps/pep-0508/#environment-markers
@@ -708,5 +708,7 @@ Julia takes over.*
 Decision rationale
 ------------------
 
-<To be filled in by the coordinating committee when the APE is accepted or rejected>
-
+Following *extensive* discussion on the mailing list and Github (141 comments on
+this APE's PR...), and corresponding revision, this APE received unanimous
+approval from the attendees of  the 2019 Astropy Coordination meeting (including
+all members of the coordinating committee).
--- a/APE17.rst
+++ b/APE17.rst
@@ -122,8 +122,8 @@ optionally include a link to more information::
     $
 
 We could then provide a clear documentation page in a centralized place on how
-to use pytest and how to do common things the test command used to allow (e.g.
-running tests for specific sub-packages, or with remote data).
+to use pytest and how to do common things the test command used to allow e.g.
+running tests for specific sub-packages, or with remote data.
 
 Some flags such as ``--remote-data`` that worked with ``python setup.py test``
 should continue to work with pytest, and we could ensure that options passed to
@@ -603,17 +603,17 @@ affiliated and other packages using astropy-helpers change over to the new
 infrastructure.
 
 Once any issues have been ironed out in the core and coordinated packages, the
-migration guide should be advertised widely to all maintainers of packages
-using astropy-helpers. Throughout this process, astropy-helpers should continue
+migration guide will be advertised widely to all maintainers of packages
+using astropy-helpers. Throughout this process, astropy-helpers will continue
 to be supported, and once the migration guide is advertised the final date for
-support of astropy-helpers should be made clear. The recommendation in this
+support of astropy-helpers must be made clear. The recommendation in this
 APE is to support astropy-helpers until the end of the 4.0 LTS cycle at the
 latest (end of 2021), but this can be re-assessed depending on the ease of
 migration and the feasibility of maintaining astropy-helpers if there are
 extensive breaking changes in its dependencies.
 
-Impact for users
-----------------
+Impact for users who are not developers
+---------------------------------------
 
 If done properly, these changes should have no noticeable impact for users.
 Users will still be able to pip install (or conda install when available)
@@ -690,7 +690,20 @@ developers do not make any changes.
 Alternatives
 ------------
 
-In the words of Stuart: *Do nothing, suffer submodules and maintaining the
+It is likely that continuing changes to ``pip`` and other Python packaging
+infrastructure will cause things in ``astropy_helpers`` to break (see
+[#501](https://github.com/astropy/astropy-helpers/issues/501)). This will
+introduce a progressively increasing maintenance burden as our hacks become less
+and less standard.
+
+It is also possible but [explicitly discouraged by
+tox](https://tox.readthedocs.io/en/latest/example/basic.html#integration-with-setup-py-test-command)
+to overload the ``setup.py test`` command to run tox. While also possible to
+continue to overload these setup commands, doing this in a way that doesn't
+involve copying files between repositories would continue to mandate the use of
+the submodule.
+
+Finally, in the words of Stuart: *Do nothing, suffer submodules and maintaining the
 spaghetti code of astropy-helpers until we demoralise the whole community and
 Julia takes over.*
 
--- a/APE17.rst
+++ b/APE17.rst
@@ -16,6 +16,7 @@ status: Discussion
 .. _PEP 517: https://www.python.org/dev/peps/pep-0517/
 .. _PEP 518: https://www.python.org/dev/peps/pep-0518/
 .. _tox: https://tox.readthedocs.io/en/latest/
+.. _pip: https://pip.pypa.io
 .. _setuptools: https://setuptools.readthedocs.io/en/latest/
 .. _setuptools_scm: https://github.com/pypa/setuptools_scm
 .. _astropy-helpers: https://astropy-helpers.readthedocs.io/en/stable/
@@ -36,13 +37,14 @@ GitHub appear to use astropy-helpers).
 The original APE proposing the creation of astropy-helpers (`APE 4`_) was
 accepted in June 2014. Since then, the state of the Python packaging ecosystem
 has changed significantly, with advances such as binary wheels that can be
-provided for all major platforms, tools such as `tox`_ and `setuptools_scm`_,
-and more recently support for `PEP 517`_ and `PEP 518`_ (in pip 19.0 or later),
-and many incremental updates and new features to `setuptools`_ (in setuptools
-30.3 or later). The purpose of this APE is to show how by relying on these
-tools, we can significantly simplify the packaging for astropy and all the
-affiliated packages and other packages using astropy-helpers, and remove the
-need for astropy-helpers altogether in pure-Python packages.
+provided for all major platforms, tools such as `tox`_ and `setuptools_scm`_
+which are now stable and mature, and more recently support for `PEP 517`_ and
+`PEP 518`_ (in pip 19.0 or later), and many incremental updates and new features
+to `setuptools`_ (in setuptools 30.3 or later). The purpose of this APE is to
+show how by relying on these tools, we can significantly simplify the packaging
+for astropy and all the affiliated packages and other packages using
+astropy-helpers, and remove the need for astropy-helpers altogether in
+pure-Python packages.
 
 Detailed description
 --------------------
@@ -72,7 +74,7 @@ the scenes built the package, installed it into a temporary directory, installed
 pytest-astropy if needed, and ran the tests there.
 
 However, this can now be done more cleanly using `tox`_. Tox is a tool that
-makes is easy to install packages into a temporary clean Python environment
+makes it easy to install packages into a temporary clean Python environment
 along with the dependencies declared in setup.cfg, and run a set of custom
 commands - in a sense it does what the test command from astropy-helpers did,
 but in a more general way (including the use of an isolated test environment).
@@ -162,7 +164,7 @@ described in its documentation and we therefore do not repeat here. However, one
 subtlety is that since it relies on tags to determine versions, for packages
 such as the core astropy package which have all their tags on branches (at least
 in recent years), we will need to add a ‘developer’ tag on ``master`` straight
-after branching, e.g. after creating a ``v4.0.x`` branch we should tag the next
+after branching, e.g., after creating a ``v4.0.x`` branch we should tag the next
 commit on ``master`` as ``v4.1.dev``.
 
 Package data and entry points
@@ -375,16 +377,16 @@ Timeline for changes
 
 Many of the changes described here can already be safely made now. The only
 change that requires some consideration in terms of timeline is the use of
-``pyproject.toml`` (which is only needed for package with compiled extensions).
+``pyproject.toml`` (which is only needed for packages with compiled extensions).
 Full support for this file, including the environment markers (which allows
 different Numpy dependencies for different Python versions for example) and the
 build isolation, only became available in pip 19.0 onwards.
 
-Therefore, we recommend that Pure-python packages can make all the changes
+Therefore, we recommend that pure-Python packages can make all the changes
 described here now. On the other hand, packages that need to rely on
 ``pyproject.toml`` for building C extensions should make the switch at the time
 they are happy to rely on pip 19.0 (released 2019-01-22) for installs from
-source distributions, but they can still make other changes now, e.g. using
+source distributions, but they can still make other changes now, e.g., using
 ``setuptools_scm`` or making a greater use of ``setup.cfg``.
 
 We note however that provided that wheels are available for a package with
@@ -433,7 +435,7 @@ Users who wish to contribute fixes to the core astropy package or other packages
 will be encouraged to have `tox`_ installed if they want to easily run
 tests or build documentation locally. However, this is an easy package to
 install with `pip`_ and we could also add code in `setup.py` so that running
-e.g. ``python setup.py test`` or ``python setup.py build_docs`` gives a helpful
+``python setup.py test`` or ``python setup.py build_docs`` gives a helpful
 error message with instructions on updating and using tox to run tests and build
 the documentation.
 
--- a/APE17.rst
+++ b/APE17.rst
@@ -319,10 +319,8 @@ in recent years), we will need to add a 'developer' tag on ``master`` straight
 after branching, e.g., after creating a ``v4.0.x`` branch we should tag the next
 commit on ``master`` as ``v4.1.dev``. In any case, the documentation about `how
 to release the core package
-<http://docs.astropy.org/en/stable/development/releasing.html>`_ as well as
-`packages using the astropy package template
-<http://docs.astropy.org/en/stable/development/astropy-package-template.html#releasing-a-python-package>`_
-will need to be updated to reflect the use of setuptools_scm.
+<http://docs.astropy.org/en/stable/development/releasing.html>`_ will need to
+be updated to reflect the use of setuptools_scm.
 
 Package data and entry points
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/APE17.rst
+++ b/APE17.rst
@@ -137,18 +137,26 @@ For developers who do not want to always use tox to run the test, it is also
 possible to run pytest directly. For pure-Python packages, developers can run,
 e.g.::
 
-    pytest packagename
+    pytest
+
+Provided that the following are defined in ``setup.cfg``::
+
+    doctest_plus = enabled
+    addopts = --doctest-rst
+
+then running ``pytest`` will pick up the tests from the ``rst`` docs in addition
+to the package tests.
 
 For developers working on packages with compiled extensions, the preferred
 solution is to install the package in editable mode before running the tests::
 
     pip install -e .
-    pytest packagename
+    pytest
 
 Alternatively, one can build the extensions in-place before running the tests::
 
     python setup.py build_ext --inplace
-    pytest packagename
+    pytest
 
 The latter can also be simplified using a plugin such as `pytest-build
 <https://github.com/astrofrog/pytest-build>`_ which allows this example to be
@@ -164,6 +172,68 @@ dependencies, data files, entry points, etc.
 We note that these changes have no impact on the availability of the
 ``package.test()`` function which is unrelated to astropy-helpers and relies
 instead on the astropy core package to provide a test runner.
+
+Below we provide a table of correspondence between options passed to
+``python setup.py test`` and options passed to pytest (options passed to
+``package.test()`` would be unchanged):
+
++----------------------------------+--------------------------------------+
+| **python setup.py test option**  | **pytest equivalent**                |
++----------------------------------+--------------------------------------+
+| ``--args (-a)``                  | No longer needed since this just     |
+|                                  | passed arguments to pytest           |
++----------------------------------+--------------------------------------+
+| ``--coverage (-c)``              | ``--cov package --cov-report html``  |
+|                                  | with pytest-cov plugin               |
++----------------------------------+--------------------------------------+
+| ``--docs-path``                  | No longer needed since can run pytest|
+|                                  | directly on rst files                |
++----------------------------------+--------------------------------------+
+| ``--open-files (-o)``            | ``--open-files``                     |
+|                                  | with pytest-openfiles plugin         |
++----------------------------------+--------------------------------------+
+| ``--package (-P)``               | Could be added as a pytest plugin    |
++----------------------------------+--------------------------------------+
+| ``--parallel (-j)``              | ``-n``                               |
+|                                  | with pytest-xdist plugin             |
++----------------------------------+--------------------------------------+
+| ``--pastebin (-b)``              | ``--pastebin``                       |
++----------------------------------+--------------------------------------+
+| ``--pdb (-d)``                   | ``--pdb``                            |
++----------------------------------+--------------------------------------+
+| ``--pep8 (-8)``                  | ``--pep8``                           |
+|                                  | with pytest-pep8 plugin              |
++----------------------------------+--------------------------------------+
+| ``--plugins``                    | ``PYTEST_PLUGINS`` environment       |
+|                                  | variable                             |
++----------------------------------+--------------------------------------+
+| ``--readonly``                   | Not easily translatable but could    |
+|                                  | find other ways to prevent tests     |
+|                                  | from creating files                  |
++----------------------------------+--------------------------------------+
+| ``--remote-date (-R)``           | ``--remote-data``                    |
+|                                  | with pytest-remotedata plugin        |
++----------------------------------+--------------------------------------+
+| ``--repeat``                     | ``--count``                          |
+|                                  | with pytest-repeat plugin            |
++----------------------------------+--------------------------------------+
+| ``--skip-docs``                  | No longer needed since can specify   |
+|                                  | ``pytest package`` to avoid docs     |
++----------------------------------+--------------------------------------+
+| ``--temp-root``                  | No longer needed since equivalent is |
+|                                  | to use tox which controls the        |
+|                                  | temporary environment                |
++----------------------------------+--------------------------------------+
+| ``--test-path``                  | Specify path to files directly       |
++----------------------------------+--------------------------------------+
+| ``--verbose-results (-V)``       | ``-v``                               |
++----------------------------------+--------------------------------------+
+
+An example implementation of the ``--package (-P)`` option as a pytest plugin
+is provided in `astropy/pytest-astropy#19 <https://github.com/astropy/pytest-astropy/pull/19>`_.
+
+In addition to the options above, using pytest directly gives easy access to the
+whole ecosystem of pytest plugins and associated command-line options and flags.
 
 Building documentation
 ~~~~~~~~~~~~~~~~~~~~~~
@@ -497,8 +567,8 @@ and their proposed replacement:
 | astropy-helpers provided by git     |                                     |
 | submodule                           |                                     |
 +-------------------------------------+-------------------------------------+
-+ ``--use-system-*`` options          | ``ASTROPY_USE_SYSTEM_*``            |
-+                                     | environment variables               |
+| ``--use-system-*`` options          | ``ASTROPY_USE_SYSTEM_*``            |
+|                                     | environment variables               |
 +-------------------------------------+-------------------------------------+
 
 We propose that a new package called extension-helpers be created starting from
--- a/APE17.rst
+++ b/APE17.rst
@@ -519,10 +519,11 @@ by doing e.g.:
 Branches and pull requests
 --------------------------
 
+* The `extension-helpers <https://github.com/astropy/astropy-helpers/tree/extension-helpers>`_
+  branch in the astropy-helpers contains the minimal set of functionality still
+  needed for packages with C extensions.
 * The pull request `astrofrog/astropy#83 <https://github.com/astrofrog/astropy/pull/83>`_
-  shows the changes necessary for the astropy core package, and includes an
-  experimental version of `extension-helpers
-  <https://github.com/astrofrog/extension-helpers>`_.
+  shows the changes necessary for the astropy core package.
 * The pull request `astropy/photutils#915 <https://github.com/astropy/photutils/pull/915>`_
   shows the changes necessary for the photutils package, which includes
   defining extensions directly without using extension-helpers.
--- a/APE17.rst
+++ b/APE17.rst
@@ -1,0 +1,482 @@
+A roadmap for package infrastructure without astropy-helpers
+------------------------------------------------------------
+
+author: Stuart Mumford, Thomas Robitaille
+
+date-created: 2019 March 7
+
+date-last-revised: 2019 March 7
+
+type: Process
+
+status: Discussion
+
+.. _APE 4: https://zenodo.org/record/1043892
+.. _PEP 508: https://www.python.org/dev/peps/pep-0508/#environment-markers
+.. _PEP 517: https://www.python.org/dev/peps/pep-0517/
+.. _PEP 518: https://www.python.org/dev/peps/pep-0518/
+.. _tox: https://tox.readthedocs.io/en/latest/
+.. _setuptools: https://setuptools.readthedocs.io/en/latest/
+.. _setuptools_scm: https://github.com/pypa/setuptools_scm
+.. _astropy-helpers: https://astropy-helpers.readthedocs.io/en/stable/
+.. _cython: https://cython.org/
+.. _pytest: https://pytest.org/
+.. _jinja2: http://jinja.pocoo.org/docs/2.10/
+.. _package-template: https://github.com/astropy/package-template
+
+Abstract
+--------
+
+The `astropy-helpers`_ package contains a collection of utilities intended to
+help build and distribute the core astropy package, as well as coordinated and
+affiliated packages in the Astropy ecosystem, and it is also used by packages
+outside of the ecosystem (at the time of writing, almost 400 repositories on
+GitHub appear to use astropy-helpers).
+
+The original APE proposing the creation of astropy-helpers (`APE 4`_) was
+accepted in June 2014. Since then, the state of the Python packaging ecosystem
+has changed significantly, with advances such as binary wheels that can be
+provided for all major platforms, tools such as `tox`_ and `setuptools_scm`_,
+and more recently support for `PEP 517`_ and `PEP 518`_ (in pip 19.0 or later),
+and many incremental updates and new features to `setuptools`_ (in setuptools
+30.3 or later). The purpose of this APE is to show how by relying on these
+tools, we can significantly simplify the packaging for astropy and all the
+affiliated packages and other packages using astropy-helpers, and remove the
+need for astropy-helpers altogether in pure-Python packages.
+
+Detailed description
+--------------------
+
+We now go through the various aspects of package infrastructure that
+astropy-helpers has solved in the past and how we can now address these instead
+using standard packaging tools.
+
+Changes common to all packages
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We first take a look at what changes are needed for all packages, regardless of
+whether they have compiled extensions or not. For pure-Python packages, if all
+these changes are made, astropy-helpers is no longer needed and can be removed
+from the repository.
+
+Running tests
+~~~~~~~~~~~~~
+
+Astropy-helpers provides the ability for anyone to run tests in the source
+distribution to do this with the::
+
+    python setup.py test
+
+command, which behind
+the scenes built the package, installed it into a temporary directory, installed
+pytest-astropy if needed, and ran the tests there.
+
+However, this can now be done more cleanly using `tox`_. Tox is a tool that
+makes is easy to install packages into a temporary clean Python environment
+along with the dependencies declared in setup.cfg, and run a set of custom
+commands - in a sense it does what the test command from astropy-helpers did,
+but in a more general way (including the use of an isolated test environment).
+For example, by putting the following into ``tox.ini``::
+
+    [testenv:test]
+    deps =
+      pytest-astropy
+    commands =
+      pytest aplpy
+
+developers will then easily be able to run the tests with::
+
+    tox -e test
+
+The benefit of using tox over just running pytest directly is that since behind
+the scenes it creates a source distribution of the package and installs it into
+a clean environment, it ensures that source distributions for the package are
+fully functional and that all required files and dependencies are declared. For
+packages with C/Cython extensions it should also guarantee that they are always
+rebuilt before running the tests.
+
+However, for pure-Python packages, developers can also opt to simply run the
+tests directly with pytest, e.g.::
+
+    pytest packagename
+
+so using tox is optional.
+
+We note that these changes have no impact on the availability of the
+``package.test()`` function which is unrelated to astropy-helpers and relies
+instead on the astropy core package to provide a test runner.
+
+Building documentation
+~~~~~~~~~~~~~~~~~~~~~~
+
+Astropy-helpers provides a ``python setup.py build_docs`` command that
+behind-the-scenes built the package then added it to ``sys.path``, then ran the
+documentation build. Having the package be importable is needed as documentation
+often includes API sections that are dynamically created based on the package.
+
+As for testing, we can accomplish the same process more cleanly instead using
+`tox`_ - a minimal tox configuration might look like::
+
+    [testenv:build_docs]
+    deps =
+      sphinx-astropy
+    commands =
+      sphinx-build docs docs/_build/html -W -b html
+
+and this would be run using::
+
+    tox -e build_docs
+
+Developers not wishing to use tox could also accomplish the same by doing::
+
+    pip install -e .
+    cd docs
+    make html
+
+These changes will have no impact on ReadTheDocs as that service never made use
+of the ``build_docs`` command, instead invoking sphinx directly with
+``sphinx-build``.
+
+Version helpers
+~~~~~~~~~~~~~~~
+
+Currently, astropy-helpers includes version helpers that take the version
+defined in ``setup.cfg`` or ``setup.py`` and add a developer string when in a
+checked out version of a git repository. The developer string consists of
+``version.devN`` where N is the number of commits since a release.
+
+The Python Packaging Authority (PyPA) now provide a setuptools extension called
+`setuptools_scm`_ which can entirely replace the version helpers in
+astropy-helpers. The way this package works is that versions are no longer
+specified in ``setup.cfg`` or ``setup.py`` - instead the versions are taken from
+tags. Developer version strings produced in this way are much more sophisticated
+and can indicate for example if the working copy is clean or has local changes,
+and whether it is a stable tagged version or a developer version. We note that
+the default way of setting up setuptools_scm results in the version string not
+being updated automatically in ‘editable’ installs of a package (i.e. ``pip
+install -e .``), however `a workaround
+<https://github.com/pypa/setuptools_scm/issues/273>`_ is available.
+
+Switching to `setuptools_scm`_ requires minimal configuration, which is well
+described in its documentation and we therefore do not repeat here. However, one
+subtlety is that since it relies on tags to determine versions, for packages
+such as the core astropy package which have all their tags on branches (at least
+in recent years), we will need to add a ‘developer’ tag on ``master`` straight
+after branching, e.g. after creating a ``v4.0.x`` branch we should tag the next
+commit on ``master`` as ``v4.1.dev``.
+
+Package data and entry points
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+At the moment, package data and entry points for packages can be defined via
+``get_package_data`` and ``get_entry_points`` in ``setup_package.py`` files.
+However, this adds unnecessary complexity, as even for the core package it is
+simple to define the data and entry points in ``setup.cfg`` using the
+``[options.entry_points]`` and ``[options.package_data]`` sections, e.g.::
+
+    [options.entry_points]
+    console_scripts =
+        fits2bitmap = astropy.visualization.scripts.fits2bitmap:main
+    ...
+
+    [options.package_data]
+    astropy = astropy.cfg, CITATION, **/data/**/*
+    ...
+
+Removing the ability to specify package data in `setup_package.py` files removes
+the dependence of the `setup.py sdist` command on astropy-helpers, which is
+essential to using the PEP 518 build time dependencies discussed below.
+
+Using ``setup.cfg`` to define package data and other options does rely on
+setuptools 30.3 or later, which is now over two years old. Nevertheless, to
+minimize issues for users with older Python installations, we recommend
+including a version check for setuptools inside the ``setup.py`` file.
+
+Removal of pre/post-processing hooks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A little-known feature of astropy-helpers is the ability to define hooks in
+``setup_package.py`` files for steps to be carried out before/after specific
+``setup.py`` commands. We propose removing this functionality since this feature
+was never well advertised and likely only used in the core astropy package, and
+our proof-of-concept implementation in the core package shows that we can easily
+remove this.
+
+Changes for packages with compiled extensions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Build-time dependencies
+~~~~~~~~~~~~~~~~~~~~~~~
+
+One of the common issues that packages have regularly run into and which `PEP
+518`_ solves is how to define dependencies required for building a package. The
+``setup()`` function provided by setuptools takes a ``setup_requires`` argument
+which can include dependencies for the build, but unfortunately the ``setup.py``
+file has to be executed before the ``setup()`` function is run, which leads to
+the circular issue that any code in ``setup.py`` can’t rely on packages
+specified in ``setup_requires`` - thus, ``setup_requires`` does not properly
+solve the issue of build-time dependencies since those dependencies are only
+installed part way through the build process.
+
+PEP 518 instead specifies that build-time dependencies can be specified in a
+file called ``pyproject.toml`` that contains for example::
+
+    [build-system]
+    requires = ["setuptools", "wheel", "numpy==1.13.3"]
+
+Provided that the package is installed with a tool such as `pip`_ which
+understands this file, the build-time dependencies will be installed before the
+setup.py file is executed. `PEP 517`_ takes this concept further by specifying
+that the build should happen in an isolated environment, which means that one
+could specify a pinned version of `cython`_ or `jinja2`_ to use even if a
+different version is installed in the user’s environment.
+
+With this in mind, if astropy-helpers was still needed, it could therefore now
+be included as a build-time dependency in ``pyproject.toml`` which removes the
+need for the ``ah_bootstrap.py`` file and the git submodule. Furthermore,
+astropy-helpers could be pinned to specific versions in ``pyproject.toml`` and
+different packages could use different versions (thanks to the build isolation,
+this will not be a problem).
+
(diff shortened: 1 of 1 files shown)


APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+packages.
+
+Impact for contributors
+-----------------------
+
+Users who wish to contribute fixes to the core astropy package or other packages
+will be encouraged to have `tox`_ installed if they want to easily run
+tests or build documentation locally. However, this is an easy package to
+install with `pip`_ and we could also add code in `setup.py` so that running
+e.g. ``python setup.py test`` or ``python setup.py build_docs`` gives a helpful
```
bsipocz: This docs washer over on the fact that the current `python setup.py test` solution provides access to all the conveniences the astropy test runner has. Neither tox nor pytest provides any of those, so this proposal means removal of features many of the core devs relied on (e.g. don't underestimate the importance of e.g. `-P table,units` picks up all the tests from both subpackages and their documentations. It's not just much more typing but more error prone to need to think about where all the relevant tests live in a repo).

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+the scenes it creates a source distribution of the package and installs it into
+a clean environment, it ensures that source distributions for the package are
+fully functional and that all required files and dependencies are declared. For
+packages with C/Cython extensions it should also guarantee that they are always
+rebuilt before running the tests.
+
+However, for pure-Python packages, developers can also opt to simply run the
+tests directly with pytest, e.g.::
+
+    pytest packagename
```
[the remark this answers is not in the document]
astrofrog: We could always add these back via a pytest plugin in astropy (like the display plugin at the moment which all affiliates have access to)

APE17.rst
astrofrog: As mentioned above, we could add this back as a pytest option. What other test runner options do you find useful?

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+developers will then easily be able to run the tests with::
+
+    tox -e test
+
+The benefit of using tox over just running pytest directly is that since behind
+the scenes it creates a source distribution of the package and installs it into
+a clean environment, it ensures that source distributions for the package are
+fully functional and that all required files and dependencies are declared. For
+packages with C/Cython extensions it should also guarantee that they are always
+rebuilt before running the tests.
```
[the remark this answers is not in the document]
astrofrog: This was done before with ``python setup.py test`` too by the way, it's not something that would be new with tox. But I always run tests by doing:

```
python setup.py build_ext --inplace  # just do this once
pytest astropy/table
```

for example. Then unless I change the C code I can just use pytest directly on the source in the repo.

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+The `astropy-helpers`_ package contains a collection of utilities intended to
+help build and distribute the core astropy package, as well as coordinated and
+affiliated packages in the Astropy ecosystem, and it is also used by packages
+outside of the ecosystem (at the time of writing, almost 400 repositories on
+GitHub appear to use astropy-helpers).
+
+The original APE proposing the creation of astropy-helpers (`APE 4`_) was
+accepted in June 2014. Since then, the state of the Python packaging ecosystem
+has changed significantly, with advances such as binary wheels that can be
+provided for all major platforms, tools such as `tox`_ and `setuptools_scm`_,
```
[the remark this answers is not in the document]
astrofrog: Good point, we can rephrase this.

APE17.rst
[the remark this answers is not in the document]
bsipocz: for the record, `-R` is another such nice conveniences

APE17.rst
bsipocz: `-R` is another such nice conveniences

APE17.rst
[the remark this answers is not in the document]
bsipocz: so yo don't test the narrative docs?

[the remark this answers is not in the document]
bsipocz: > python setup.py build_ext --inplace  # just do this once
pytest astropy/table

a way to automate this would be great (e.g. the behaviour we have atm).

APE17.rst
bwinkel: What about the `--remote-data` option? Also, when I develop tests, I like to use the pytest options, which can be used via `python setup.py test -a <pytest args>`, e.g., `-x` to exit after first failure. (Several of my packages have C extensions, so running pytest directly wouldn't be possible.)

bsipocz: `-R` is short for `--remote-data` 😉

bwinkel: I know, it was just a race condition 😁

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+    commands =
+      pytest aplpy
+
+developers will then easily be able to run the tests with::
+
+    tox -e test
+
+The benefit of using tox over just running pytest directly is that since behind
+the scenes it creates a source distribution of the package and installs it into
+a clean environment, it ensures that source distributions for the package are
```
taldcroft: Does this mean it will need to rebuild the entire astropy every time you want to run tests in this way?

taldcroft: Sorry, should have scrolled down two lines further.  So I think the answer is that developers need to manage the build themselves in order to do repetitive light-weight testing.

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+solve the issue of build-time dependencies since those dependencies are only
+installed part way through the build process.
+
+PEP 518 instead specifies that build-time dependencies can be specified in a
+file called ``pyproject.toml`` that contains for example::
+
+    [build-system]
+    requires = ["setuptools", "wheel", "numpy==1.13.3"]
+
+Provided that the package is installed with a tool such as `pip`_ which
```
mwcraig: Looks like you haven't defined the reference for this pip link.

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+install -e .``), however `a workaround
+<https://github.com/pypa/setuptools_scm/issues/273>`_ is available.
+
+Switching to `setuptools_scm`_ requires minimal configuration, which is well
+described in its documentation and we therefore do not repeat here. However, one
+subtlety is that since it relies on tags to determine versions, for packages
+such as the core astropy package which have all their tags on branches (at least
+in recent years), we will need to add a ‘developer’ tag on ``master`` straight
+after branching, e.g. after creating a ``v4.0.x`` branch we should tag the next
+commit on ``master`` as ``v4.1.dev``.
```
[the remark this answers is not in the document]
mwcraig: It seems like this should be noted somewhere, though, or it will be forgotten in the future. Maybe part of release instructions?

olebole:
+1 since it simplifies packaging.

hamogu:
+1 I also note that of the 400 packages that use astropy-helpers, I'm probably responsible for 8, only 2 of which are active and one 1 of which is not pure-python. (I'll try to remember to delete experiemental remaining packages from github ...)

mhvk:
Another :+1: though like @bsipocz I use `./setup.py test -P <module>` a *lot* (also for my own package), and like that it does the docs as well; I also use `--remote-data` somewhat regularly.

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+The changes described here are opt-in, so barring any breaking changes in
+Python, setuptools, or Sphinx, everything should continue to work as expected if
+developers do not make any changes.
+
+Alternatives
+------------
+
+In the words of Stuart: *Do nothing, suffer submodules and maintaining the
+spaghetti code of astropy-helpers until we demoralise the whole community and
+Julia takes over.*
```
dpshelio: :point_up_2: @giordano @helgee @kbarbary are you really planning to take over?

giordano: :joy:

APE17.rst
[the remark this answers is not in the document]
eteq: It also seems a bit problematic because it means *affiliated* packages would have to start doing this.  And many affiliated packages do not have standardized release procedures/instructions...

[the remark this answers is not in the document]
astrofrog: No I don't think affiliated packages would need to do this. This issue only occurs for packages that have a workflow where they tag releases in branches, but in my experience most affiliated packages don't do release branches.

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
++-------------------------------------+ `PEP 518`_                          |
+| Transpilation of Cython to C before |                                     |
+| sdist                               |                                     |
++-------------------------------------+                                     |
+| Package specific versions of        |                                     |
+| astropy-helpers provided by git     |                                     |
+| submodule                           |                                     |
++-------------------------------------+-------------------------------------+
+
+We propose that a new package called extension-helpers be created starting from
```
pllim: I am not convinced yet another package is necessary. Instead, we can show them how things should be done "natively" over at template?

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+--------------------
+
+Many of the changes described here can already be safely made now. The only
+change that requires some consideration in terms of timeline is the use of
+``pyproject.toml`` (which is only needed for package with compiled extensions).
+Full support for this file, including the environment markers (which allows
+different Numpy dependencies for different Python versions for example) and the
+build isolation, only became available in pip 19.0 onwards.
+
+Therefore, we recommend that Pure-python packages can make all the changes
```
pllim:
```suggestion
Therefore, we recommend that pure-Python packages can make all the changes
```

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+described here now. On the other hand, packages that need to rely on
+``pyproject.toml`` for building C extensions should make the switch at the time
+they are happy to rely on pip 19.0 (released 2019-01-22) for installs from
+source distributions, but they can still make other changes now, e.g. using
```
pllim:
```suggestion
source distributions, but they can still make other changes now, e.g., using
```

APE17.rst
pllim:
```suggestion
``python setup.py test`` or ``python setup.py build_docs`` gives a helpful
```

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+
+    [testenv:test]
+    deps =
+      pytest-astropy
+    commands =
+      pytest aplpy
+
+developers will then easily be able to run the tests with::
+
+    tox -e test
```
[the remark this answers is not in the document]
StanczakDominik: I recently restructured PlasmaPy's CI setup to use Tox following SunPy's example - there was nothing I needed to change in the tests themselves.

APE17.rst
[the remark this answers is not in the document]
astrofrog: Actually setuptools_scm was only released in 2015, but tox has been around a lot longer

APE17.rst
astrofrog: Fixed

APE17.rst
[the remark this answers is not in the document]
astrofrog: @StanczakDominik is right, this doesn't require restructuring the tests - all we need is to write a ``tox.ini`` file but no changes are needed to the package or tests.

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+``setup()`` function provided by setuptools takes a ``setup_requires`` argument
+which can include dependencies for the build, but unfortunately the ``setup.py``
+file has to be executed before the ``setup()`` function is run, which leads to
+the circular issue that any code in ``setup.py`` can’t rely on packages
+specified in ``setup_requires`` - thus, ``setup_requires`` does not properly
+solve the issue of build-time dependencies since those dependencies are only
+installed part way through the build process.
+
+PEP 518 instead specifies that build-time dependencies can be specified in a
+file called ``pyproject.toml`` that contains for example::
```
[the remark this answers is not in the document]
astrofrog: There have been some issues with some versions of pip, but I don't think these would affect a simple case like what we need here.

APE17.rst
astrofrog: Fixed

APE17.rst
astrofrog: Fixed

APE17.rst
astrofrog: Fixed

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+many as possible of the options for the ``setup()`` function in ``setup.py`` to
+the ``setup.cfg`` file as described in `the setuptools documentation
+<https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files>`_.
+
+Timeline for changes
+--------------------
+
+Many of the changes described here can already be safely made now. The only
+change that requires some consideration in terms of timeline is the use of
+``pyproject.toml`` (which is only needed for package with compiled extensions).
```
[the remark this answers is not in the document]
astrofrog: Fixed

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+distribution to do this with the::
+
+    python setup.py test
+
+command, which behind
+the scenes built the package, installed it into a temporary directory, installed
+pytest-astropy if needed, and ran the tests there.
+
+However, this can now be done more cleanly using `tox`_. Tox is a tool that
+makes is easy to install packages into a temporary clean Python environment
```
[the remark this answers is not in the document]
astrofrog: Fixed

APE17.rst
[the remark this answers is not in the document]
astrofrog: I think we should leave this in the APE - the APE is *not* documentation but just a technical document for the project, and I think it's important to note this here. But I've added a sentence that says

*In any case, the documentation about how to release the core package as well astropy-affiliated packages will need to be updated to reflect the use of setuptools_scm.*

APE17.rst
astrofrog: Ok I've added a section on this, and a link to my PR to pytest-astropy to add the ``-P`` option for pytest: https://github.com/astropy/pytest-astropy/pull/19

APE17.rst
astrofrog: I've now added some notes about this and what solutions exist to make this as easy as possible.

APE17.rst
```diff
@@ -0,0 +1,482 @@
...
+Python version, environment markers can be used (see `PEP 508`_), e.g.::
+
+    "numpy==1.13.1; python_version<'3.7'",
+    "numpy==1.14.5; python_version>='3.7'",
+
+A side benefit of this is that with these pinnings in place, building wheels
+with ``pip wheel .`` will automatically create wheels compatible with all
+available versions of Numpy.
+
+Note that ``setup_requires`` should no longer be used in
```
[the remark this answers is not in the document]
astrofrog: it shouldn't be used at all, I've clarified

astrofrog:
I've made the following significant additions based on feedback above:

* I've added a recommendation that ``python setup.py test`` and ``python setup.py build_docs`` should not just crash but instead give useful information on how to run the tests and build the docs, and also provide a link to more detailed documentation online.

* I've added a mention of how developers could use pytest (for both pure-python packages and packages with extensions) without having to use tox. I've still made sure tox is the primary recommended way, but acknowledging that developers sometimes want to just do quick development/testing with pytest directly.

* I've added a mention of how we can replicate options like ``-P`` and ``-R`` with pytest and tox the latter of which can just pass command-line arguments to pytest).

* I've mentioned more explicitly that the release docs will need to be updated to reflect the updated workflow with setuptools_scm

* I've also made small updates to wording in places based on feedback here.

karllark:
Easier and more generic package maintenance is something I support.  The overhead of keeping up with this on my half dozen packages that use the package template/astropy-helpers is increasing with time.   Thumbs up!

bsipocz:
> Easier and more generic package maintenance is something I support. The overhead of keeping up with this on my half dozen packages that use the package template/astropy-helpers is increasing with time. Thumbs up!

I don't think it will get easier to keep generic files in sync with the template than using a submodule that is versioned and understood not to be edited (auto updating the helpers is always way easier than any of the other auto updates we do, as folks might have edited those files, but not the submodule). 
But either way, getting rid of the submodule, things definitely become more intuitive as the concept of submodules are not that widely known.

And of course it gets easier for us maintainers to not support functionality that is now widely available in generic tools.

pllim:
Can we accept this already? :smile: 

p.s. When it is time... https://www.atlassian.com/blog/git/git-submodules-workflows-tips#tip2

bsipocz:
@pllim - first we said that the PRs should be set up and work for a couple packages.

pllim:
Re: the PRs -- Ah okay, I see @larrybradley is already working on `photutils`. astropy/photutils#915

bsipocz:
yes, @astrofrog mentioned one being planned for ccdproc, too, and I might be overly ambitious, but having seen one for astropy would be nice (in addition to https://github.com/astropy/pytest-astropy/pull/19 being merged).

pllim:
So, @astrofrog , can you please list the PRs as acceptance criteria check list on the original post, and then when they are all checked off, we accept this?

astrofrog:
Just to avoid duplicated effort, I do have a branch of removing astropy-helpers in the core package here: https://github.com/astrofrog/astropy/pull/83 - but needs cleaning up.

Ultimately the CoCo will need to make the final decision, but I'll send another email out to astropy-dev to let people know I've updated it and it's ready for a re-review for last objections.

cdeil:
@astrofrog - Thanks!

I'm removing `astropy-helpers` from Gammapy: https://github.com/gammapy/gammapy/pull/2279

I tried adding `pyproject.toml`, but then what happened (I think) was that `pip install -e .` was creating a virtualenv with Numpy and Cython and using that to run the build. This is not what I want, I find this annoying. We have a [environment-dev.yml](https://github.com/gammapy/gammapy/blob/master/environment-dev.yml) and devs use that, it has Numpy and Cython. Is there a way to have a `pyproject.toml` but still by default use the existing Python env for developers? For released versions, it would be nice to have a `pyproject.toml`, and to have `pip install gammapy` work, instead of getting a Numpy or Cython related import error.

Another point is that we have a [Makefile](https://github.com/gammapy/gammapy/blob/master/Makefile) and just use the default dev env locally for tests, docs, etc. Some day we might start using tox, but for now what we have is fine and none of our devs knows or likes tox. One thing I'd suggest for the APE is to make it a bit clearer that e.g. using tox is a recommendation, not a requirement, to get rid of astropy-helpers.

One more thing that's unclear to me: what about in-progress pull requests and developer git checkouts? What commands do they have to run to get rid of the astropy-helpers git submodule? I think they have to manuallly remove it, and then if they have a mix of local branches before or after the `astropy-helpers` removal, they are in trouble? Maybe add a few lines and commands they need during the transition period (until there are no more branches with `astropy-helpers` around for a given project)?

astrofrog:
@cdeil - thanks for trying this out!

> I tried adding pyproject.toml, but then what happened (I think) was that pip install -e . was creating a virtualenv with Numpy and Cython and using that to run the build. This is not what I want, I find this annoying. We have a environment-dev.yml and devs use that, it has Numpy and Cython.

The isolated build environment is a feature recommended by PEP517 - see [here](https://www.python.org/dev/peps/pep-0517/#recommendations-for-build-frontends-non-normative), specifically:

*A build frontend SHOULD, by default, create an isolated environment for each build, containing only the standard library and any explicitly requested build-dependencies.*

However I thought that when doing ``-e`` pip didn't enforce build isolation - maybe @Cadair has some thoughts on this?

> Is there a way to have a pyproject.toml but still by default use the existing Python env for developers? For released versions, it would be nice to have a pyproject.toml, and to have pip install gammapy work, instead of getting a Numpy or Cython related import error.

It sounds like something weird is going on here - normally your pyproject file should declare numpy and cython as build dependencies, and your package should also define numpy as an install_requires, and you shouldn't get any errors related to numpy or cython. Can you point me to the error you were seeing?

> Another point is that we have a Makefile and just use the default dev env locally for tests, docs, etc. Some day we might start using tox, but for now what we have is fine and none of our devs knows or likes tox. One thing I'd suggest for the APE is to make it a bit clearer that e.g. using tox is a recommendation, not a requirement, to get rid of astropy-helpers.

Sounds good, I'll try and clarify that it's recommended but optional.

> One more thing that's unclear to me: what about in-progress pull requests and developer git checkouts? What commands do they have to run to get rid of the astropy-helpers git submodule? I think they have to manuallly remove it, and then if they have a mix of local branches before or after the astropy-helpers removal, they are in trouble? Maybe add a few lines and commands they need during the transition period (until there are no more branches with astropy-helpers around for a given project)?

Sounds good, submodules are indeed a pain to remove so I'll provide some clear instructions.

astrofrog:
Just as a reminder to myself, other projects such as Numpy are now seeing issues with shipping pre-cythonized files (https://github.com/numpy/numpy/issues/14403) which the approach recommended in the APE should fix.

cdeil:
It turned out that there were no transition issues with removing the git sub-package. All devs had to do was `rm -r astropy_helpers` and even switching back and forth with branches that had or didn't have the sub-module any more just worked (see 
 https://github.com/gammapy/gammapy/pull/2279#issuecomment-525180658).

We had one more small issue where we didn't manage to run `sphinx-build` from the top-level project folder to get rid of a sphinx warning, we had to `cd docs` and run the sphinx build there (see https://github.com/gammapy/gammapy/pull/2279#issuecomment-512762075). This could be worth mentioning somewhere (in the project template)?

Concerning `pyproject.toml` I looked around a bit:
- It seems that in April & May 2019 the main issues with `pip install -e` and `pyproject.toml` were resolved (see [here](At https://discuss.python.org/t/pip-19-1-and-installing-in-editable-mode-with-pyproject-toml/1553)).
- pip now has options [--no-use-pep517](https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-use-pep517) and [--no-build-isolation](https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-no-build-isolation)

So it looks like the solution for a project like Gammapy where a conda dev env is provided, if we want to add `pyproject.toml`, would be to recommend the use of `pip install --no-build-isolation -e .` (or put that command as `make install-dev` in our Makefile to make it slightly easier to remember and type).

I still don't understand why we'd want to pin `numpy==1.14.2` for the build. Why should a user have to download and build an old Numpy when they `pip install gammapy`, wouldn't it be better to use the same Numpy that they have in the env where Numpy & Astropy & Gammapy is being installed. Why can't pip simply use the info and install Numpy first? Is it really the idea to always install two Numpy on all user machines just to install a package that has Cython & Numpy code?

astrofrog:
> I still don't understand why we'd want to pin numpy==1.14.2 for the build

Just to be clear, this version is *only* used for building the package, and is not installed into the user's environment. The idea is to build the package with the oldest compatible version of Numpy. So for example, if you were to just put 'numpy' (with no version pinning) in pyproject.toml and 'numpy' in install_requires, then you could run into the following if you started with a clean environment

* ``pip install gammapy`` would temporarily download Numpy 1.17, build a gammapy wheel with that version of Numpy, then install the wheel into the user environment along with Numpy 1.17. So far so good.
* Now the user decides they want to downgrade Numpy to 1.15 because of some of the code, so ``pip install "numpy==1.15.*"``. Now if they try and import gammapy C extensions, they will get numpy errors since there was a compatibility break in 1.16.

So by pinning the oldest compatible version in ``pyproject.toml``, you effectively do the following:

* ``pip install gammapy`` would temporarily download Numpy 1.14, build a gammapy wheel with that version of Numpy, then install the wheel into the user environment along with Numpy **1.17** (taken from install_requires).
* Now the user decides they want to downgrade Numpy to 1.15 because of some of the code, so ``pip install "numpy==1.15.*"``. Now if they try and import gammapy C extensions, everything will work fine. In fact, they can downgrade Numpy to any compatible version and everything should be ok.

Does this make sense? The key here is that the version in ``pyproject.toml`` is just used when building a wheel, and in the same way that when building wheels for releases we need to use the oldest compatible Numpy version, we need to do this here too.

cdeil:
@astrofrog - So the Numpy version pinning makes sense when building and distributing binary wheels.

That's not so common for devs, I'm not sure using `pyproject.toml` for that use case is a great idea.

If you want to build and distribute binary wheels, then you probably want to do it for different platforms, and you don't want to do it on a dev machine. Rather you want to have something like the `meta.yml` in conda-forge, and their tooling. That's where the Numpy version is specified to be compatible across conda-forge packages, it's not using `pyproject.toml`.

Now pip also builds binary wheels locally and installs those. For that use case, I prefer if the Numpy from my existing environment is used. This way I avoid old and possibly buggy or not as performant Numpy versions (not sure if that matters in practice), and just the extra time spent to make a separate build env and to install Numpy.

Note that it's not obvious which Numpy to pin against. E.g. in this APE you have `numpy=1.13.3` in the example. I'm on Python 3.7, and for that Numpy no binary wheels were build: https://pypi.org/project/numpy/1.13.3/#files
Now for me on macOS today, `pip` did manage to build that old Numpy from source and I think it linked against the Apple BLAS libs which is no longer recommended. But I can imagine there's a pretty good chance this install of old Numpy from source will fail on some machines.

While I love the idea of having a `pyproject.toml` to solve the problem we had forever in Python that `pip` didn't install Numpy before installing packages with C extensions, I'm not sold on the idea of this separate build env and doing extra builds with old Numpy versions on user machines.

astrofrog:
@cdeil - I understand your concerns, so I'll be sure to mention the flag to skip the build isolation.

> Note that it's not obvious which Numpy to pin against. E.g. in this APE you have numpy=1.13.3 in the example

Search for PEP 508 in this APE, as this will show an example of how to specify different versions of Numpy versions for different Python versions using environment markers. The oldest version that should be specified should be the oldest one with wheels for that Python version.

> While I love the idea of having a pyproject.toml to solve the problem we had forever in Python that pip didn't install Numpy before installing packages with C extensions, I'm not sold on the idea of this separate build env and doing extra builds with old Numpy versions on user machines.

One interesting point is that since most *users* actually install from wheels (and therefore don't have to do builds anyway) most users will be using packages built with the oldest compatible Numpy version and with a different installed Numpy version, so in fact if you use the oldest Numpy in pyproject.toml you are actually doing something that will be closer to the average user of your package.

Cadair:
> This way I avoid old and possibly buggy or not as performant Numpy versions (not sure if that matters in practice)

I think where this matters would be a pretty small subset of changes to Numpy. What you are doing by specifying an old version of numpy is building your C extensions against the oldest version of the Numpy ABI. The actual code you execute will be the one in your final installed environment. So the only case you would take a performance penalty that I can think of is the case where worse performing code is kept in Numpy for the sole reason of ABI compatibility.

dhomeier:
> Just as a reminder to myself, other projects such as Numpy are now seeing issues with shipping pre-cythonized files ([numpy/numpy#14403](https://github.com/numpy/numpy/issues/14403)) which the approach recommended in the APE should fix.

Cf. my message on slack - looks like the numpy issue was just due to running Cython 0.29.11 instead of 0.29.13 on them; so hopefully it will still be possible to create sources that work with Pythons 3.6-3.8+, if necessary.

astrofrog:
@cdeil - I've tried to clarify some of the points you raised - you can see just the changes relevant to your comments at https://github.com/astropy/astropy-APEs/pull/52/commits/ff6d75fea1fcb3480f5fa082f8b06bc8ceaa6abe?short_path=7e96e0f#diff-7e96e0f54f0335dc6dfa73709a7c40d7

cdeil:
@astrofrog - Thanks!

mhvk:
I have been following this on and off, as I'd definitely love not to have a submodule anymore, but am also more than a little used to the current workflow. Reading the whole APE again, I see that much of my initial worries have now been addressed -- I'm almost ready to get rid of `astropy-helpers` in my own packages! 

But the on thing that is holding me back is still what I use most -- `./setup.py test`. Like @cdeil, I think it is important to distinguish testing during development and testing for continuous integration. In particular, during development, one really does not want to be slowed down by repeatedly installing in a clean environment, so it would seem that for that case using `pytest` directly is what should be recommended. Especially for astropy, one also does *not* want to test the whole package, but rather the submodule one is working on. Hence, preserving the ability to test a submodule with `-P` or equivalent is really important, and it would be very helpful if it continues to be possible to test the associated documentation at the same time. (For complete tests done before submitting PRs and for continuous integration, `tox` seems to be a good replacement.)

Also, a question: since it is already mentioned that one should let the old `setup.py` commands emit text that tells about the new state, and that, therefore, it is necessary to make a more complicated `setup.py`, would it not be possible instead to just let those commands *use* the new recommended procedure? I.e., could one get away with replacing `astropy-helpers` with a more complicated `setup.py`? I realize this partially beats the whole point, but if it means that for most users the transition is essentially invisible, that would be worth quite a bit too.

astrofrog:
@mhvk - could you try out the following plugin which adds a ``-P`` option for pytest: https://github.com/astropy/pytest-astropy/pull/19

Also I have tried to make it clear in the current wording that developers can choose to just run pytest directly:

```
For developers who do not want to always use tox to run the test, it is also
possible to run pytest directly. For pure-Python packages, developers can run,
e.g.::

    pytest packagename

For developers working on packages with compiled extensions, the preferred
solution is to install the package in editable mode before running the tests::

    pip install -e .
    pytest packagename

Alternatively, one can build the extensions in-place before running the tests::

    python setup.py build_ext --inplace
    pytest packagename

The latter can also be simplified using a plugin such as `pytest-build
<https://github.com/astrofrog/pytest-build>`_ which allows this example to be
written as::

    pytest packagename --build-inplace
```

and in fact this will be faster than the current ``python setup.py test`` since that was installing into a temporary directory every time. If you think we could phrase this better though, could you open a PR with suggestions for how to word things differently? For any packages other than the core package, tox will be fast though so I don't see the downside in still recommending that - there's a misconception that it's a CI tool when in fact it's just as useful for local use and it catches a number of packaging bugs that otherwise might go unnoticed (I've found issues on many packages where I've added a tox config).

For continuing to support ``python setup.py test`` as an alias, the issue is that we can't just rely on just wrapping pytest since that doesn't set up things like entry points directly, so we would then need to keep our whole test runner infrastructure which installs to a temporary directory (and one of the points of this APE was to reduce the amount of custom infrastructure we have to support) - or we could wrap tox, but this is explicitly discouraged [in the tox docs](https://tox.readthedocs.io/en/latest/example/basic.html#integration-with-setup-py-test-command). In addition, the idea is that in future ``setup.py`` commands are going to become discouraged anyway in Python packaging, so I personally prefer the approach of just telling people about the new way so that they can learn sooner rather than later.

olebole:
I would like to put another CI case here, which are integration tests for packages: For the Debian packages, it is very helpful to be able to test the complete package, on its final place, with all the dependencies installed. Our infrastructure then runs this whenever a dependency is changed, which is not limited to the Python dependencies, but also f.e. cfitsio or external programs.
This was a kind-of struggle with the `python setup.py test`, one then has to find out what really happened; but with pytest it is convenient (and standard, which makes my life easier).

mhvk:
@astrofrog - OK, point taken about it being better not to use `setup.py test` any more.

I now tried `tox` and indeed it is not much slower than the current `setup.py test` - as you say, both create a temporary environment. But at first, it didn't run the tests in the docs, and after adding a suitable `addopts = --doctest-rst` to `setup.cfg`, it fails all the tests... (somehow I seem to need `pytest-3` rather than `pytest`...). I was also not so pleased with now having 140MB worth of downloaded packages under `.tox`. I guess this echoes @cdeil's comment that I much prefer to just run tests with the standard system packages.

Anyway, a bit more reading of docs and experimenting led me to the following, which does roughly what I would need (separately: I just put it in `setup.cfg`; maybe eventually one would move to `pyproject.toml`, but I'd definitely want things together, so am not keen on `tox.ini`)
```
[testenv:test]
sitepackages = true
commands =
  python -m pytest scintillometry docs

[testenv:build_docs]
sitepackages = true
commands =
  python -m sphinx docs docs/_build/html -W -b html
```

p.s. One pity is that by default the nice pytest header disappears for affiliated packages - looks like a `conftest.py` is needed in the main directory - I'll raise a separate issue for that in `astropy-template`.

astrofrog:
For the header, I've noticed that before and I think this might be to actually make the header plugin into a proper plugin rather than require a ``conftest.py`` file to be added for all packages. Then we could opt in to it by having an option in the pytest section of ``setup.cfg`` and maybe even configure there what packages to show (cleaner than relying on ``conftest.py`` files).

astrofrog:
For the site-packages option of tox, note that you can also run tox with the ``--sitepackages`` flag (in case we don't want it to be included by default in CI for example).

mhvk:
:+1: to the display header plugin. It might then make sense to get the things listed from the dependencies already given. Could it be part of `pytest-astropy`?

And thanks for the `--sitepackages` flag. I guess easiest might be to have it set in the default "test" environment, but not set it in environments actually tried in CI. (I do like the idea that I can (try to) reproduce CI problems by trying `tox -e <CI env>`)

astrofrog:
@mhvk - regarding the display plugin - yes it could be part of pytest-astropy and in fact we had an issue open for it (https://github.com/astropy/pytest-astropy/issues/15). I'll work on this now.

mhvk:
One more thing I regularly use is `setup.py test --coverage` - would be good to add its replacement as well.

cdeil:
@mhvk - What about this?
```
pytest -v scintillometry --cov= scintillometry --cov-report=html
```
This will generate the coverage report as before, no?

Cadair:
Yeah and it works with xdist 😉

mhvk:
Thanks, indeed, after installing `python3-pytest-cov`, things work (removing the space after `cov=`), though it does look like I should set up a coverage section to exclude the test files themselves.

In any case, something to mention in the APE (and perhaps also add a dependency on `pytest-cov` in `pytest-astropy`?)

cdeil:
@mhvk - You can configure the coverage from a `.coverage` file, or put the config in `setup.cfg` (and probably other ways).

If you don't want to measure coverage for your test files, it's easy to omit.
Example of what we put (mostly copy & paste from the Astropy package template): https://github.com/gammapy/gammapy/blob/983d3055e477125bb2449fe1d6c669372a7a53b0/setup.cfg#L94-L118

mhvk:
OK, thanks. I can also see how it would be easy to create a `tox -e test-coverage` section.

Sorry, to keep on harassing, but another thing I like about `setup.py` is the built-in help. E.g., `setup.py test --help` tells me that I need to pass on `--coverage`. I guess `tox -a` may help a bit here.

cdeil:
@mhvk - You can get help on the options from pytest directly, it also includes the help on options from plugins.
```
$ pytest -h | grep cov
coverage reporting with distributed testing support:
  --cov=[SOURCE]        Path or package name to measure during execution
                        (multi-allowed). Use --cov= to not do any source
  --cov-report=TYPE     Type of report to generate: term, term-missing,
                        missing may be followed by ":skip-covered". annotate,
                        specifies the output location. Use --cov-report= to
  --cov-config=PATH     Config file for coverage. Default: .coveragerc
  --no-cov-on-fail      Do not report coverage if test run fails. Default:
  --no-cov              Disable coverage report completely (useful for
  --cov-fail-under=MIN  Fail if the total coverage is less than MIN.
  --cov-append          Do not delete coverage but append to current. Default:
  --cov-branch          Enable branch coverage.
                        discovery
                        prefixes or glob names for Python test class discovery
                        method discovery
```

For Gammapy what we've done is to add a [Makefile](https://github.com/gammapy/gammapy/blob/master/Makefile) so that all that devs and maintainers have to remember is to type `make` and get a help printout on all the common commands (see [here](https://gist.github.com/cdeil/b712a97cb0f7fed24a79a5d1aa62305d)). Really nothing else, just to type `make`.

I haven't used `tox` yet (except played with it briefly), but I think you could use it in the same way -- pre-define the common commands there, and also add some help printout with further information to make the live of developers easy.

mhvk:
Hmm, that `Makefile` is not a bad idea. Wonder if a more python specific standard will arise...

cdeil:
@mhvk - Makefiles are nice, but also have big issues. If you use spaces instead of tabs it gives cryptic errors, very weird for Python people. And it doesn't work on Windows.

I think the choice made here in the APE to recommend tox is probably the best option. It's not a standard, but the tagline at https://tox.readthedocs.io is "vision: standardize testing in Python" and it's widely used by many Python projects. There's also https://nox.thea.codes which seems nice if you want to automate your tasks with Python instead of a config file, but I think tox is much more widely used.

@astrofrog - Apologies for derailing the discussion here. Overall I'm strongly 👍 for this APE. There are issues and details that are complex, IMO especially concerning pyproject.toml and where builds take place in the dev workflow. My largest concern for distribution is that you recommend to pin on old Numpy versions for the build env, and then some users will get install errors because the build of that old Numpy will fail on their machine. E.g. you don't know if it'll work on the next version of macOS. But those are things we will see and figure out over the next years, the recommendations given here don't have to be the last word. So 👍  to merge.

Cadair:
> because the build of that old Numpy will fail on their machine

I think this should be a seriously edgey edge case, it should be using wheels for this install almost all the time, basically unless you are on a platform which doesn't have wheels built for it (i.e. alpine linux) or similar. When you pin these versions you know if the wheels exist or not, barring someone going back and removing them for old versions, they should stay there.

I agree with you that we can work this out as we go along (which really isn't any different from the equally (more?) complex but more familiar situation we have at the moment).

I think that having a controlled build environment for wheels and sdist fixes more problems than it's likely to introduce on developer's machines, but :man_shrugging:

cdeil:
@Cadair - the case of using a modern Python (3.7, 3.8, 4.0) is common, and then you won't have Numpy wheels for older Numpy versions.

Cadair:
You should update your pins in `pyproject.toml` to the latest version of numpy for a given Python with a wheel, which I think is in the APE. I appreciate this takes intervention at each Python release.

cdeil:
@Cadair - Did you really mean "latest" as most recent? I think @astrofrog was arguing that one should pin against as old versions of Numpy as possible?

When Python 3.8 comes out next month, and only after that release Numpy wheels for 3.8 will appear, what Numpy should one pin against?

Whatever the answer, this is really not a good solution, for each project maintainer to have to watch Python and Numpy releases and who supports which old Numpy version, every time before cutting a new release. I don't want to do that, that's for the packagers (like conda-forge or Debian) to figure out.

Cadair:
>  Did you really mean "latest" as most recent?

No I meant oldest for a given Python.

>  that's for the packagers (like conda-forge or Debian) to figure out.

If you are releasing binary wheels of your project you have taken on that role for PyPI packages. The build isolation is designed specifically to handle things that debian and conda are already doing for this reason.

If you aren't building binary wheels then I think it probably matters less what version you pin numpy to (or just don't at all), because in that situation all the build requirements are doing is making sure certain packages are installed before `setup.py` is executed so you can lookup the numpy headers etc.

cdeil:
Gammapy is one of the packages that isn't building binary wheels. I think that's the case for most Astropy affiliated packages.

So yes, adding a `pyproject.toml` without Numpy pinned is an option worth considering for us.
- Pro: Solves the installation order issue
- Con: Creates extra virtual envs during install for users

I'm still a bit fuzzy on what exactly would happen. Gammapy depends on `numpy`, `astropy`, `astropy-regions` and soon `astropy-healpix` which all have C extensions. So potentially if no wheels are available for some of those (which is common with new Python versions), the build of each of those would create their own venv and install Numpy several times during `pip install gammapy`, of those packages had `pyproject.toml` files?
Seems pointless if in the end they will all have to work together in the one user env that already exists when starting pip.

mhvk:
I'm also happy with the APE - I only would suggest adding a line or two about `--coverage` (maybe both the `tox` setup and the use with `pytest-cov` from @cdeil above) - it is the one thing we do in travis which currently is missing.

For eventual documentation (and perhaps for here), I would also swap the order: describe first how test/coverage/docs are done by the individual commands and only then describe how the current `astropy-helpers` + `setup.py`automation can be easily replaced with `tox`.

Cadair:
> which all have C extensions

Just for clarity, you only need to put packages in the `pyproject.toml` file if they are *build* time dependencies. So this is normally only `numpy` if you are building a C extension, or maybe `numpy` and `Cython` if you are build a Cython extension (and using numpy arrays in them, which I am assuming you are lol!).

> the build of each of those would create their own venv and install Numpy several times

When you are installing just from sdists, and you have a chain of dependencies which are linking against numpy, pip would build the numpy wheel once, and then use it for all the following builds. It would use multiple isolated build envs though. (Which is pretty cheap when you are only installing numpy, setuptools, wheel and maybe cython into it).

> Seems pointless if in the end they will all have to work together in the one user env that already exists when starting pip.

It allows each of those packages to use different versions of the build time dependencies though. For instance different Cython versions or different versions of `astropy_helpers` (or whatever a build time dep version of it gets called). Which is one of the things the submodule gives us and this mechanism makes redundant.

The overhead of doing this is really pretty low (maybe if you are unlucky one extra numpy compile time) and the advantages are massive both for wheels and for having control over your build environment, which is something setuptools has no way of doing without PEP 517.

astrofrog:
Just for anyone following this - my plan is now to write up some package maintainer-facing packaging docs that assume the infrastructure described here.

@mhvk: I think I'd rather keep the order of things as-is in the APE, and focus instead on writing good docs for maintainers - ideally I don't think maintainers should have to read the APE in detail, as APEs shouldn't be substitutes for documentation (and the order of things in the APE makes more sense to me from the point of view of a proposal for change).

mhvk:
@astrofrog - yes, definitely OK to keep the APE as is if you are writing more general documentation!

APE17.rst
```diff
@@ -0,0 +1,593 @@
...
+the default way of setting up setuptools_scm results in the version string not
+being updated automatically in ‘editable’ installs of a package (i.e. ``pip
+install -e .``), however `a workaround
+<https://github.com/pypa/setuptools_scm/issues/273>`_ is available.
+
+Switching to `setuptools_scm`_ requires minimal configuration, which is well
+described in its documentation and we therefore do not repeat here. However, one
+subtlety is that since it relies on tags to determine versions, for packages
+such as the core astropy package which have all their tags on branches (at least
+in recent years), we will need to add a 'developer' tag on ``master`` straight
```
pllim: @jdavies-st , does this sufficiently address your concern about dev tags?

jdavies-st: Yes, it does.  So that would be something like a `4.1.a0` tag on `master` so that the version `setuptools_scm` reports would be something meaningful, such as `astropy 4.1.a0.dev23` meaning that `master` is now 23 commits after version `4.0.0` was tagged (on its own branch).  That `a0` is important if you plan on using conda for anything with dev versions of astropy, but otherwise is probably not necessary.

Cadair: according to the version numbering PEP `.dev` is also fine for that, you would then get a version like `4.1.dev23`

pllim: `astropy` never plans to have a dev release on `conda` though, so is the `a0` relevant then?

jdavies-st: No.  We were doing nightly dev builds of conda packages for `jwst`, so it was relevant to us.  Not relevant here though. This looks good to me. 👍

APE17.rst
```diff
@@ -500,10 +502,14 @@ and their proposed replacement:
...
-auto-discovery of Cython extensions. This package would need to be declared as a
-build-time dependency in ``pyproject.toml``.
+astropy-helpers but with only functionality related to:
+
+* Collecting extension definitions from ``setup_package.py`` files
+* Auto-defining Cython extensions
+* Determining flags for OpenMP compilation
+
+This package would need to be declared as a build-time dependency in
+``pyproject.toml`` for packags that wish to use it.
```
Cadair: packages

bwinkel: It would be great, if the `extension-helpers` could also figure out build/linker flags for C++ (if used), because this is sometimes a really pain, especially, if C++11 features are desired and one wants to build on multiple platforms.

APE17.rst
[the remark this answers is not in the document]
astrofrog: I now mention the ``pytest-build`` plugin to do in-place builds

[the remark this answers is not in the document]
astrofrog: I've also changed the ``pytest`` command in the APE from ``pytest package`` to just ``pytest`` which will allow the rst docs to get picked up too.

APE17.rst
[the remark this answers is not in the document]
astrofrog: I've now added a table of correspondence for ``setup.py test`` to ``pytest`` options in https://github.com/astropy/astropy-APEs/pull/52/commits/8841b8aea62edab63cf19e029c96922cb1915f55 - and mention my PR for the ``-P`` option.

APE17.rst
astrofrog: I've added a table showing that most of the previous options/conveniences translate over to options with pytest: https://github.com/astropy/astropy-APEs/pull/52/commits/8841b8aea62edab63cf19e029c96922cb1915f55

APE17.rst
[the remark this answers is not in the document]
mhvk: Thanks for the table! I agree with @bsipocz that the `-P` option is the one that I feel I would miss most, so :+1: to having that as part of the `astropy-pytest` (or so) plugin.

APE17.rst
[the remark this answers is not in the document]
astrofrog: We are including instructions about how to use setuptools_scm with release branches in an upcoming packaging guide.

eteq:
Following some discussions at the Astropy Coordination Meeting (mostly with @astrofrog and myself), I've identified a few things to add to this APE (or possibly say "out of scope for the APE, but should do in the implementation stage"):

1. We should commit to a "migration guide" for users of the existing helpers to switch to the new scheme.  Tentatively it would be generated based on the Astropy core experience, and would be "released" once the process is done in the core.
2. We should state a deadline we're aiming for to end support for the "old" `astropy-helpers`.  2 years (i.e. Astropy LTS 5.0) might be a good deadline.  Dependent on the "migration guide" being done and demonstrated successful (i.e., "used" by some affiliated or coordinated package)
3. Relatedly, we should be sure the package template is up to date and at least as usable (maybe more, with these changes!) before we can declare victory on the transition.

astrofrog:
@eteq - I think the following section should now address all your points:

https://github.com/astropy/astropy-APEs/blob/7f2b4dabb9386a47697e0e4a0da5499804e75386/APE17.rst#timeline-for-changes

Please check whenever you can and confirm whether you agree!

taldcroft:
The new Timeline for changes section looks fine to me, so approving this.  In the second paragraph there are a few instances of the word "should" that might be changed to "shall" or "will" (e.g. " the migration guide should be advertised widely to all maintainers") to convey a more definitive requirement, but there is no doubt that these things will happen so this change is optional to me.

APE17.rst
```diff
@@ -0,0 +1,701 @@
...
+
+        pip install tox
+
+    For more information, see https://docs.astropy.org/...
+
+    $
+
+We could then provide a clear documentation page in a centralized place on how
+to use pytest and how to do common things the test command used to allow (e.g.
+running tests for specific sub-packages, or with remote data).
```
hamogu: I think the parentheses in this sentence should be removed.

APE17.rst
```diff
@@ -0,0 +1,701 @@
...
+subtlety is that since it relies on tags to determine versions, for packages
+such as the core astropy package which have all their tags on branches (at least
+in recent years), we will need to add a 'developer' tag on ``master`` straight
+after branching, e.g., after creating a ``v4.0.x`` branch we should tag the next
+commit on ``master`` as ``v4.1.dev``. In any case, the documentation about `how
+to release the core package
+<http://docs.astropy.org/en/stable/development/releasing.html>`_ as well as
+`packages using the astropy package template
+<http://docs.astropy.org/en/stable/development/astropy-package-template.html#releasing-a-python-package>`_
+will need to be updated to reflect the use of setuptools_scm.
```
eteq: I think the template needs to be updated for quite a bit more than just `setuptools_scm`, right?  That probably should be stated explicitly as a mandatory part of the process.

APE17.rst
eteq: I don't quite understand how this would work as @pllim is proposing.  Then wouldn't we have to push updates to *all* users of this stuff if e.g. we need to update the `OpenMP` stuff?

APE17.rst
eteq: I guess it's in the "timeline" section, although it's not clear exactly what makes this special, then?

APE17.rst
```diff
@@ -0,0 +1,701 @@
...
+support of astropy-helpers should be made clear. The recommendation in this
+APE is to support astropy-helpers until the end of the 4.0 LTS cycle at the
+latest (end of 2021), but this can be re-assessed depending on the ease of
+migration and the feasibility of maintaining astropy-helpers if there are
+extensive breaking changes in its dependencies.
+
+Impact for users
+----------------
+
+If done properly, these changes should have no noticeable impact for users.
```
eteq: As discussed in the coordination meeting, the one thing it will do is make it so that releases no longer contain c files.  I'm OK with that, but it should be noted here so we're sure we all understand.

APE17.rst
```diff
@@ -0,0 +1,701 @@
...
+----------------------
+
+The changes described here are opt-in, so barring any breaking changes in
+Python, setuptools, or Sphinx, everything should continue to work as expected if
+developers do not make any changes.
+
+Alternatives
+------------
+
+In the words of Stuart: *Do nothing, suffer submodules and maintaining the
```
eteq: I think it's worth stating that `pip` will probably eventually break the helpers (if that's true?).  That's a stronger "status quo won't work"/there are no alternatives argument.

APE17.rst
```diff
@@ -0,0 +1,701 @@
...
+packages.
+
+Impact for contributors
+-----------------------
+
+Users who wish to contribute fixes to the core astropy package or other packages
+will be encouraged to have `tox`_ installed if they want to easily run
+tests or build documentation locally. However, this is an easy package to
+install with `pip`_ and we could also add code in ``setup.py`` so that running
+``python setup.py test`` or ``python setup.py build_docs`` gives a helpful
```
eteq: Is it worth pointing out that we *can* potentially keep these commands, if they translate into `tox` or `pytest` commands?

APE17.rst
```diff
@@ -0,0 +1,701 @@
...
+migration guide should be advertised widely to all maintainers of packages
+using astropy-helpers. Throughout this process, astropy-helpers should continue
+to be supported, and once the migration guide is advertised the final date for
+support of astropy-helpers should be made clear. The recommendation in this
+APE is to support astropy-helpers until the end of the 4.0 LTS cycle at the
+latest (end of 2021), but this can be re-assessed depending on the ease of
+migration and the feasibility of maintaining astropy-helpers if there are
+extensive breaking changes in its dependencies.
+
+Impact for users
```
eteq:
```suggestion
Impact for users who are not developers
```

because on some level the "non-core developers" are really just users - this makes it less ambiguous

APE17.rst
Cadair: I don't think this is the right section for that. I wouldn't consider someone who cares if the tarball contains pre-generated or auto-generated C files our regular user. I am also pretty sure we mention it further down the document.

APE17.rst
astrofrog: Our draft packaging guide shows both how to define extensions natively for simple cases and with extension-helpers:

https://oa-packaging-guide-preview.readthedocs.io/en/latest/extensions.html

For astropy, it's definitely the case that we have enough extensions that using extension-helpers would be a lot simpler than defining everything in the setup.py file, and the auto-detection of cython files is also a nice feature.

@pllim - are you satisfied with this or do you have specific wording recommendations for the APE?

APE17.rst
Cadair: This is under the "Version Helpers" section, and is describing things which reference the old version stuff that would need adjusting for the new version stuff.

APE17.rst
Cadair: I think for packages that have one or two C extensions might not need to use `extension_helpers` as the standard ways of registering extensions would be sufficient. Certainly for core I think `extensions_helpers` would simplify things enough to warrant it's existence.

APE17.rst
Cadair: I added a note to this effect in the "alternatives" section.

APE17.rst
astrofrog: Done

APE17.rst
astrofrog: Done!

APE17.rst
pllim: Okay. I think I understand now. Thanks!

APE17.rst
astrofrog: Indeed there's actually no reason that extension-helpers couldn't be extended to help with extensions beyond C (e.g. C++, rust, etc.) as long as we keep it general.

APE17.rst
astrofrog: Fixed in https://github.com/astropy/astropy-APEs/pull/52/commits/7695c58e8e539acf8dc1f94e741448a1045c4d7e

APE17.rst
Cadair: This is elaborated upon in [this](https://github.com/astropy/astropy-APEs/pull/52/files#diff-7e96e0f54f0335dc6dfa73709a7c40d7R455) section.

eteq:
Thanks for addressing all my last bits there, @Cadair.  And thanks to @astrofrog and @Cadair for all their hard work on this.  

Or as @cadair would say: *burn it all*

eteq:
This APE has been accepted following unanimous approval at the 2019 coordination meeting.

PR #60, opened 2020-07-22
Update APE1 to include procedures for updating APEs

eblur:
I have updated a section of APE1 to include procedures for updating APEs via pull request.

The procedure is much like the creation of APEs, in that one person is expected the "champion" the modification PR with feedback from astropy-dev and similar communication venues. This policy will be useful as we go forward with more formal governance structures, e.g., those outlined in the forthcoming APE-0.

PR #63, opened 2020-10-16
APE 18: Adopt NEP 29

### LINKED ISSUE #62: APE 10: What is Python support beyond 4.0?
For instance, when are we dropping Python 3.6? Is astropy 5.0 a reasonable release to do so? Where will such decision be documented? Asking for astropy/astropy#10662 .

For that matter, where do we document plans to bump minversions for all the required dependencies for `astropy`? See https://github.com/astropy/astropy/blob/fecb178391f94eb69eecd67b82d9e00323072719/setup.cfg#L31-L34

Dropping Python 3.6 allows:

* Better annotation support.
* Usage of module level `__getattr__`.

Cadair:
The future is now.

Fixes #62 

[Rendered Version](https://github.com/Cadair/astropy-APEs/blob/NEP29/APE18.rst)

---

The following table shows the next few releases of the astropy core package and the  versions of CPython and Numpy it will support under this proposal:

 Astropy Release    | Release Date       | CPython Version | Numpy Version      |
--------------------|--------------------|-----------------|--------------------
 4.2                | November 2020      | 3.7+            | 1.17+ (Historical) |
 4.3                | April 2021         | 3.7+            | 1.17+              |
 5.0 (LTS)          | November 2021      | 3.7+            | 1.18+              |
 5.1                | April 2022         | 3.8+            | 1.19+              |
 5.2                | November 2022      | 3.8+            | 1.20+              |
 5.3                | April 2023         | 3.9+            | 1.20+              |


pllim:
Does this fix #62 ?

pllim:
Does this mean we also have to retire [APE 10](https://github.com/astropy/astropy-APEs/blob/master/APE10.rst) when this is accepted?

Cadair:
> Does this fix #62 ?

yes



> Does this mean we also have to retire [APE 10](https://github.com/astropy/astropy-APEs/blob/master/APE10.rst) when this is accepted?

Pass

pllim:
What does "pass" mean?

5.0 and 6.0 are LTS, right? If so, it should be stated as such in your APE.

Cadair:
> What does "pass" mean?

 I state that this APE has precedence over 10 but I don't know what the rules are on retiring APEs.


> 5.0 and 6.0 are LTS, right? If so, it should be stated as such in your APE.
 
Added

pllim:
> I don't know what the rules are on retiring APEs.

Hopefully @astropy/coordinators and/or @perrygreenfield know...

pllim:
Wait a minute... why is APE 9 in the diff?

Cadair:
whoops, think that was on my master branch

APE18.rst
```diff
@@ -0,0 +1,69 @@
...
+
+type: Standard Track
+
+status: Discussion
+
+
+Abstract
+--------
+
+After the release of astropy 4.0 APE 10 no longer specifies the supported versions of CPython.
```
astrofrog, line 18: Technically APE 10 does provide a recipe for future versions. Maybe just say that APE 10 previously determined the versions of CPython we supported and provided a specific list up to 4.0 LTS or something like that?

APE18.rst
```diff
@@ -0,0 +1,69 @@
...
+Since APE 10 was written the wider scientific Python community have written and largely adopted `NEP 29 <https://numpy.org/neps/nep-0029-deprecation_policy.html>`__ as a policy for supported versions of both CPython and Numpy.
+This APE proposes that Astropy follow the projects who have already adopted this policy.
```
astrofrog, line 20: Adopt this policy following other projects in the scientific python ecosystem or something like that? (That is we aren't pinning our policy to other packages)

APE18.rst
```diff
@@ -0,0 +1,69 @@
...
+
+
+Detailed description
+--------------------
+
+This APE fully adopts NEP 29 for both CPython and Numpy versions, it supersedes APE 10 wherever there is a conflict between that APE and this APE or NEP 29.
+
+While this APE will not replicate the whole of NEP 29, and NEP 29 should be considered the canonical source for the support policy, the core scheme to be adopted is as follows:
+
+astropy will support:
```
astrofrog, line 30: The core astropy package as well as other coordinated and infrastructure packages will support

astrofrog: This is perhaps something different from APE 10, that this would be project wide

APE18.rst
```diff
@@ -0,0 +1,69 @@
...
+
+* All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
+* All minor versions of numpy released in the 24 months prior to the project, and at minimum the last three minor versions.
+
+
+This means that for the next few releases of astropy the following versions of CPython and Numpy will be supported:
```
astrofrog, line 36: Of astropy -> the core astropy package

APE18.rst
```diff
@@ -0,0 +1,69 @@
...
+
++--------------------+--------------------+-----------------+--------------------+
+| Astropy Release    | Release Date       | CPython Version | Numpy Version      |
++--------------------+--------------------+-----------------+--------------------+
+| 4.2                | November 2020      | 3.7+            | 1.16+              |
++--------------------+--------------------+-----------------+--------------------+
+| 5.0 (LTS)          | April 2021         | 3.7+            | 1.17+              |
```
astrofrog, line 43: That doesn't seem right, 5.0 must be the following release? (4.0 was released in winter)

saimn:
:+1: to this, just two comments:

- We already dropped Numpy 1.16 : https://github.com/astropy/astropy/pull/10664

- I think Python adopted the PEP for yearly releases, and I'm not sure how it affects NEP 29.

APE18.rst
```diff
@@ -0,0 +1,69 @@
...
+| 6.0 (LTS)          | November 2022      | 3.8+            | 1.20+              |
++--------------------+--------------------+-----------------+--------------------+
+| 6.1                | April 2023         | 3.9+            | 1.20+              |
++--------------------+--------------------+-----------------+--------------------+
+
+
+Backward compatibility
+----------------------
+
+This version scheme is more rapid than that proposed in APE 10, however, this allows the core library to adopt modern features, and reduces burden on maintainers.
```
astrofrog, line 58: Also consistency with other packages

APE18.rst
```diff
@@ -0,0 +1,69 @@
...
+See the text of NEP 29 for further justification.
+
+Alternatives
+------------
+
+See NEP 29.
+
```
astrofrog, line 65: Link to NeP 29 whenever mentioned

astrofrog (changes requested):
Just some small comments!

APE18.rst
bsipocz: I would not force infra and coordinated to follow this, e.g. they currently usually support more, especially the infra packages.

bsipocz: So I would just go that they can't drop more than what is suggested in NEP29, but they could still support older versions of they want to

pllim: Yeah, not sure how we can enforce this project-wide in a realistic manner. We barely have enough people to maintain the core as it is.

APE18.rst
pllim: https://github.com/astropy/astropy/wiki/Release-Calendar doesn't say. You sure we won't have 4.3?

APE18.rst
Cadair: I have scoped this to core, not sure we really need to do anything else here?

bsipocz: @pllim - and my thinking is: not sure why we _want_ to enforce this project wide.

bsipocz: I think it's important that none of the infra/coordinated drop things sooner and switch to python 3.10+ now, but if they have a reason to still support 3.4, then why not (usually it only means that they didn't run into an issue and everything still works, while dropping is more work)

Cadair:
>   I think Python adopted the PEP for yearly releases, and I'm not sure how it affects NEP 29.

I think NEP 29 saw this coming hence they decided on a time based scheme, rather than one tied to the CPython release schedule. So I think the answer is it doesn't.

From the alternatives section of NEP 29:

> Given the current release cadence of the Python, the proposed time (42 months) is roughly equivalent to “the last two” Python minor versions. However, if Python changes their release cadence substantially, any rule based solely on the number of minor releases may need to be changed to remain sensible.

APE18.rst
eteq: Maybe we should say the "rule" applies to core, but coordinated&infrastructure packages will "aim" to follow this unless there's a reason not to?  I.e. just say for everything else it's a guideline.

I think the virtue of that is that it gives some guidance for packages that are sort of like "we don't have a rule because we haven't really thought that deeply about it".  Or alternatively/additionally, it provides an excuse to ask institutional players to help us bring things "up to standard" (I'm thinking here about how the py 3.x transition provided some people at institutions a way to fund additional work by existing Astropy devs)

bsipocz: many of the infra packages didn't really have a reason to drop old versions for a while, and I fail to see why it would be different in the future. As I said above, in many cases dropping a version is more work than keeping it, thus the "aim" is problematic in practice. And frankly, what is important for the ecosystem is that all coordinated and infra support at least the NEP29, that there is no situation when something gets dropped sooner than this schedule.

APE18.rst
```diff
@@ -0,0 +1,53 @@
...
+type: Standard Track
+
+status: Discussion
+
+
+Abstract
+--------
+
+APE 10 previously determined supported CPython versions, and provided a list up to the 4.0 release.
+Since APE 10 was written the wider scientific Python community have written and largely adopted `NEP 29`_ as a policy for supported versions of both CPython and Numpy.
```
eteq, line 19: While *I'm* fine with this, I think it might be good to give at least one concrete example of a way the more "accelerated" schedule will help things. Like what py3.7 features would we want to implement sooner rather than later? (manylinux2010 might be the answer! Or data classes?)

APE18.rst
```diff
@@ -0,0 +1,53 @@
...
+* All minor versions of numpy released in the 24 months prior to the project, and at minimum the last three minor versions.
+
+
+Backward compatibility
+----------------------
+
+This version scheme is more rapid than that proposed in APE 10, however, this allows the core library to adopt modern features, and reduces burden on maintainers.
+In addition to this aligning ourselves with other projects in the ecosystem is consistent and prevents potential minimum version conflicts.
+See the text of `NEP 29`_ for further justification.
+
```
eteq, line 42:
```suggestion

This APE fully replaces APE10 where they are in conflict.

```

APE18.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Detailed description
+--------------------
+
+This APE fully adopts `NEP 29`_ for both CPython and Numpy versions, it supersedes APE 10 wherever there is a conflict between that APE and this APE or `NEP 29`_.
+
+While this APE will not replicate the whole of `NEP 29`_, and `NEP 29`_ should be considered the canonical source for the support policy, the core scheme to be adopted is as follows:
+
+The core astropy package will support:
```
eteq, line 30: One thing I'm confused by here: is this at the time of release or at the time of latest *major version* release?  That is, if we do a 4.0.x release 18 months after 4.0 is released, are we bound by the Python version from 42 months before 4.0.x or 60 months?

eteq (approved):
I'm happy with the spirit of this, but have a few clarifying questions/suggestions.  (Also this is definitely the record for time-to-APE since the discussion prompting this started earlier today :star: !)

APE18.rst
bsipocz: There are some graphics in NEP29, that makes this clear.

bsipocz: (except the ascii rendering is messed up, the text still makes it clear)

APE18.rst
astrofrog: Agree we can make it a rule for core and recommendation for other parts of the project?

APE18.rst
Cadair: It's based on the *scheduled* release date of non-bugfix versions, so a LTS version will support the version at the scheduled release for it's lifetime. So it would be 60 months in your example.

APE18.rst
Cadair: I added a small justification in the backwards compatibility section, I don't really want to do a "let's adopt 3.7 because new features things" because I would rather this stand the test of time. I am happy to pull the justification section up here if you think it's a better place for it?

APE18.rst
```diff
@@ -0,0 +1,55 @@
...
+
+
+Backward compatibility
+----------------------
+
+This version scheme is more rapid than that proposed in APE 10, however, this allows the core library to adopt modern features, and reduces burden on maintainers.
+In addition to this aligning ourselves with other projects in the ecosystem is consistent and prevents potential minimum version conflicts.
+See the text of `NEP 29`_ for further justification.
+
+This APE fully replaces APE10 where they are in conflict.
```
bsipocz, line 43: this should also go as an addendum for ape10

Cadair: sounds like a good job for the CoCo after merge ;)

eteq: Hmm.  I guess I'd say that *should* be part of this PR in practice, but I guess we sort of need #60 to do that :shrug:. But it would be good to make sure there's a note at the bottom that whoever merges this knows they have to do it (since it's not part of the normal merge process).

APE18.rst
eteq: Gotcha, that definitely makes sense, but I was suggesting that we state this explicitly.  I'll make a concrete suggest for that shortly in a follow-up comment.

APE18.rst
```diff
@@ -0,0 +1,55 @@
...
+
+This APE fully adopts `NEP 29`_ for both CPython and Numpy versions, it supersedes APE 10 wherever there is a conflict between that APE and this APE or `NEP 29`_.
+
+While this APE will not replicate the whole of `NEP 29`_, and `NEP 29`_ should be considered the canonical source for the support policy, the core scheme to be adopted is as follows:
+
+The core astropy package will support:
+
+* All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
+* All minor versions of numpy released in the 24 months prior to the project, and at minimum the last three minor versions.
+
```
eteq, line 34:
```suggestion

Note that, as stated in `NEP 29`_, these timelines all apply for the versions
set at the time of the non-bugfix release.  This means that the full
support timelines for LTS versions of astropy are 66 and 48 months for Python
and numpy, respectively.

```

astrofrog: Maybe worth saying since these are currently supported for two years (in case we change that some day)

bsipocz: I find this phrasing very confusing. 

What about rewording to say the 42 and 24 are for the start of a release "family", and support won't be dropped in a bugfix release, neither on LTS nor on normal feature release bugfix?

bsipocz: (also, I promised Stu a blurb about the exceptation, e.g. the latest LTS dropped a numpy version the release, as it would have been dropped  based on NEP29 very soon after the release)

Cadair: > What about rewording to say the 42 and 24 are for the start of a release "family", and support won't be dropped in a bugfix release, neither on LTS nor on normal feature release bugfix?

:+1: This is in NEP 29, but I am happy to call it out here in terms of the LTS if people feel it needs more clarity.

bsipocz: yes, I'm OK just the reference back to NEP29, it's very clear in there, but the above blurb about the 66 and 48 is very very confusing to say the least. A start of a release is a very clear date, but the end of it? And why would it even matter, by definition things are not going to be dropped between 4.0.10 and 4.0.11.

bsipocz:
There has been also some starting momentum for setting maximum versions of dependencies, but it hasn't yet formalized in a formal document. If they will be any, I'm happy to put in a separate APE for it.

APE18.rst
taldcroft: >  if they have a reason to still support 3.4, then why not 

Just thinking about #10888 that I just put in, which would require 3.7+ for astropy to work properly. If a coordinated package tries to force using Python 3.4 + astropy 4.2 then there will be problems. Not sure if that is credible.

astrofrog:
@Cadair - maybe you can add #10900 under implementation?

APE18.rst
```diff
@@ -0,0 +1,59 @@
...
+APE 10 previously determined supported CPython versions, and provided a list up to the 4.0 release.
+Since APE 10 was written the wider scientific Python community have written and largely adopted `NEP 29`_ as a policy for supported versions of both CPython and Numpy.
+This APE proposes that Astropy adopt this policy following other projects in the scientific ecosystem.
+
+
+Detailed description
+--------------------
+
+This APE fully adopts `NEP 29`_ for both CPython and Numpy versions, it supersedes APE 10 wherever there is a conflict between that APE and this APE or `NEP 29`_.
```
Cadair, line 26: I question raised by @bsipocz is should we allow, or even mandate, that an LTS release can jump forward a numpy (or even CPython?!) version to reduce the long term maintenance burden.

I think a good way to do this would be to say something like "An LTS release of the core package may set the minimum versions for the next release" (i.e. 5.0 could adopt the versions specified by 5.1).

taldcroft: I would just say that institutions that are using the LTS are more likely going to have a stack that has generally older versions of Python and dependencies. So the target audience for LTS may not be happy about this.

E.g. Chandra operations is currently using Python 3.6 + numpy 1.12. We're working on an upgrade now, but it turns out to be a lot of work in our setting (review board for any software changes, etc).

astrofrog: i would just stick to the NEP 29 rules - if we had done for 4.0 I believe we would have required Numpy 1.15? Not too different from 1.16.

Cadair: The fact that it was bumped for 4.0 is the reason for this thread. That decision was clearly motivated by something. @mhvk ?

mhvk: Yes, we bumped to numpy 1.16 since then we could support quantities on nearly all numpy functions, i.e., there was what was considered a very compelling reason (and LTS played a role, as with the bump the user experience would not depend on numpy version; IIRC, it also was the case that in our old scheme it would have been OK just a few months later). 

Of course, generally numpy versions do not introduce major changes that are as directly relevant to astropy, so the general rule should definitely be NEP 29. But obviously there can be some room for exceptions - I guess it would be an argued addendum to this APE perhaps, i.e., a high bar.

APE18.rst
mhvk: Agreed that this adds more confusion than it solves. I'd leave the text as is.

mhvk (approved):
Looks all very good to me. I would put it in as is.

APE18.rst
bsipocz: I would not mandate but allow. As I recall for 4.0, the timeline was just a few months early,   so it clearly made a sense both for the Quantity support, but also for planning ahead, to not to deal with 1.16 issues 2 years down the line

astrofrog: Yes having a clause about allowing it is very specific cases makes sense - and doesn't have to apply to just LTS eg if 4.3 came out a few days before the next bump.

Cadair: I have added some words. If people want to review it.

APE18.rst
Cadair: Done

APE18.rst
bsipocz: I'm not sure what you mean credible here, it's always up to the package manager to resolve the version requirements, I don't see any force being involved here.

E.g. on 3.4, it could be possible to install the latest version of a coordinated package, and naturally, it would come with an older astropy version. It's down to the CI of the coordinated package to make sure they are compatible with their minimum supported version combination, after all the user should know best why they are stuck on an older python/numpy combo.

(3.4 is probably a bad example, but a real life example, with astroquery we had to stay python2.7 compatible much much longer than astropy core as it took long for casa to move off it. We, the astroquery maintainers decided that it's something we were willing to do, and it was totally independent of the version support policy of core that was decided separately. )

APE18.rst
```diff
@@ -0,0 +1,65 @@
...
+
+While this APE will not replicate the whole of `NEP 29`_, and `NEP 29`_ should be considered the canonical source for the support policy, the core scheme to be adopted is as follows:
+
+The core astropy package will support:
+
+* All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
+* All minor versions of numpy released in the 24 months prior to the project, and at minimum the last three minor versions.
+
+This version support policy allows, at the discretion of the maintainers, to adopt the supported version of the next release in situations where it is judged to be valuable.
+For example, in the case of an important change in numpy or a change over date very close to the release.
```
astrofrog, line 36:
```suggestion
For example, this could be done in the case of an important change in numpy or a change over date very close to the release.
```

APE18.rst
```diff
@@ -0,0 +1,65 @@
...
+
+
+Implementation
+--------------
+
+Some initial work has been done in `#10900 <https://github.com/astropy/astropy/pull/10900>`__ on bumping core to Python 3.7+.
```
astrofrog, line 42:
```suggestion
Some initial work has been done in `#10900 <https://github.com/astropy/astropy/pull/10900>`__ on bumping the minimum Python version in the core astropy package to Python 3.7+.
```

APE18.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Backward compatibility
+----------------------
+
+This version scheme is more rapid than that proposed in APE 10, however, this allows the core library to adopt modern features, and reduces burden on maintainers.
```
astrofrog, line 47:
```suggestion
The version schedule here is more rapid than that proposed in APE 10, however, this allows the core library to adopt modern features, and reduces burden on maintainers.
```

APE18.rst
```diff
@@ -0,0 +1,65 @@
...
+In addition to this aligning ourselves with other projects in the ecosystem is consistent and prevents potential minimum version conflicts.
```
astrofrog, line 48:
```suggestion
In addition to this, aligning ourselves with other projects in the ecosystem is consistent and prevents potential minimum version conflicts.
```

APE18.rst
```diff
@@ -0,0 +1,65 @@
...
+
+APE 10 previously determined supported CPython versions, and provided a list up to the 4.0 release.
+Since APE 10 was written the wider scientific Python community have written and largely adopted `NEP 29`_ as a policy for supported versions of both CPython and Numpy.
+This APE proposes that Astropy adopt this policy following other projects in the scientific ecosystem.
+
+
+Detailed description
+--------------------
+
+This APE fully adopts `NEP 29`_ for both CPython and Numpy versions, it supersedes APE 10 wherever there is a conflict between that APE and this APE or `NEP 29`_.
```
astrofrog, line 26:
```suggestion
This APE fully adopts `NEP 29`_ for both CPython and Numpy versions for the core astropy package, and it supersedes APE 10 wherever there is a conflict between that APE and this APE or `NEP 29`_.
```

astrofrog (changes requested):
Some very minor wording improvements (hopefully) below

APE18.rst
```diff
@@ -0,0 +1,65 @@
...
+
+This APE fully adopts `NEP 29`_ for both CPython and Numpy versions for the core astropy package, and it supersedes APE 10 wherever there is a conflict between that APE and this APE or `NEP 29`_.
+
+While this APE will not replicate the whole of `NEP 29`_, and `NEP 29`_ should be considered the canonical source for the support policy, the core scheme to be adopted is as follows:
+
+The core astropy package will support:
+
+* All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
+* All minor versions of numpy released in the 24 months prior to the project, and at minimum the last three minor versions.
+
```
bsipocz, line 34:
```suggestion
* All minor versions of Python released 42 months prior to a non-bugfix, and at minimum the two latest minor versions.
* All minor versions of numpy released in the 24 months prior to a non-bugfix, and at minimum the last three minor versions.
* Versions of other runtime dependencies released 24 prior to a non-bugfix release.

These versions of the dependencies are kept supported for bugfix lifetime of the given astropy version.

```

bsipocz: So, this would introduce a policy for scipy/matplotlib/etc. We should not drop them to support only latest greatest, but then should linger with an old version neither. (inspired by @pllim seeking for input whether to drop 0.18 scipy)

astrofrog: I'd be up for adding such a clause, though we should make it clear this is an addition compared to NEP 29?

bsipocz: Yes, I'm happy to add a blurb about that if this is a go ahead.

mhvk: Scipy sticks to NEP 29 itself, so we could state that in general we try to main synchronism, i.e., pick that scipy that supports all of the numpy/python we support, but not more.

bsipocz: I like to say we guarantee to support x versions, rather that we stay in sync, e.g. if there is no reason popping up to drop I would not drop a version (e.g., we could support very old scipies for a long time), but I would not put any extra effort in keeping up supporting an old versions when it's outside of the range.

mhvk: @bsipocz - agreed there is no reason to explicitly fail on older versions (as we do for numpy); at some level it is more about what version we still test -- and for which we thus *guarantee* support.

bsipocz: The oldest supported one we should test with, anything older would not work as those versions are set in package metadata. I would only say we don't bump the versions unless any issues came up with them, e.g. now many versions turn out to be not python3.7 compatible, so we need to bump them along with dropping support for 3.6.

APE18.rst
eteq: I see your point on "stand the test of time", but I was thinking the point of this is to give a couple of concrete examples to what is currently a very abstract phrasing.  So just meant something like "For example, at the time of the py3.7 release infrastructure maitainers found their life would be easier if manylinux2010 was an option". Or similar.

But if you really want it to be timeless, I'm ok with either pulling it up here or leaving as-is :shrug:.

APE18.rst
eteq: I disagree that it's very clear in NEP 29 for the *casual* reader.  That is it's very specific and when I read very carefully I understand the above, and really anyone in this conversation is clearly interested enough to dig deeply on this.  But I do *not* think the typical reader who comes upon this via google and just wants to know "how long will an astropy support this Python version" will take the time to click through to NEP 29, map it onto our Astropy-specific release candence, etc.

That said, if y'all think this wording is not clearer, maybe the better thing is to just give a few concrete examples and let the reader extrapolate?  Something like:
```suggestion
Note that, as stated in `NEP 29`_, these timelines all apply for the versions
set at the time of the non-bugfix release.  So for example the v4.0 release, an LTS, would support Python 3.6 (released 23 mo earlier), and would continue to support it until the end of the LTS window (24 mo after the 4.0 release).  Whereas 4.2, planned for late 2020, would  unambiguously only support 3.7 (as its full lifetime falls within the 3.7 window).
```
Or @cadirs wording from https://github.com/astropy/astropy-APEs/pull/63/files#r506678119 makes sense to me too.

bsipocz: See my extra sentence about the bugfix lifetimes.

bsipocz: (I don't find these concrete examples any useful, try to swap it out with e.g 1.3 and it will all read obsolete for the casual reader who you say would not bother to read/try to understand the tables and figures in NEP29)

APE18.rst
Cadair: I am :+1: on this suggestion

Cadair:
I have adopted @bsipocz's suggestion regarding all the other versions, and edited the text to make it clear we are extending the numpy policy in NEP 29 by doing that. I think given we merged astropy/astropy#10900 there is unlikely to be strong disagreement to this proposal, we should probably move to getting it merged?

APE18.rst
Cadair: There is a line now stating that the versions are supported for the lifetime of the bugfixes, so hopefully that resolves the core of your comment @eteq in a way people are happy with?

astrofrog (approved):

APE18.rst
```diff
@@ -0,0 +1,70 @@
...
+This APE will not replicate the content of `NEP 29`_, and the full text of `NEP 29`_ is adopted via this APE.
+In addition to the content of `NEP 29`_, the APE also extends the numpy policy to all of the core package dependencies.
+This leads to the following support policy for dependencies of the core package.
+
+Astropy core will support:
+
+* All minor versions of Python released 42 months prior to a non-bugfix, and at minimum the two latest minor versions.
+* All minor versions of numpy released in the 24 months prior to a non-bugfix, and at minimum the last three minor versions.
+* Versions of other runtime dependencies released 24 prior to a non-bugfix release.
```
mhvk, line 36: Insert months after 24

APE18.rst
mhvk: Reads well to me.

mhvk (approved):
Re-approving; just one omitted word.

APE18.rst
Cadair:
```suggestion
* Versions of other runtime dependencies released 24 months prior to a non-bugfix release.
```

mhvk:
I cannot merge this, but maybe someone who can can do the honour? Thanks, @Cadair!

APE18.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Alternatives
+------------
+
+See `NEP 29`_.
+
+Decision rationale
+------------------
+
+<To be filled in by the coordinating committee when the APE is accepted or rejected>
```
mhvk, line 66: I guess this line does need to be changed.

Cadair:
@astropy/coordinators Can this get wrapped up?

APE18.rst
```diff
@@ -0,0 +1,70 @@
...
+
+This version support policy allows, at the discretion of the maintainers, to adopt the supported version of the next release in situations where it is judged to be valuable.
+For example, this could be done in the case of an important change in numpy or a change over date very close to the release.
+
+
+Implementation
+--------------
+
+Some initial work has been done in `#10900 <https://github.com/astropy/astropy/pull/10900>`__ on bumping the minimum Python version in the core astropy package to Python 3.7+.
+
```
astrofrog, line 48: The CoCo discussed this APE just now and we agree it is good to go but wanted to make a small suggestion to mention the release calendar as follows:

```suggestion
Some initial work has been done in `#10900 <https://github.com/astropy/astropy/pull/10900>`__ on bumping the minimum Python version in the core astropy package to Python 3.7+.

The release calendar for the astropy core package will be updated to include the minimum Python and Numpy version for each future release.
```

We didn't want to include the link since it's just a wiki page which may move in future. Does this seem reasonable?

eteq:
With the change above, the CoCo was happy to approve this APE, so I am merging now.

Issue #66, opened 2021-03-09
MNT: Rename default branch from master to main


pllim:
Please rename your default branch from `master` to `main` as part of astropy/astropy-project#151 , preferably by 2021-03-22. Also a friendly reminder to check documentation, workflows, etc., and update them accordingly. Please don't forget to communicate this change to your users and stakeholders. To summarize:

- [x] Rename branch from `master` to `main`, preferably by 2021-03-22.
- [x] Update documentation, workflows, etc., accordingly. -- See #67
- [x] Communicate this change to your users and stakeholders.

Once this is taken care of, you may close this issue.

*This is an automated issue. If this is opened in error, please let @pllim know!*

pllim:
A lot of stuff across the project refer to this repo, so this is one of the repos in the project that needs to change first.

eteq:
Change done right now @pllim!

eteq:
One thing to note about this that's especially important for this repo - the github magic recognizes renames such that https://github.com/astropy/astropy-APEs/blob/master/APE8.rst now redirects to https://github.com/astropy/astropy-APEs/blob/main/APE8.rst , so all the links don't *have* to be updated (although they should be where possible!)

pllim:
Please also review #67

eteq:
Oops! Sorry @pllim, I missed #67 and did some of that in `main` already.  But will follow-up in #67

PR #67, opened 2021-03-09
MNT: Replace master with main

pllim:
This PR attempts to replace all mention of `master` to `main` in this repository. This should be reviewed and merged right after the maintainer has renamed the default branch.

This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

This is part of #66

### DIFF
--- a/APE0.rst
+++ b/APE0.rst
@@ -362,7 +362,7 @@ Infrastructure
 (`YTEP 1776 <https://ytep.readthedocs.io/en/latest/YTEPs/YTEP-1776.html>`_), and
 earlier less-formal descriptions of the Astropy governance.
 
-.. _APE 1: https://github.com/astropy/astropy-APEs/blob/master/APE1.rst
+.. _APE 1: https://github.com/astropy/astropy-APEs/blob/main/APE1.rst
 
 
 Branches and Pull Requests
--- a/APE10.rst
+++ b/APE10.rst
@@ -143,7 +143,7 @@ download Astropy v3.0 regardless of the active Python version.
 the ``Requires-Python`` metadata field for Python packages, which can be used
 to explicitly specify which Python versions a package is compatible with.
 Support for this was added to `setuptools v24.2
-<https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v2420>`_ via the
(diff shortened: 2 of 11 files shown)


eteq:
Note I think I made the changes in this PR in the README in 5f4b69e750cb196b6e96b6f91a2071cd264512dc - sorry, @pllim, I didn't see this PR until *after* I did that! 

Note this also includes updates to the APE content.  But I'm realizing now that you've done this @pllim, that I'm not sure if this falls under the just-done "update" clause.  Which technically means @pllim  becomes a revision author.  And regardless technically it requires CoCo approval to update the text of APEs (but not the README).  So I see three options really:

1. Follow the normal revision process which means @pllim has to do some more changes here to reflect the revision of the documents, update Zenodo entries, etc
2. Don't bother because these are sort of "fixed time" documents - as long as the links still work that's probably OK.
3. Just say this fine and accept it.
 

Maybe @astrofrog or @taldcroft have opinions here?

eteq (changes requested):
Note I think I made the changes in this PR in the README in 5f4b69e750cb196b6e96b6f91a2071cd264512dc - sorry, @pllim, I didn't see this PR until *after* I did that! 

Note though, that this PR also includes updates to the APE content.  But I'm realizing now that you've done this @pllim, that I'm not sure if this falls under the just-done "update" clause.  Which technically means @pllim  becomes a revision author.  And regardless technically it requires CoCo approval to update the text of APEs (but not the README).  So I see three options really:

1. Follow the normal revision process which means @pllim has to do some more changes here to reflect the revision of the documents, update Zenodo entries, etc
2. Don't bother because these are sort of "fixed time" documents - as long as the links still work that's probably OK.
3. Just say this fine and accept it.
 

Maybe @astrofrog or @taldcroft have opinions here?

bsipocz:
> I'm not sure if this falls under the just-done "update" clause.

wouldn't this be on the level of typos? We had those fixed before without doing the update cycle and a round of approval. Basically there is no situation where these updates should be rejected...

hamogu:
It seems that all the changes have been made (looking into the current `main` on the GH web interface). Is there a particular reason this PR is still open?

pllim:
@eteq requested changes, so he should re-review.

pllim:
Thanks!

PR #68, opened 2021-04-28
Update APE-6 for subtype support

taldcroft:
This updates APE-6 to add support for extended data types, in particular multidimensional fixed-dimension columns, variable-length arrays, and object columns. This is done primarily by adding a ``subtype`` data keyword and specifying the convention for using this keyword for new supported extended types.

The discussion in https://github.com/astropy/astropy/issues/11368 was a large driver for this update including details of the new specification.

Writing standards documents is not my forte, so comments are most welcome!

cc: @mbtaylor @mhvk @hamogu @dhomeier

### DIFF
--- a/APE6.rst
+++ b/APE6.rst
@@ -547,6 +547,7 @@ what we call a "mapping", for instance ``{"a":1, "b":2}``.
 The example below shows writing an object array to ECSV. Note that JSON requires
 a double-quote around strings, and ECSV requires ``""`` to represent
 a double-quote within a string, hence the double-double quotes.
+::
 
   # %ECSV 1.0
   # ---
--- a/APE6.rst
+++ b/APE6.rst
@@ -407,10 +407,9 @@ Each column specifier is a dictionary structure with the following keys:
   Some implementations may not support all types.
 
 ``subtype``: string, optional
-   The ``subtype`` keyword indicates the presence of an extended data type such
-   as a variable-length array or an object column. If ``subtype`` is defined the
-   ``datatype`` must be set to ``string``. See the section below on `Subtype
-   data`_ for details.
+   The ``subtype`` keyword describes an extended data type such as a
+   variable-length array or an object column. See the section below on `Subtype
+   data`_ for details. This is new in version 1.0 of the ECSV standard.
 
 ``unit``: string, optional
    Data unit (unit system could be part of schema?).
@@ -463,44 +462,40 @@ Subtype data
 """"""""""""
 
 From version 1.0 and later it is possible to embed extended data types beyond
-simple typed scalars in the data section. The data are encoded as a string, with
-details of the encoded data being captured in the ``subtype`` keyword which is
-new in version 1.0 of ECSV.
+simple typed scalars in the data section. The column data in the ECSV output
+shall be consistent with the specified ``datatype``, with additional details of
+the data being captured in the``subtype`` keyword.
 
-For extended data the ``datatype`` is set to ``string``. This informs the table
-reader that the literal values in the data section will be strings, ensuring
-back-compatibility with readers that do not implement the functionality to
-decode the values using the ``subtype`` key.
+If table readers do not recognise or support the ``subtype`` then they may
+ignore it and use the ``datatype`` only.
 
 The ECSV standard defines three types of extended data that can be represented:
 fixed-dimension array data, variable-length array data, and object data. These
 correspond to specified ``subtype`` values described below. It is also allowed
-to define new custom ``subtype`` values for specific applications. If table
-readers do not recognize the subtype then the column should be returned as a
-string.
+to define new custom ``subtype`` values for specific applications.
 
 Fixed-length array data
 @@@@@@@@@@@@@@@@@@@@@@@
 
 For columns where each data cell is an array with consistent dimensions, the
 ``datatype`` is set to ``string`` and the ``subtype`` is set to the actual data
-type (one of the allowed values of the ``datatype`` keyword specified
-previously) followed by the `JSON <https://www.json.org/>`_ representation of
-the shape (dimensions) of each cell.
+type (one of the allowed values of the ``datatype`` keyword) followed by the
+`JSON <https://www.json.org/>`_ representation of the shape (dimensions) of each
+cell.
 
 The contents of each cell are represented as a string using the JSON encoding of
 the array values. The encoding shall use row-major ordering with array shapes
 defined accordingly.
 
-In the example below each cell is a ``3 x 2`` array of ``float64`` type. The shape
-is ``[3,2]`` so the ``subtype`` is ``float64[3,2]``::
+In the example below each cell is a ``3 x 2`` array of ``float64`` type. The
+shape is ``[3,2]`` so the ``subtype`` is ``float64[3,2]``::
 
   # %ECSV 1.0
   # ---
   # datatype:
-  # - {name: a, datatype: string, subtype: 'float64[3,2]'}
+  # - {name: array3x2, datatype: string, subtype: 'float64[3,2]'}
   # schema: astropy-2.0
-  a
+  array3x2
   [[0.0,1.0],[2.0,3.0],[4.0,5.0]]
   [[6.0,7.0],[8.0,9.0],[10.0,11.0]]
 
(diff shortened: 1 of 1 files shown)


hamogu:
Looks good to me.

mbtaylor:
@taldcroft, this basically looks good, but I have couple of suggestions for change.

First, that use of `subtype` is not restricted to `string` columns.  I agree that most uses, including the array and object ones enumerated here, will be for string columns, but _(a)_ I don't see any cost to relaxing this restriction, and _(b)_ I can imagine possibilities where non-string subtypes might be desirable, e.g. `datatype: float64, subtype: MJD`; I don't know if python/numpy has specific datatypes for time values, but some languages or parsing environments might do.  Usages like this don't have to be written into the APE now or ever, but if either ECSV v1.x or individual users decide something like that would be a good idea in future, it won't conflict with the existing specification.

So where the `subtype` keyword is introduced I would remove or replace the sentence _"If ``subtype`` is defined the ``datatype`` must be set to ``string``"_, and in the _Subtype data_ section replace the sentence:

> If table readers do not recognize the subtype then the column should be returned as a string.

with something like:

> If table readers do not recognise or support the subtype then they may ignore it and use the `datatype` only.

A few other knock-on changes elsewhere in the text will be required for consistency.

Second, for the "object" subtype, I'd suggest renaming it to "json".  This is more descriptive, but also more accurate: in your example file in this section, one of the cells has the value `true`, which is JSON, but is not a JSON `object` (which would have to be enclosed in curly brackets).

taldcroft:
> Relaxing the restriction on requiring `string`

Sounds fine and I agree with allowing flexibility for future enhancements.

> Second, for the "object" subtype, I'd suggest renaming it to "json". 

OK. This requires a small change to the merged ECSV PR (https://github.com/astropy/astropy/pull/11662).

mhvk:
Now had a look. I like the addition very much. I also like the changes proposed by @mbtaylor - and am happy to quickly review the necessary follow-up to the astropy implementation.

taldcroft:
@mbtaylor @mhvk - I think have addressed the comments and that this is ready for final review. It's worth viewing the rendered file since I picked up some problems I had not previously noticed.

mhvk (approved):
Looks good to me!

mbtaylor (changes requested):
Mostly looks good, but at least one change required: the file example in the *Object columns* section still has `subtype: object` instead of `subtype: json`.  I wonder if it would also make sense to retitle that section "JSON columns" and change the language in the text to avoid the term "object" given that the things in the column can be scalars with JSON representation ... but that might end up being more confusing, so only do that if you think it's a good idea.

There are also a couple of typos inherited from earlier versions of the document:
* "reprented" -> "represented"
* "include column types" -> "including column types"
* "tablular" -> "tabular"

taldcroft:
@mbtaylor - my hope was that the "As a point of clarification" sentence would make things clear. I'm certainly coming from a Python or Java perspective where everything (scalars included) is an object. IMO the JSON spec choose poorly in labeling a dict/mapping as "object", so my preference is to stick with the existing text and more common usage of "object".

We can always come back to this choice of wording if it ends up causing confusion. It won't change any implementation details.

Thanks for the sharp eye I'll fix those. I did put each of the examples into the Python parser, but it turns out that since `object` is a valid numpy dtype having `"object"` there also works for Python.

mbtaylor:
Fine, I'm happy with the existing text then.  All OK by me.

APE6.rst
```diff
@@ -451,25 +458,106 @@ the following rules:
 - A double quote character in a field must be represented by two double quote
   characters.
 
+Subtype data
+""""""""""""
+
+From version 1.0 and later it is possible to embed extended data types beyond
+simple typed scalars in the data section. The column data in the ECSV output
+shall be consistent with the specified ``datatype``, with additional details of
+the data being captured in the``subtype`` keyword.
```
mbtaylor, line 467: "subtype" not formatted correctly here - missing space?

mbtaylor:
One other formatting error I spotted: in the second paragraph of the "Header Details" section, there is a long run of monospaced text after "start with the two characters"

mbtaylor:
@taldcroft yes, we did discuss this before at https://github.com/astropy/astropy/pull/11569#issuecomment-824199421, and I said I didn't have strong views about whether blank values should be permitted for array-valued columns.

However, on reflection I think blanks should be permitted here, as you say for consistency and also because I think people are going to want to be able to write data like that, for instance a Gaia column containing a flux time series where some rows have time series data and others don't.

Yes my java implementation handles this naturally.  How the data is stored ... it depends, but as far as the API goes, the content of each cell is a java Object, and that can be returned as either an array object or a null value (the data for the whole column is not available in the API as a free-standlng data structure).

If it turns out to be hard or impossible to represent those blank values faithfully within the Astropy implementation I'd say that's OK, just provide the best representation of a blank value that's feasible, maybe a zero-length array or an array of all nan's or whatever.  My representation of the ECSV data model is not going to be perfect either, e.g. I can't represent blank array elements of `subtype int*[...]` array-valued cells, so they'll just have to be zeros.

taldcroft:
@mbtaylor - It turned out to be just a few lines of code to support blank values for masking data (plus tests of course, https://github.com/astropy/astropy/pull/11720), so I'll update the APE-6 PR accordingly.  I did end up on a detour with https://github.com/numpy/numpy/issues/18981. Unfortunately in Python if you mask a variable length array with a blank, numpy does something funky to make it appear as an unmasked zero-length array instead of a masked value. We can live with that and it doesn't impact APE-6.

taldcroft:
BTW, I was not able to understand that issue with the long literal line in the Header Details section. I stared at the source for a long time and AFAIK it was valid RST. For some reason it wasn't rendering right, so I took an end run and changed the words.

APE6.rst
```diff
@@ -559,6 +582,9 @@ a double-quote within a string, hence the double-double quotes.
   "{""b"":[2.5,null]}"
   true
 
+In this subtype, the ``null`` marker is decoded by JSON as the language-specific
```
mbtaylor, line 585: This sentence sounds a bit confusing to me (maybe it's a language-specific
use of the term "JSON" to mean "the JSON parsing library").  I'd suggest
instead something like:

> The ``null`` token is the JSON representation of a null value.  JSON parsers will typically decode this to some language-specific value, for instance ``None`` in Python.

mbtaylor:
I only have two remaining tiny niggles:
* The _Subtype data_ section refers to "three types of extended data that can be represented: fixed-dimension array data, variable-length array data, and object data".  But the first two subsections are titled "fixed-length" and "variable-length" not "fixed-dimension" and "variable-length" array data.  Make this terminology consistent by rewording text or subtitles?
* The first sentence in the "Variable-length array data" section says "the data cell are arrays", should be "the data cells are arrays".

Otherwise, all looks good to go!

taldcroft:
@mbtaylor - looking at this fresh I didn't like the "fixed-dimension" description. I changed to "multidimensional", which I hope is clear enough and definitely more common usage.

mbtaylor:
I'm not sure that replacing "fixed-length" with "multidimensional" is an improvement, since "variable-length array" subtypes can be multidimensional too (and "multidimensional array" subtypes can be 1-dimensional).

But the terminology is consistent, and the explanatory text and examples are clear, so if you prefer it this way I'm happy to accept it.

taldcroft:
@mbtaylor - thanks. In that case let's stick with what is there. Since @mhvk has approved and nobody else has commented since I posted on astropy-dev, I think we can go to the next step and I will request final approval from the rest of the Coordination Committee.

Thanks for the great input which has been immensely helpful. I'm pretty happy with where this has ended up now as a result.

astrofrog (approved):

eteq (approved):

eteq:
The Coordination Committee discussed this today and have approved this APE.  In the last commit I updated the info accordingly, and will now merge

mbtaylor:
Excellent.  Thanks a lot @taldcroft et al. for getting this done, I agree it's worked out well.
One question: ECSV 0.9 had a DOI, will ECSV 1.0 get one, and where should I go looking for it?

taldcroft:
@mbtaylor - the updated DOI will be added as part of the standard APE acceptance process, hopefully soon.

PR #71, opened 2021-12-06
Clarifications to APE 0

Cadair:
Having run the first set of elections for Voting Members and for the Coordinating Committee the Governance Implementation Working Group noticed some rough edges and places where the meaning of APE 0 wasn't clear. This pull request makes a few (hopefully) small changes to the text to help those running future elections.

This PR is co-authored by @mwcraig and @ceb8 with input from @eteq.

### DIFF
--- a/APE0.rst
+++ b/APE0.rst
@@ -122,8 +122,9 @@ Electing the Coordination Committee
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 A Coordination Committee election is held annually to fill any vacant seats.
 Notice of the election must be posted via the Project’s standard communication
-channels (reaching all active members) at least one month before the election
-begins. The election consists of two phases, each lasting two weeks:
+channels for users, contributors and voting members, at least one month before
+the voting phase begins. The election consists of two phases, each lasting at
+least two weeks:
 
 * Phase 1, Nominations: Candidates will be nominated and confirm their interest
   in serving. Candidates can be nominated by anyone, including themselves, and
--- a/APE0.rst
+++ b/APE0.rst
@@ -134,9 +134,10 @@ least two weeks:
   nominations period.
 * Phase 2, Voting: Each `Voting Member <votingmembers>`_ can vote for zero or
   more of the candidates, up to the number of candidates. Voting is performed
-  anonymously. Candidates are ranked by the total number of votes they receive.
-  If a tie occurs, it may be resolved by mutual agreement among the candidates,
-  or else the winner will be chosen at random.
+  anonymously. Candidates who receive votes from a majority of the voting
+  members participating in the election are ranked by the total number of votes
+  they receive. If a tie occurs, it may be resolved by mutual agreement among
+  the candidates, or else the winner will be chosen at random.
 
 The election process is managed by a non-voting Returns Officer nominated by
 the outgoing Coordination Committee, although the process (specific choice of
--- a/APE0.rst
+++ b/APE0.rst
@@ -129,7 +129,9 @@ least two weeks:
 * Phase 1, Nominations: Candidates will be nominated and confirm their interest
   in serving. Candidates can be nominated by anyone, including themselves, and
   do not have to be `Voting Members <votingmembers>`_ to be nominated.
-  Candidates should provide a brief statement in support of their candidacy.
+  Candidates should provide a brief statement in support of their candidacy,
+  these statements should be made public as they are received during the
+  nominations period.
 * Phase 2, Voting: Each `Voting Member <votingmembers>`_ can vote for zero or
   more of the candidates, up to the number of candidates. Voting is performed
   anonymously. Candidates are ranked by the total number of votes they receive.
--- a/APE0.rst
+++ b/APE0.rst
@@ -308,7 +308,7 @@ The procedure for voting to add new Voting Members is:
   and additional nominations will be accepted until two weeks before the
   election date. The voting will be open for at least two weeks. To successfully
   gain an appointment as a Voting Member, the nominee must receive at least four
-  positive votes, as long as that constitutes a majority of those voting.
+  positive votes, as long as that constitutes a majority of those voting for that candidate.
 * The candidate will be informed promptly at the close of voting by a
   Coordination Committee member. If the vote is not affirmative, the
   Coordination Committee will provide feedback to the nominee.
--- a/APE0.rst
+++ b/APE0.rst
@@ -326,6 +326,8 @@ response is received, the Coordination Committee may automatically change a
 Voting Member’s status to *emeritus*. To record and honor their contributions,
 *emeritus* Voting Members will continue to be listed. *Emeritus* Voting Members
 are not able to participate in votes.
+An emeritus member may request to be marked as active at any time if they feel
+they are making contributions.
 
 Removing Voting Members
 ^^^^^^^^^^^^^^^^^^^^^^^
--- a/APE0.rst
+++ b/APE0.rst
@@ -300,12 +300,14 @@ The procedure for voting to add new Voting Members is:
 * The Coordination Committee and Ombudsperson receive each nomination, check
   that it is factually accurate, that the nominated person accepts the
   nomination, and that their record of community activity adheres to the Code of
-* Once a nomination is accepted by the Coordination Committee, the Coordination
-  Committee will put it to a vote among active Voting Members. Votes for adding
-  voting members will be carried out at most twice a year to avoid too many
-  elections. The voting dates will be announced with at least four weeks notice,
   Conduct. If the candidate accepts, they should provide a brief statement in
   support of their nomination.
+* Once a nomination is accepted by the Coordination Committee, it will be put it
+  to a vote among active Voting Members. Votes for adding voting members will be
+  carried out at most twice a year to avoid too many elections. The returns
+  officer for the voting member election can be a voting member, and those
+  officer are entitled to vote in the election.
+* The voting dates will be announced with at least four weeks notice,
   and additional nominations will be accepted until two weeks before the
   election date. The voting will be open for at least two weeks. To successfully
   gain an appointment as a Voting Member, the nominee must receive at least four
--- a/APE0.rst
+++ b/APE0.rst
@@ -300,11 +300,12 @@ The procedure for voting to add new Voting Members is:
 * The Coordination Committee and Ombudsperson receive each nomination, check
   that it is factually accurate, that the nominated person accepts the
   nomination, and that their record of community activity adheres to the Code of
-  Conduct.
 * Once a nomination is accepted by the Coordination Committee, the Coordination
   Committee will put it to a vote among active Voting Members. Votes for adding
   voting members will be carried out at most twice a year to avoid too many
   elections. The voting dates will be announced with at least four weeks notice,
+  Conduct. If the candidate accepts, they should provide a brief statement in
+  support of their nomination.
   and additional nominations will be accepted until two weeks before the
   election date. The voting will be open for at least two weeks. To successfully
   gain an appointment as a Voting Member, the nominee must receive at least four


APE0.rst
```diff
@@ -122,18 +122,22 @@ Electing the Coordination Committee
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 A Coordination Committee election is held annually to fill any vacant seats.
 Notice of the election must be posted via the Project’s standard communication
-channels (reaching all active members) at least one month before the election
-begins. The election consists of two phases, each lasting two weeks:
+channels for users, contributors and voting members, at least one month before
+the voting phase begins. The election consists of two phases, each lasting at
+least two weeks:
```
Cadair, line 127: This change makes it clear that there doesn't have to be a four week window before nominations open. The objective of this wording is to mandate that nominations should be open for at least two weeks, but to give those running the election more time to process nominations or give notice.

We would recommend using this change to give one weeks notice before nominations open, two weeks of nominations and then one week of admin before the voting starts.

APE0.rst
```diff
@@ -122,18 +122,22 @@ Electing the Coordination Committee
...
 
 * Phase 1, Nominations: Candidates will be nominated and confirm their interest
   in serving. Candidates can be nominated by anyone, including themselves, and
   do not have to be `Voting Members <votingmembers>`_ to be nominated.
-  Candidates should provide a brief statement in support of their candidacy.
+  Candidates should provide a brief statement in support of their candidacy,
+  these statements should be made public as they are received during the
+  nominations period.
```
Cadair, line 134: This change makes it so that nominations have to be made public as they come in, not in one lump at the end of the nominations period.

APE0.rst
```diff
@@ -296,15 +300,18 @@ The procedure for voting to add new Voting Members is:
 * The Coordination Committee and Ombudsperson receive each nomination, check
   that it is factually accurate, that the nominated person accepts the
   nomination, and that their record of community activity adheres to the Code of
-  Conduct.
-* Once a nomination is accepted by the Coordination Committee, the Coordination
-  Committee will put it to a vote among active Voting Members. Votes for adding
-  voting members will be carried out at most twice a year to avoid too many
-  elections. The voting dates will be announced with at least four weeks notice,
+  Conduct. If the candidate accepts, they should provide a brief statement in
+  support of their nomination.
```
Cadair, line 304: This change makes it so voting member candidates have to provide a statement to the voting members supporting their election.

APE0.rst
```diff
@@ -296,15 +300,18 @@ The procedure for voting to add new Voting Members is:
...
+* Once a nomination is accepted by the Coordination Committee, it will be put it
+  to a vote among active Voting Members. Votes for adding voting members will be
+  carried out at most twice a year to avoid too many elections. The returns
+  officer for the voting member election can be a voting member, and those
+  officer are entitled to vote in the election.
```
Cadair, line 309: This change makes it clear that it's ok for a voting member to be the returns officer and to vote in the election while being the returns officer.

APE0.rst
```diff
@@ -322,6 +329,8 @@ response is received, the Coordination Committee may automatically change a
 Voting Member’s status to *emeritus*. To record and honor their contributions,
 *emeritus* Voting Members will continue to be listed. *Emeritus* Voting Members
 are not able to participate in votes.
+An emeritus member may request to be marked as active at any time if they feel
+they are making contributions.
```
Cadair, line 333: Make it explicit that once emeritus you can become active again.

hamogu:
APE 0 says that changes to APE 0 need approval from 2/3 of the voting members. Since this is not terribly urgent, the Coco suggests to just vote on this the next time we have a voting members vote anyway. It's not long until we vote on adding voting members again, so we can just do that at the same time.

Cadair:
My only comment on that would be that this modifies the process by which voting members are elected, with the aim of making that process clearer and easier for those running the next election. I know that this is obviously a trade off with dealing with voting on this APE, but worth considering.

Cadair:
It's also not clear how the vote for the modifications should be run (might modify that language in a new commit), I would interpret that as requiring 2/3s of the voting members to approve the change, but there is no further indication on how that vote should proceed.

We could decide that 2/3s of the voting members should approve this pull request before we merge it.

mwcraig:
> We could decide that 2/3s of the voting members should approve this pull request before we merge it.

The proposal, after some discussion in the CoCo was to simply combine it with the new voting member vote that needs to be done soon (every 6 months, I think?). Not set in stone but it seemed like it would minimize the work needed to get it done.

It does read like it requires 2/3 of all voting members (not just those voting) for the changes to be approved.

pllim (approved):
Seems reasonable to me. Thanks!

APE0.rst
astrofrog: Nominations are not currently public as they require a discourse login - it would be good to make them fully public IMHO

mwcraig: I agree the should be fully public, but when I go to https://community.openastronomy.org/c/astropy/8 in an incognito window I can see the list of CoCo nominations and read them without logging in. Should we also send links to them to the astropy and/or astropy-dev lists?

mwcraig: (so that in addition to being public they are easier to find)

APE0.rst
dhomeier, line 309:
```suggestion
  officers are entitled to vote in the election.
```
Trivial (?) grammar correction – actually, can there be multiple people in the role of returns officer in one election?

APE0.rst
```diff
@@ -296,15 +300,18 @@ The procedure for voting to add new Voting Members is:
...
+  to a vote among active Voting Members. Votes for adding voting members will be
+  carried out at most twice a year to avoid too many elections. The returns
+  officer for the voting member election can be a voting member, and those
+  officer are entitled to vote in the election.
+* The voting dates will be announced with at least four weeks notice,
   and additional nominations will be accepted until two weeks before the
   election date. The voting will be open for at least two weeks. To successfully
   gain an appointment as a Voting Member, the nominee must receive at least four
-  positive votes, as long as that constitutes a majority of those voting.
+  positive votes, as long as that constitutes a majority of those voting for that candidate.
```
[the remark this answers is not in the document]
bsipocz: Why is this? Is it mandatory to select either of the 3 options for all cases to produce a valid ballot (e.g. "abstain" is set as default), or basically two ways of abstinence are possible, only one of which is counted into the election?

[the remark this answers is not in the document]
Cadair: I am trying to remember the details, it's been a while. We wanted people to be able to abstain from voting on individuals, but I think there was some limitation with Helios where we needed to add an abstain button so that this all worked properly. Maybe @mwcraig can remember the details.

From the perspective of APE 0 this is just clarifying that you are allowed to abstain from an individual candidate and not have to abstain from the whole election with multiple candidates. i.e. if there are 20 voting members and 15 of them voted in the election, but 5 abstained in your vote then you would only need a majority of the 10 who voted to vote yay for you to be elected.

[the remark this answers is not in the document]
bsipocz: >  if there are 20 voting members and 15 of them voted in the election, but 5 abstained in your vote then you would only need a majority of the 10 who voted to vote yay for you to be elected.

and that I think is wrong. in the corner case there are 15 who vote in an election but all but one abstain for a candidate, that candidate should not be selected, the majority of the 15 should be what matters.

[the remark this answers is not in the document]
Cadair: I disagree. If you think about it like this: why should an election with only one candidate be any different from one with 10?

If it wasn't for the admin we could elect voting members one at a time when they are nominated, one per election. In my eyes we should allow as much granularity as possible in how the elections run, if people want to abstain from a candidate they should be allowed to, much in the same way they could abstain from a whole election with one candidate in it.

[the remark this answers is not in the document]
mwcraig: > in the corner case there are 15 who vote in an election but all but one abstain for a candidate, that candidate should not be selected, the majority of the 15 should be what matters.

I'd argue that the problem in this sort of case would be with the 14 who choose to abstain; folks would need understand that "abstain" is not "no"; it is, instead, a "count me out for this one".

I can dig up the language we have in conda-forge, which I think also specifies a minimum yes vote required for election.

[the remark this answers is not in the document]
bsipocz: re both of the above: "abstain" is an "I don't care, either way" in my book, but with the caveat of "if others are supporting the idea", and yes, it's stronger than not even picking up the ballot paper. A minimum number of yes votes may be a good idea.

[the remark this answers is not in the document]
Cadair: The wording here is you have to have at least 4 votes and those have to be a majority of those who voted (either yes or no) for that candidate.

adrn:
These changes have been approved by the Astropy voting members in the latest Astropy election https://vote.heliosvoting.org/helios/elections/707e7c5c-fdf6-11ec-99a3-86fb037d8109/view

APE0.rst
Cadair, line 309:
```suggestion
  officer(s) for the election of the voting members can be a voting member, and those
  officer(s) are entitled to vote in the election.
```

Cadair: I have proposed a slightly different tweak here: https://github.com/astropy/astropy-APEs/pull/71/files#r928318906

Cadair: @mwcraig @adrn Is the slight grammar tweak good to merge?

APE0.rst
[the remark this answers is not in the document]
mwcraig: Ah, right. the full clause is "To successfully gain an appointment as a Voting Member, the nominee must receive at least four  positive votes, as long as that constitutes a majority of those voting for that candidate." 

The intent of this change was to avoid the situation where a candidate receives 4 yes votes and 16 no votes -- in the original wording that person would be appointed. 

I think we included an abstain option on the ballot so that we could make every question on the ballot required. 

I see a couple options here:

1. Leave as-is, since there is a minimum number of required yes votes (so a 1 yes, 16 abstain would fail).
2. Change the minimum of yes votes (would require a new APE 0 amendment, which I'm happy to propose)
3. Impose a minimum percent of voting members for a quorum. (this is what conda-forge does)

Regarding 3:Z conda-forge has had several problems meeting that quorum, so there are a couple of sort-pf-complicated ways to skirt the quorum (see https://conda-forge.org/docs/orga/governance.html#voting and scroll down to the section about quorum).

IIRC, we went with an absolute number here just in case apathy sets in. It could be changed to, for example, X number or Y% of voting members, whichever is higher. In that case I'd propose something like X=4 and Y=10.

APE0.rst
mwcraig: I'm inclined to say no since it isn't quite what the voting members voted for.

APE0.rst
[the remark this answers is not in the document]
bsipocz: Re 3: this is what PSF does, too, at least 1/3 of all vote eligible members should cast a vote for an election to be valid, otherwise it's a void one. I would be very much for having policies that follow more closely of PSF, e.g. just as we originally proposed. mostly covered in section 3.9: https://www.python.org/psf/bylaws/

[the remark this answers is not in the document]
bsipocz: e.g. the 4 is already <10% of voting members, and that percentage will keep shrinking more.

Cadair:
Can this be merged then?

APE0.rst
dhomeier: Since there has been significant discussion beyond just grammatical issues after voting had started, should all those be moved to a follow-up PR then? Possibly with a tag indicating that the PR is frozen before voting starts, and to be merged as is including spelling errors and everything.

mwcraig: Yes -- I think the next step is to merge this PR (which passed) and open a new PR for discussion then a vote along with the next CoCo election, which needs to be held sometime in the next few months.

That PR would include at least  a proposed change in the minimum number of participants needed to elect someone (to raise it from 4) and this grammatical fix.

Our next CoCo meeting will be this week or next, I will add it to the agenda.

eteq:
There was some discussion on changes above in the last few weeks *after* the vote occurred. So after discussion in the CoCo we agreed it makes sense to merge this *as-is* since the vote succeeded, but the two items above (https://github.com/astropy/astropy-APEs/pull/71#pullrequestreview-1048691995 and https://github.com/astropy/astropy-APEs/pull/71#pullrequestreview-824151602) will go in a follow-on change to be included in the next CoCo election.

PR #76, opened 2022-08-25
Add APE to adopt Black formatting in astropy

taldcroft:
This APE proposes that the astropy core package adopt [Black](https://github.com/psf/black) as the default code format.

Following discussion of this topic on astropy-dev, @eteq and @taldcroft developed this together over a series of virtual meetings, sharing our different perspectives and working to converge to a common view. We are hoping this draft APE will generate constructive discussion and hopefully lead to adoption of the APE.


APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+
+History
+~~~~~~~
+This changes almost every Python file. Git blame can skip the commit, but git
+diff won't. But we are already doing other bulk changes that hit every file.
+
+Code correctness
+~~~~~~~~~~~~~~~~
+Black guarantees correctness of the code formatter to not change the code AST.
+
```
kelle, line 159: What is "AST"?

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+design, most of Black_ is inflexible, and there is not a way to pick-and-choose
+some elements and not others of Black's format.
+
+Implementing this change requires:
+
+1. Updating the `coding style`_:
+
+   * Adding documentation about Black, with a particular emphasis on the rationale expressed here and the exception policy described above.
+   * Adding documentation about using pre-commit, ala `pandas pre-commit`_.
+   * Document the black exception guidelines (as described in this APE)
```
kelle, line 218:
```suggestion
   * Document the Black exception guidelines (as described in this APE)
```

kelle (commented):
non-substantive edits. LGTM.

APE_Black_Formatting.rst
WilliamJamieson: Abstract Syntax Tree. its what your Python code gets parsed into. The python interpreter then executes the AST to run your code.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+settings, that is, 88 characters per line and double quotes. Note that, by
+design, most of Black_ is inflexible, and there is not a way to pick-and-choose
+some elements and not others of Black's format.
+
+Implementing this change requires:
+
+1. Updating the `coding style`_:
+
+   * Adding documentation about Black, with a particular emphasis on the rationale expressed here and the exception policy described above.
+   * Adding documentation about using pre-commit, ala `pandas pre-commit`_.
```
WilliamJamieson, line 217: Adding `pre-commit` documentation is already in progress as part of astropy/astropy#13552.

We already provide a `pre-commit` configuration and use it for our current "Code style checks" already, it just has not been properly documented in the `astropy` documentation.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+   does not provide any added checks that ``black –check`` does not.
+
+3. Checking Black in CI.
+
+   * Add a CI check that runs ``black --check`` on the whole code base, failing
+     if it meets a violation
+   * Allow for the Black CI bot to fix formatting errors via documented single
+     line comments. This is an opt-in step which could be used by maintainers to
+     fix errors prior to merging. Having such an option available could lower
+     the barrier for new contributors.
```
WilliamJamieson, line 233: The usage of the [pre-commit.ci](https://pre-commit.ci/) bot proposed in astropy/astropy#13320, see the implementation plan https://github.com/astropy/astropy/issues/13320#issuecomment-1171488051 will perform all of these tasks for us. The function of this bot will be documented by astropy/astropy#13553 when we enable the bot.

Note the usage of pre-commit and the related pre-commit.ci bot are separate issues than adopting black.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Implementation
+--------------
+
+All Python code in ``astropy`` is formatted with Black_, using its default
+settings, that is, 88 characters per line and double quotes. Note that, by
+design, most of Black_ is inflexible, and there is not a way to pick-and-choose
+some elements and not others of Black's format.
+
+Implementing this change requires:
```
WilliamJamieson, line 212: PR astropy/astropy#13253 sets up the changes to `astropy` to enable enforcing `Black` on a per-subpackage basis, but does not run black on any part of `astropy`. Note that the changes to look at are to the `.pre-commit-config.yaml` and `pyproject.toml` files, all the other file changes are due to a necessary change to the configuration for `isort` so that it is compatible with `Black`.

One minor change to this PR, so that it fits the currently proposed APE, is updating the `line_length` values (for `flake8`, `isort`, and `Black`) so that they all conform to the default `Black` line length of 88.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+     pull request.
+   * The maintainer(s) should review all changes and highlight for discussion
+     any  changes that are judged to be valid exceptions (see above for what
+     "valid exception" means). These can be reverted by copying-and-pasting
+     from the "changes" interfaces to the local copy and applying the
+     appropriate Black directives to prevent auto-formatting. Commit and push
+     these changes for review.
+   * Once the PR submitter and reviewer are both satisfied with the changes
+     then squash the commit(s) down to a single commit and force push.
+   * Run ``black astropy/<sub-package>`` to apply just the string normalization
```
WilliamJamieson, line 253: Note that string normalization converts the quotes around all strings to be `"` unless the string in question contains a `"` character.

WilliamJamieson:
An add-on comment, should we consider applying [`isort`](https://pycqa.github.io/isort/) to `astropy` while performing the all the `Black` changes. Tagging @nstarman to explain `isort` in more detail.

APE_Black_Formatting.rst
nstarman: Or changing Black to a line length of 100 (https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html?highlight=length#line-length)

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+thinks the improved readability warrants it.
+
+History
+~~~~~~~
+This changes almost every Python file. Git blame can skip the commit, but git
+diff won't. But we are already doing other bulk changes that hit every file.
+
+Code correctness
+~~~~~~~~~~~~~~~~
+Black guarantees correctness of the code formatter to not change the code AST.
```
nstarman, line 158:
```suggestion
Black guarantees semantic equivalence of the formatted code, and will not (`except in a few limited cases <https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#ast-before-and-after-formatting>`_) change the code AST.
```

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+This is addressed in the Implementation section.
+
+New contributors
+~~~~~~~~~~~~~~~~
+New contributors may be more discouraged by more steps. However:
+
+This *replaces* the existing PEP8 check, so that it is not "another step" To
+make it better, leave very clear explicit messages in the status checks on GH
+(not putting up a barrier) Scikit-learn had a few issues with contributors but
```
nstarman, line 176: Some sentence fragments here.

nstarman (commented):
If it's not too burdensome for this APE, bundling ``black + isort`` might be worthwhile. They are are frequently used together and we already have a similar per-package setup for ``isort``...

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+type: Standard Track
+
+status: Discussion
+
+
+Abstract
+--------
+
+This APE proposes that the `astropy core package
+<https://github.com/astropy/astropy>`_ adopt Black_ as the default code format.
```
astrofrog, line 26: Is the word 'default' needed here, or what about just putting 'adopt Black for formatting code'?

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Detailed description
+--------------------
+
+Foreword
+^^^^^^^^^
+
+This APE presents a rationale that code formatting with Black_ can be beneficial
```
astrofrog, line 34: This might be missing a paragraph saying what Black actually is

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+surround the matrix with the ``#fmt: off``/``#fmt: on`` escape hatch, as this
+clearly makes the code more readable.  While many examples are less clear than
+the above, and indeed in most cases "readability" is a subjective concept, this
+mechanism allows for such cases to be explicitly excepted where the developer
+thinks the improved readability warrants it.
+
+History
+~~~~~~~
+This changes almost every Python file. Git blame can skip the commit, but git
+diff won't. But we are already doing other bulk changes that hit every file.
```
astrofrog, line 154: I'm not sure that touching every file is really that bad, the concern is more here that it might in fact touch most lines (which is still fine because we can mark the commit to be skipped, but I do think this is unprecedented). If you keep the last sentence, give some examples?

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Code correctness
+~~~~~~~~~~~~~~~~
+Black guarantees semantic equivalence of the formatted code, and will not (`except in a few limited cases <https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#ast-before-and-after-formatting>`_) change the code AST.
```
astrofrog, line 158: Wrap text and spell out and explain what AST is.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Backports / release branches
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+This is addressed in the Implementation section.
+
+New contributors
+~~~~~~~~~~~~~~~~
+New contributors may be more discouraged by more steps. However:
+
+This *replaces* the existing PEP8 check, so that it is not "another step" To
```
astrofrog, line 174: Is it necessary to have the paragraph break with the :?

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+should be excluded from Black formatting.**
+
+
+Branches and pull requests
+--------------------------
+
+- `Add configuration for black autoformatter <https://github.com/astropy/astropy/pull/13253>`_
+- `Apply black to modeling <https://github.com/astropy/astropy/pull/13254>`_
+
+Implementation
```
astrofrog, line 204: This section is missing something about how we will deal with release branches, and when the change will be done, both to minimize impact on release branches and open PRs. As far as I can see there are two options:

* Make the change just before the first 6.0 release candidate is made, because at that point the release branches will no longer be active and open PRs may be at a minimum.
* Have a sprint some other time to minimize open PRs, then pick an arbitrary date to make the change and simultaneously reformat ``main`` and any active release branch.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+   * Convert the draft PR to a normal PR (ready for review). Once it passes CI
+     then it can be merged without further review.
+
+A sub-package will not have Black checked in CI until it has completed these
+steps.
+
+Backward compatibility
+----------------------
+
+This APE renders all previous code formatting styles no longer compatible. In
```
astrofrog, line 266: Not clear what 'compatible' means - maybe better said as 'This APE would superseded any previous code formatting styles recommended for the core package'

astrofrog (commented):
I'm generally on board with this but I would like to see more details on how/when we will do this logistically to minimize impact.

APE_Black_Formatting.rst
Cadair: One line per sentence is the correct way :wink:

bsipocz: you mean one line per word 😉

taldcroft:
> I'm generally on board with this but I would like to see more details on how/when we will do this logistically to minimize impact.

This is where I'm hoping that folks in the infrastructure team can help out. Personally I don't know enough of the details in that area to make a detailed impact statement. 

One thing I can say is that it is correct that a large fraction of lines will be touched. Applying black to `astropy/time` resulted in `22 files changed, 3599 insertions(+), 2599 deletions(-)` in about 12000 lines of code (total lines, not just source lines).

APE_Black_Formatting.rst
taldcroft: I like the second option. 

The main concern there is the extent of exceptions that are put in by hand for `main`. The APE is explicit in allowing exceptions, but I don't see an easy way to apply those exceptions to formatting any active release branch -- they would be formatted with a single run of `black` on the entire repo. I'm hoping that in practice the exceptions will be rare and won't generate conflicts.

APE_Black_Formatting.rst
taldcroft: Indeed this touches about a 1/3 of lines in the `time` subpackage (see below). So I'm happy to rewrite this paragraph as:
```
This changes almost every Python file and a large fraction of code lines 
(for example about one third of non-blank lines in the ``astropy.time`` 
subpackage). This is more than previous bulk changes and it is an impact that
needs to be accepted with this APE.
```

APE_Black_Formatting.rst
taldcroft: Will fix.

APE_Black_Formatting.rst
taldcroft: Will do.

APE_Black_Formatting.rst
taldcroft: Yes, will do.

APE_Black_Formatting.rst
WilliamJamieson: I agree as well. Indeed we may want to plan a sprint to coincide with the 6.0 release so that we can push all these changes in quickly. The creation of the initial `Black` changes will be simple, but reviewing the changes for any possible exceptions will be the tedious part, and we may want to have people ready to do this work.

I worry about any remaining open PRs after `Black` changes are merged. We should work with the contributors of any of the remaining open PRs to both warn them and help them get their PRs migrated.

APE_Black_Formatting.rst
taldcroft: Will do.

APE_Black_Formatting.rst
taldcroft: Will fix.

APE_Black_Formatting.rst
taldcroft: Default (88) please! :smile:  I've been using Black with 88 and I like it.

taldcroft:
> An add-on comment, should we consider applying [isort](https://pycqa.github.io/isort/) to astropy while performing the all the Black changes. Tagging @nstarman to explain isort in more detail.

I've gone back and forth, but it seems like keeping them separate might be easier. Since `isort` is less controversial, I could see doing that for the entire astropy package in one commit. I think that would be easier and decouple this from the package-by-package process for black.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+
+While there are multiple code formatters for Python, the Black Python code
+formatter produces good enough results that it is getting significant traction
+in many open source projects, including: scikit-learn, pytest, tox, Pyramid,
+Django, Hypothesis, SQLAlchemy, virtualenv, pandas, Pillow, and Twisted. While
+adoption by this list of high-profile community-based open source projects does
+not automatically imply that it is a good fit for ``astropy``, it demonstrates
+unequivocally several points:
+
+Technical issues related to the initial transition and subsequent maintenance of
```
mwcraig, line 97: Not a big deal, but seems like this should be a bulleted list.

mwcraig (approved):
LGTM, just one or two minor comments.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+Some developers may feel that individual style and humanity gets lost in
+autoformatting. We hope to convey that humanity and individuality is expressed
+more fully in the elegance and clarity of the code logic and implementation. In
+most cases the placement of white space and line breaks can be viewed as a
+mechanistic process that humans need not control.
+
+Exception Policy
+^^^^^^^^^^^^^^^^
+As described above, there are cases Black formatting makes the code
+substantially more difficult to read or understand.. These should be treated
```
hamogu, line 191:
```suggestion
substantially more difficult to read or understand. These should be treated
```

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
+.. _Black: https://github.com/psf/black
+.. _Django DEP-0008: https://github.com/django/deps/blob/main/final/0008-black.rst
```
bsipocz, line 2: I feel bringing Django into the argument is comparing apples to oranges. black is absolutely awful with mathematical expression formatting as well as with nested containers, so comparing the astropy case to e.g. scipy would be more realistic.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+
+.. _astropy-dev thread: https://groups.google.com/g/astropy-dev/c/6cRJCMgaFyM/
+
+Note that ideas and some text for this APE have been adapted from a similar
+enhancement proposal for Django `Django DEP-0008`_.
+
+Rationale
+^^^^^^^^^
+
+**Consistent and readable code formatting** reduces effort in both code review
```
bsipocz, line 58: consistent and readable code formatting can be achieved without black. However there has been a pushback from the community for year to enforce linting, even though that step could be done along with pre-commit hooks, and without black. Forcing the pre-commit hook to be run would similarly result of fixed code without separate pep8 commits, so it is kind of an unfair accounting to claim that black would fix this issue.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+Automated code formatting will increase consistency across the ``astropy`` code
+base. Currently the style of each subpackage reflects the preferences of the
+maintainers. For instance, the ``time``, ``table``, and ``io.ascii`` subpackages
+allow line lengths of up to 100, while most others use 80 or 88. There are other
+more subtle differences and astropy contributors find this confusing. This has
+come up in the past.
+
+While there are multiple code formatters for Python, the Black Python code
+formatter produces good enough results that it is getting significant traction
+in many open source projects, including: scikit-learn, pytest, tox, Pyramid,
```
bsipocz, line 91: for completeness I would also suggest to cite the  projects that decide of not to go down this route. E.g. quite a couple of projects are waiting for black to implement some reasonable approach for mathematical and nested structure formatting. 

For an absolutely ugly reformatting example, one can look at scipy: https://github.com/scipy/scipy/pull/14330 where black wouldn't result a more readable code, but a bloat of 45% of number of lines.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Undesirable formatting
+~~~~~~~~~~~~~~~~~~~~~~
+On occasion Black's formatting produces undesirable results. The code may look
+very different from expectation due to mathematical or other domain-specific
+conventions, or some specific formatting that conveys meaning may be lost. A
+simple example of this is a constant identity matrix which would typically be
+written as::
+
+    matrix = [[1, 0, 0],
```
bsipocz, line 135: An even more undesirable formatting (though not with this small, int array, but a large float one would be the examples present in the scipy PR: https://github.com/scipy/scipy/pull/14330)


```
self.a = asarray([1.5750000E-04, 1.6990000E-04, 2.3500000E-04,
                          3.1020000E-04, 4.9170000E-04, 8.7100000E-04,
                          1.7418000E-03, 4.6400000E-03, 6.5895000E-03,
                          9.7302000E-03, 1.4900200E-02, 2.3731000E-02,
                          4.0168300E-02, 7.1255900E-02, 1.2644580E-01,
                          2.0734130E-01, 2.9023660E-01, 3.4456230E-01,
                          3.6980490E-01, 3.6685340E-01, 3.1067270E-01,
                          2.0781540E-01, 1.1643540E-01, 6.1676400E-02,
                          3.3720000E-02, 1.9402300E-02, 1.1783100E-02,
                          7.4357000E-03, 2.2732000E-03, 8.8000000E-04,
                          4.5790000E-04, 2.3450000E-04, 1.5860000E-04,
                          1.1430000E-04, 7.1000000E-05])
```

vs
``` 
        self.a = asarray(
            [
                1.5750000e-04,
                1.6990000e-04,
                2.3500000e-04,
                3.1020000e-04,
                4.9170000e-04,
                8.7100000e-04,
                1.7418000e-03,
                4.6400000e-03,
                6.5895000e-03,
                9.7302000e-03,
                1.4900200e-02,
                2.3731000e-02,
                4.0168300e-02,
                7.1255900e-02,
                1.2644580e-01,
                2.0734130e-01,
                2.9023660e-01,
                3.4456230e-01,
                3.6980490e-01,
                3.6685340e-01,
                3.1067270e-01,
                2.0781540e-01,
                1.1643540e-01,
                6.1676400e-02,
                3.3720000e-02,
                1.9402300e-02,
                1.1783100e-02,
                7.4357000e-03,
                2.2732000e-03,
                8.8000000e-04,
                4.5790000e-04,
                2.3450000e-04,
                1.5860000e-04,
                1.1430000e-04,
                7.1000000e-05,
            ]
        )
```

APE_Black_Formatting.rst
bsipocz: the enforcing of pre-commit bot, without black could be compatible, and therefore I feel it's a mistake to merge these two proposals together and sell the advantages of one as the advantage of the other and a reasoning to live with the disadvantages. (e.g. pre-commit without black but enforcing linting has the huge potential to be the middle ground)

APE_Black_Formatting.rst
taldcroft: @bsipocz - good idea, can you provide a list of other projects that have decided to either (1) reject auto-formatting entirely as being a bad idea or (2) wait until the black implementation is better.

The APE states that cases where black does poorly can and should be excepted. Based on testing by myself for time/table/io.ascii and @eteq (coordinates), requiring exceptions is not common but they will be allowed. Any others e.g. @WilliamJamieson @nstarman have data on the need for exceptions?

APE_Black_Formatting.rst
taldcroft: Since this example is not even in astropy, and we would just skip black formatting anyway, I'm not sure what this is telling us.

APE_Black_Formatting.rst
bsipocz: a few projects (scipy is rejected, numpy is waiting, they may be others, too) are waiting for the math formatting summit, and the promise from the side of the black developers to take scientific python needs into account. Most of those mailing list and discourse and PR discussions are linked to the scipy and numpy PRs cited above.

afaik all that implemented it and likes it, in fact likes the fact that linting is now enforced, rather than the formatting itself, and thus my opinion of enforcing linting would do the job without enforcing to live together with the nonsense of the choices black forces on us (e.g. all of those in the project were against linting cited less serious issues and have been unwilling to accept the choices of other team members, but now are willing to accept the nonesences around the 3rd party, black's verbose whitespacing?)

APE_Black_Formatting.rst
bsipocz: >  and we would just skip black formatting anyway

Then what this APE is about, did I miss the suggestion of not accepting the APE? IMO this is a very much representative example of what's wrong with black, more so than the example in the APE. And there are arrays in astropy, much more than in some other packages that are cited as examples what opted in using it.

APE_Black_Formatting.rst
taldcroft: Django is similar to astropy in being a large community-driven open-source project with a range of experience levels for contributors. So the comparison is useful in terms of developer and community experience.

As mentioned elsewhere, formatting of the code and the small fraction of cases where this goes badly is not a problem when you actually run black on astropy and look at the results. For instance try `modeling`, which is probably one of the more math-heavy parts of astropy. A lot of equations are changed, and IMO the changes are mostly improvements and at worst neutral.

But again, if the developer and PR reviewer both agree that black is doing badly then the that code can be skipped.

APE_Black_Formatting.rst
taldcroft: @bsipocz -I don't know precisely what you mean by "enforcing linting". We already do CI flake8 checks, so do you mean something different?

What I like most about black is fast *auto-formatting* in my IDE. But in general I am happy with the format choices black makes, so you should not presume to speak to the opinions of the large community of black users.

APE_Black_Formatting.rst
WilliamJamieson: As far as pre-commit and the pre-commit bot are concerned, the infrastructure team has been moving forward with using them in astropy independent of this APE. The introduction of pre-commit has already occurred, while this APE was being drafted it was realized that pre-commit was not documented for astropy developers. The use of the bot is a natural extension of using pre-commit which should help ease the burden of any current (or future) linting or style checks on contributions.

APE_Black_Formatting.rst
bsipocz: currently, a select number of linting options are checked, rather than only a few ignored, and that is historical reasons as there was a push against being strict on linting, being conservative and allowing interpretation of the pep8 rules as basically, the aim was to keep the heterogeneity of the different submodules. We wanted and could have done e.g. autopep8 to auto format those things that are uncontroversial. 
So now, the proposal is to be being instead extra strict and enforcing linting choices that are blindly doing PEP8 and not necessarily ideal for the codebase we have. 

And I don't claim to know the opinion of all black users, but say that there is a well documented shared concern in the wider scientific python community that its enforced code style is not ideal for math and array-heavy usage. Beyond this shared concerns I have personal issues with the superfluous white spaces it applies, the 30-45% addition of lines with no useful information (single parens or characters).  I also have personal preferences for IDE and editor behaviour which I'm very happy with, but am not trying to force that on all the other developers.

APE_Black_Formatting.rst
taldcroft: I was wondering about a hybrid of the options -- 

After the normal v5.2 feature freeze on 2022-10-28, then have a 2-week period for doing the `Black` transition of all sub-packages. I think the initial PR's could be done by a core team of a few, but this requires some buy-in from sub-package maintainers to do the reviews. By doing this in a special post-freeze period this might minimize the impact on open PR's. Many PR's that stay open through a feature freeze are not being actively worked.

The 6.0 freeze is not until 2023-10-27 so I hope we don't need to wait that long!

taldcroft:
I think I have addressed all the review comments requesting a change in the APE text.

taldcroft:
About the review comments, @bsipocz is fundamentally and strongly against acceptance of this APE, so her dissent is noted here in the APE review but I did not change any of the APE to reflect her views.

APE_Black_Formatting.rst
nstarman: I like the sound of this (esp the earlier deadline 😆 )

APE_Black_Formatting.rst
nstarman: > The APE states that cases where black does poorly can and should be excepted. Based on testing by myself for time/table/io.ascii and @eteq (coordinates), requiring exceptions is not common but they will be allowed. Any others e.g. @WilliamJamieson @nstarman have data on the need for exceptions?

In my experience, I haven't found too many places where exceptions are needed.
Sometimes / often a blackened math one-liner looks worse because of the spaces around operators. However, I tend to have comments by math code with a cleaner representation of the math, because except for very simple expressions, I don't find math in **Python** very readable, regardless of style. For multi-line complex math, my thoughts on Python and needed explanatory comments are even more true. IMO Black doesn't usually make the **end** result worse as it encourages me to rewrite the math.

tl;dr, piles of multi-line math are much less readable for future maintainers than well factored code with comments. Sometimes multi-line math is unavoidable, which is why we can selectively turn black off.

APE_Black_Formatting.rst
nstarman: Is this an example from the test suite?

nstarman: I'm certainly guilty of this, but we should probably minimize hardcoding in arrays for most tests, and instead use more elegant tools like https://github.com/astropy/pytest-arraydiff

bsipocz: This is an example from the scipy PR. IMO you cannot ignore mentioning in the APE that neither scipy nor numpy opted in black, exactly because of math and array issues like e.g. this one. 
The scipy PR also shows the bloating up the codebase with whitespaces. 

And they are both very large and diverse contributor base, and much more similar to the one astropy is having, consisting of both scientists and engineers, as opposed to e.g. django.

hamogu: I agree that the numpy and scipy PRs and associated discussion should be mentioned. That does not automatically mean that we need to follow their lead, but mentioning only Django looks like cherry-picking.

taldcroft: @bsipocz @hamogu - I added 884b0a1. Is that sufficient?

hamogu: Sounds good to me. Thanks!

APE_Black_Formatting.rst
taldcroft: Release managers @astrofrog @bsipocz @saimn @embray, and others - are you OK with the plan of a 2-week post-freeze period to do the Black transition PRs for sub-packages? Or does anyone have a better plan?

eerovaher: [`darken`](https://pypi.org/project/darker/) can apply `black` formatting to only the lines that are being edited. This would allow for a transition period during which all new code would be `black`-formatted, so when the final switch to `black` finally happens the diffs would be smaller.

astrofrog: I'm happy with a 2-week post-freeze period, though we will need to decide what to do for sub-packages where maintainers end up being unavailable to review.

I think we might want to not run black on the LTS branch but as that branch has existed for almost a year the number of bug fixes that backport cleanly is going to be more and more limited, so maybe we just don't format the LTS branch and deal with conflicts by hand.

astrofrog:
A general question - as this is one of the more controversial APEs that has been re-written, and it is unlikely there will be universal consensus, how is the decision process going to be handled? Will a vote be carried out amongst voting members? core package maintainers? Or will this be left to the CoCo to decide?

taldcroft:
About process, APE 1 says *The final decision on any APE is made by the coordinating committee, though usually a consensus in the development is sufficient (but in unusual cases may be overridden by the coordinating committee).*

I had been thinking that anyone that wants to weigh in should put in a review, which will give the CoCo information for their decision. But ultimately I think the CoCo needs to (1) fully define the process and (2) make a decision.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,307 @@
+.. _Black: https://github.com/psf/black
+.. _Django DEP-0008: https://github.com/django/deps/blob/main/final/0008-black.rst
+.. _pandas pre-commit: https://pandas.pydata.org/docs/development/contributing_codebase.html#pre-commit
+.. _coding style: https://docs.astropy.org/en/latest/development/codeguide.html#coding-style-conventions
+
```
eerovaher, line 5: [PEP 8](https://peps.python.org/pep-0008/) is referred to in the text, there's no reason not to list it here.
```suggestion
.. _PEP8: https://peps.python.org/pep-0008/

```

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,307 @@
...
+generated a spirited discussion but did not result in a clear consensus. This
+strong division of opinions about automated code formatting is common and has
+been documented in other projects, for instance in `Django DEP-0008`_.
+
+Other projects, including NumPy and SciPy, have considered using Black for
+formatting (see e.g. `SciPy PR #14330
+<https://github.com/scipy/scipy/pull/14330>`_). Neither of those projects felt
+that auto-formatting with the current version of Black (circa 2022) was a good
+choice for their codebase, with particular emphasis on needing improvements
+in formatting of mathematical expressions.
```
eerovaher, line 60: It is worth pointing out that the SciPy issue mentions psf/black#538, which is resolved by now, so not all of the SciPy discussion is relevant anymore.

eerovaher (commented):
The amount of commits that do nothing but fix codestyle problems shows that some sort of automatic formatting would be beneficial. Although `black` is not the perfect tool for mathematical code, it works quite well in many cases. I would be in favor of adopting `black`, but I would also be willing to consider other code formatting tools if anyone makes a specific proposal.

APE_Black_Formatting.rst
Cadair: I heard we were painting bike sheds. Longer lines ftw.

astrofrog: Split the difference, and go with 94?

taldcroft: For my exact laptop and editor configuration, I can get two screens of 84 characters split vertically. But to be honest I very rarely do that split, so I'm fine with any value between 88 and 94. I ran black with a width of 94 and that does reduce the diffs a bit.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,308 @@
...
+^^^^^^^^^
+
+**Consistent and readable code formatting** reduces effort in both code review
+and in future maintenance. Code is read far more often than it is written, so it
+makes sense to invest effort to ensure that the code is easy to read later.
+
+Ensuring that code formatting meets ``astropy`` standards is currently done by
+applying CI checks based on :pep:`8`. Learning and consistently applying all
+these rules is not easy. There are at least 200 commits in ``astropy`` which
+were solely address PEP8 violations, so this is a real issue.
```
eerovaher, line 81: This claim about more than 200 commits fixing code style originates from the mailing list, and the number was obtained as follows (using the current `main` e4993fffb54e19b04bc4e9af084984650bc0a46f):
```shell
$ git log --grep="PEP\s*8" --pretty=oneline | wc -l
229
```
A more thorough search reveals an even bigger problem:
```shell
$ git log --grep="PEP\s*8" --grep="pep\s*8" --grep="flake\s*8" --grep="code\s*style" --pretty=oneline | wc -l
409
```

bsipocz: But these are due to the non-enforcing nature of formatting and not due to not using black and previous resistance of being more strict with pep8 enforcement. A properly set up editor with flake8/pre-commit runs with it, etc would have just as well avoided the extra commits.

taldcroft: Are you talking about autopep8 in pre commit? Or something else?

bsipocz: yes, for example, that. Or a properly set up editor that picks up on the pycodestyle/flake8/etc config from the repo.

Or simply maintainers pressure to make contributors squash commits down to logical chunks.

taldcroft: @bsipocz - about `autopep8` as an autoformatter -- that loses the advantage of the style uniformity and readability improvements that black brings (in the opinion of the proponents of this APE). Basically for something like below (and *many* other cases in the astropy code base), autopep8 is perfectly happy with the original because it is indeed PEP8-compliant.

bsipocz: The APE use the argument that we need black as we have 200 commits of pep8 fixes. And all in saying, if we could have been strict with enforcing formatting before we would have had less.

Now, this particular example you bring: The original 3 lines are perfectly readable for me, especially in the context it's used, much more than the proposed 11. And I've brought examples with maths and nested containers where it's difficult to argue that the black version is really acceptable.

taldcroft: @bsipocz - fair enough and thanks for all your input. This APE starts with a Foreword that mentions not expecting consensus. I think we have realized that expectation and I do not see any way to further modify this APE in a way that will be acceptable to both you and the APE authors.

This PR has been open for discussion for 12 days now, with not much activity in the last week. Other comments have neutral to positive, mostly focusing on improving the APE (wording, rationale, and implementation). I'd like to ask the (unconflicted) CoCo members @hamogu @kelle @adrn @mwcraig  for direction on going forward from here.

adrn: Thanks for the ping @taldcroft -- as @hamogu noted in his email, this is on our agenda to discuss on Monday.

APE_Black_Formatting.rst
WilliamJamieson: astropy/astropy#13552 has been merged, updating the documentation about how to use pre-commit is no longer necessary.

APE_Black_Formatting.rst
WilliamJamieson: The pre-commit.ci bot was turned on today and documented with the merger of astropy/astropy#13353, so in the event that this APE is accepted, simply adding Black (or some other formatting tools) to the pre-commit configuration will be sufficient to enable opt-in automatic fixing.

APE_Black_Formatting.rst
nstarman: Personally, I've found ``darken`` more finicky than ``black`` to get working with IDEs. Not saying I'm against it, but going all in would be my preference.

APE_Black_Formatting.rst
adrn: I'm used to the default 88 but I don't feel very strongly.

(When I first read this APE, I was still pro-`yapf` because of the customizable features (esp. around array reformatting), but now that I see how much discussion we would have around each setting, I'm glad `black` only has a few 😅.)

eerovaher: 88 is plenty, I see no reason not to use the default value.

astrofrog: I don't really mind either way, my suggestion of 94 above wasn't necessarily very serious.

Cadair:
In the name of science: https://github.com/astropy/reproject/pull/308

taldcroft:
@Cadair - nice example. I made a suggestion on splitting the difference https://github.com/astropy/reproject/pull/308/files/233ccd41bd06fd91f0c856ae1cc6c8890ff71721#r966983968.

APE_Black_Formatting.rst
taldcroft: OK, I'll revert back to 88. I don't mind, I just want *something*. :smile: But personally I like the simplicity of using the default.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,306 @@
...
+
+2. Updating flake8 configuration to be compatible with Black.
+
+   This is straightforward and well documented. It may even be removed if it
+   does not provide any added checks that ``black –check`` does not.
+
+3. Checking Black in CI.
+
+   * Add a CI check that runs ``black --check`` on the whole code base, failing
+     if it meets a violation
```
Cadair, line 251: This is potentially conflicting with the new use of pre-commit for these style checks as I believe it only checks the diffs. I personally don't really see a reason why we shouldn't continue to use pre-commit & pre-commit.ci so I propose the following:

```suggestion
   * Add a CI check that runs black to ensure consistency, failing if it meets a violation.
```

Being overly prescriptive here is not needed.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,306 @@
...
+
+   * Adding documentation about Black, with a particular emphasis on the
+     rationale expressed here and the exception policy described above.
+   * Document the Black exception guidelines (as described in this APE)
+   * Removing other references to code formatting in the documentation.
+
+2. Updating flake8 configuration to be compatible with Black.
+
+   This is straightforward and well documented. It may even be removed if it
+   does not provide any added checks that ``black –check`` does not.
```
Cadair, line 246: It does provide checks that black doesn't, such as checking that things in the `__all__` are defined in the code etc.

APE_Black_Formatting.rst
```diff
@@ -0,0 +1,306 @@
...
+
+3. Checking Black in CI.
+
+   * Add a CI check that runs ``black --check`` on the whole code base, failing
+     if it meets a violation
+   * Document the process for using the pre-commit CI bot to fix formatting
+     errors via single line comments. This is an opt-in step which can be used by
+     maintainers to fix errors prior to merging.
```
Cadair, line 254: fwiw this is already done: https://docs.astropy.org/en/latest/development/workflow/maintainer_workflow.html#fixing-coding-style-issues

APE_Black_Formatting.rst
Cadair: That is what I wanted to ensure was possible with this wording. We will be running black as part of the CI (in pre-commit) so imo my suggestion means this APE allows us to do the easiest thing.

Cadair (commented):
Aside from technical considerations and details, I am still decidedly neutral on this.

I think that black is ultimately lazy, but given that in my opinion a lot of things in astropy are not lazy enough, this probably as good of a place to start as any.

APE_Black_Formatting.rst
taldcroft: Fixed.

APE_Black_Formatting.rst
taldcroft: Fixed.

taldcroft:
@Cadair - thanks for the review and inputs. I think I incorporated your points and am definitely :+1: on not being overly prescriptive here. 

 @WilliamJamieson - thanks for continued helpful inputs.

WilliamJamieson (approved):
`black` is not a perfect solution to formatting e.g. https://github.com/astropy/astropy-APEs/pull/76#discussion_r956367409.  However, I think a lot of these issues are surmountable via refactoring the code (minimizing hard coded arrays) or turning off `black` for limited sections of the code. Moreover,  it is well supported by the PSF and seems to have a wide general adoption. In any case, I feel like astropy needs a stronger formatting policy which can be automatically implemented and enforced; `black`, in my opinion, represents the cleanest way to achieve this. If there is strong disagreement with using `black` I invite those parties to propose alternative tools which implement automatic formatting and enforcement; I am open to better "solutions" to code formatting in astropy than our current lack of one.

Thus, I support this APE. If its accepted I volunteer to help with the implementation of black in astropy.

nstarman (approved):
I don't love ``black`` for math or hard-coded arrays, but I much prefer having ``black`` than not. I use it in all my projects as it's fast and has broad and growing support across the Python ecosystem. I'm very much in favor of this APE. Thanks for the hard work in putting this together.

embray:
> I think a lot of these issues are surmountable via refactoring the code (minimizing hard coded arrays)

The idea of refactoring in order to satisfy a style checker is disturbing to me. Now if were talking static typing that's another thing.

But if you just mean wrt hard-coded data, I think it's better to move that into data files anyways (JSON, csv or even npz or something like that as appropriate), so +1 to that.

taldcroft:
> The idea of refactoring in order to satisfy a style checker is disturbing to me.

Recall this actual code from astropy:
```
            numer = (((((((((((((((AA[15]*Z + AA[14])*Z + AA[13])*Z + AA[12])*Z + AA[11])*Z +
                               AA[10])*Z + AA[9])*Z + AA[8])*Z + AA[7])*Z + AA[6])*Z +
                          AA[5])*Z + AA[4])*Z+AA[3])*Z + AA[2])*Z + AA[1])*Z + AA[0])
```
Black expands this out into a very long multiline expression, which may or may not suit your taste. But I think it is not disputed that the original (current) version is not ideal.  Discussion in the original astropy-dev thread on black found that there would be a straightforward way to refactor this expression programmatically that made it a lot more understandable.

The takeaway being that cases where black does poorly sometimes suggest a better way. Your suggestion of putting substantial data chunks into a files is another good example.

WilliamJamieson:
> The idea of refactoring in order to satisfy a style checker is disturbing to me. Now if were talking static typing that's another thing.

My comment, was more about the fact that when we apply `black` it tends to point out area's of our code that are less than ideal, such as what was pointed out here: https://github.com/astropy/astropy-APEs/pull/76#issuecomment-1249503215. So `black` has a tendency to point out areas that we could generally improve. Indeed, the original conversation in google groups discussed how that example could be improved in several different ways.

> But if you just mean wrt hard-coded data, I think it's better to move that into data files anyways (JSON, csv or even npz or something like that as appropriate), so +1 to that.

This is mostly what I was thinking. Especially, when it comes to some of our test cases. (Also, I would add ASDF as a possible method for storing this type of data, but that is my personal bias)

mwcraig:
Thank you for the productive discussion everyone -- the Coordination Committee has accepted this APE, which I will merge momentarily.

pllim:
Now that all is said and done, here are the lessons learned brought up in 2022-12-06 telecon. I am documenting it here if we ever use this as the model for the next big refactoring APE.

1. Assign a few overseers who can be present and make timely decisions that are consistent with the accepted APE; i.e., a governance structure for that effort. Without this, PR author was forced to make decisions on the spot on a case-by-case basis, causing the APE application to be inconsistent throughout.
2. APE should have had a "tips and tricks" section; e.g., on how to apply `black` better with trailing commas.

PR #79, opened 2022-10-13
APE 0: CoCo numbers update now that we know

### LINKED ISSUE #74: APE 0 should be updated regarding Coordination Committee membership
This is what APE 0 has to say about the number of Coordination Committee members: 
https://github.com/astropy/astropy-APEs/blob/79242a8a0fa37fb10260952ca393bb88d1bfeac0/APE0.rst#L52-L54
The number is now known to be 5, so the text needs to be updated accordingly. Related to that, the first Coordination Committee terms were written out for four members and should be updated:
https://github.com/astropy/astropy-APEs/blob/79242a8a0fa37fb10260952ca393bb88d1bfeac0/APE0.rst#L156-L158

pllim:
This is to fix #74 . cc @eerovaher

### DIFF
--- a/APE0.rst
+++ b/APE0.rst
@@ -7,7 +7,7 @@ date-created: 2020 July 31
 
 date-last-revised: 2022 October 13
 
-date-accepted:
+date-accepted: 2021 February 19
 
 type: Process
 
--- a/APE0.rst
+++ b/APE0.rst
@@ -5,9 +5,9 @@ author: The Astropy Governance Working Group; Tom Aldcroft, Matt Craig, Kelle Cr
 
 date-created: 2020 July 31
 
-date-last-revised: 2021 February 19
+date-last-revised: 2022 October 13
 
-date-accepted: 2021 February 19
+date-accepted:
 
 type: Process
 
@@ -49,8 +49,8 @@ The Coordination Committee
 
 Composition
 ^^^^^^^^^^^
-The Coordination Committee is a committee that is composed of four or five
-members (this number will be decided by vote at the same time as the first
+The Coordination Committee is a committee that is composed of five
+members (this number was decided by vote at the same time as the first
 Coordination Committee election). Unless it would
 be short of candidates otherwise, it should not have more than one member from
 any particular employer. The Ombudsperson (described in
@@ -157,7 +157,7 @@ In the case of a vacancy partway through a term, a by-election will be held.
 The term of the newly-elected member runs for the remainder of the term of
 the member they are replacing. 
 
-For the initial election of Coordination Committee members, two seats will have
+For the initial election of Coordination Committee members (four at the time instead of five), two seats will have
 terms of one year, one has a term of two years, and one has a term of three
 years to create a staggered set of replacements and provide continuity in the
 Coordination Committee. The candidate with the largest number of votes gets the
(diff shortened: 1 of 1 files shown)


APE0.rst
```diff
@@ -157,7 +157,7 @@ In the case of a vacancy partway through a term, a by-election will be held.
 The term of the newly-elected member runs for the remainder of the term of
 the member they are replacing. 
 
-For the initial election of Coordination Committee members, two seats will have
+For the initial election of Coordination Committee members (four at the time instead of five), two seats will have
 terms of one year, one has a term of two years, and one has a term of three
 years to create a staggered set of replacements and provide continuity in the
```
eerovaher, line 162: This paragraph determines the number of new members that will be elected to the CoCo in future elections, so it is important that it explicitly states the terms of all five initial members of the CoCo.

pllim: That I don't know how to fix. And to ask @astropy/coordinators to figure it out might cause conflict of interest. Hmm...

kelle: There's an email somewhere that explicitly states our different terms. Not sure where the "conflict" would be.

pllim: @kelle , do you still have that email? Would be nice to put that officially here, maybe?

hamogu: We did not spec out what would happen for 5 people on the coco, since the original plan was for 4. I think essentially Stuart decided ad hoc, that we do 2 for one year,  2 for two years, and one for three years

APE0.rst
```diff
@@ -157,8 +157,8 @@ In the case of a vacancy partway through a term, a by-election will be held.
 The term of the newly-elected member runs for the remainder of the term of
 the member they are replacing. 
 
-For the initial election of Coordination Committee members, two seats will have
-terms of one year, one has a term of two years, and one has a term of three
+For the initial election of Coordination Committee members (four at the time instead of five), two seats will have
+terms of one year, one has a term of two years (two here now that there are five), and one has a term of three
```
pllim, line 161: Probably can be worded better...

kelle:
https://community.openastronomy.org/t/results-of-the-2021-coco-election/144

> Kelle Cruz - Three Years
> Moritz Günther - Two Years
> Erik Tollerud - Two Years
> Adrian Price-Whelan - One Year
> Matt Craig - One Year

APE0.rst
eerovaher: I don't see a good reason for mentioning the terms for a 4-member CoCo at all, it just makes the text more confusing.

pllim: Removing a line from established APE might be controversial?

Feel free to suggest alternate wordings. Thanks!

eerovaher, line 161: I would find it useful if the section of the APE that describes the terms of the first CoCo would describe the terms of the actual first CoCo instead of some draft proposal:
```suggestion
For the initial election of Coordination Committee members, two seats will have
terms of one year, two have a term of two years, and one has a term of three
```
However, I am not a lawyer so I cannot claim with full confidence that such a change would be allowed.

pllim: I applied your suggestion. I am not a lawyer either. 😸

mwcraig:
LGTM

pllim:
Thanks!

PR #82, opened 2023-02-16
APE 21: APE on ending long term support releases + Updates to APE 2 to reflect new release cycle without LTS

astrofrog:
This is a draft and not ready for review - it follows the [related discussion on astropy-dev](https://groups.google.com/g/astropy-dev/c/QArGJfzPVeE). If you are interested in joining as a co-author please ping me!

- [x] Rename file to `APE_21.rst`

EDIT: Close #20

### DIFF
--- a/APE2.rst
+++ b/APE2.rst
@@ -85,8 +85,7 @@ additions are still acceptable.  The standard calendar for a release will be:
 
 This is not meant to be an exactly strict calendar, as the day of the week,
 existence of holidays, and degree of readiness of key features or critical bug
-fixes necessitate slop in the exact timing.  So more what you'd call guidelines
-than actual rules.
+fixes necessitate slop in the exact timing.
 
 Rapid Feature Releases
 ^^^^^^^^^^^^^^^^^^^^^^
@@ -104,7 +103,7 @@ release, a v6.1.0 release could occur at that time.  Once that release has
 completed, the release clock is "reset", so v6.2.0 would be released six months
 after v6.1.0, even though it is not yet 12 months after v6.0.0. The next two
 releases would then be v6.3.0 and v6.4.0, because v6.4.0 would only be 21 months after
-v6.0.0. v7.0 would then either come three months or six months later, depending on
+v6.0.0. v7.0.0 would then either come three months or six months later, depending on
 how much work is planned to be new in v7.0.0.
 
 Releases and backwards compatibility
--- a/APE2.rst
+++ b/APE2.rst
@@ -21,9 +21,9 @@ Abstract
 
 This APE describes the general plan for timing of releases of the core package
 and their version numbers. The basic premise follows the terminology of semantic
-versioning: *major* releases (e.g., 6.0.0) every two years, *minor* releases (e.g.,
-6.1.0) every six months in between, and *bugfix* releases (e.g., 6.1.2) as
-needed within the six months between minor releases.
+versioning: *major* releases (e.g., 6.0.0) every year (or longer), *minor*
+releases (e.g., 6.1.0) typically every six months in between, and *bugfix*
+releases (e.g., 6.1.2) as needed between minor releases.
 
 Detailed description
 --------------------
@@ -48,10 +48,11 @@ Release versioning and scheduling
 
 The release cycle for the core package should be as follows:
 
-* Major releases every two years. These are releases that can introduce breaking
+* Major releases every year (or longer). These are releases that can introduce breaking
   API changes, and should be numbered as x.0.0 (e.g., 6.0.0).
-* Minor releases every six months in between. These releases should minimize any
-  API changes and focus on adding new features.
+* Minor releases by default every six months in between major releases (or more
+  frequent if needed). These releases should minimize any API changes and focus
+  on adding new features.
 * Bugfix releases as needed between minor releases.
 
 As an example, starting from the 6.0.0 release, the version number schedule might be:
@@ -62,13 +63,29 @@ As an example, starting from the 6.0.0 release, the version number schedule migh
 * 6.1.0 (six months after 6.0.0)
 * 6.1.1
 * 6.1.2
+* 7.0.0 (six months after 6.1.0, one year after 6.0.0)
+
+In the above scenario, we would adopt the default 6-month release schedule, and
+introduce a new major version with breaking changes one year later. However, the
+release schedule could also look like the following:
+
+* 6.0.0
+* 6.0.1
+* 6.0.2
+* 6.1.0 (three months after 6.0.0)
+* 6.1.1
+* 6.1.2
 * 6.2.0 (six months after 6.1.0)
 * 6.2.1
-* 6.3.0 (six months after 6.2.0)
+* 6.3.0 (four months after 6.2.0)
 * 6.3.1
 * 6.3.2
 * 6.3.3
-* 7.0.0 (six months after 6.3.0, two years after 6.0.0)
+* 7.0.0 (five months after 6.3.0, 18 months after 6.0.0)
+
+In the above hypothetical scenario, the developers have decided to release minor
+releases more frequently than every six months, and a year after 6.0.0 there was
+no need for a major version, so this was delayed until 18 months after 6.0.0.
 
 Feature freeze/testing calendar
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -94,17 +111,16 @@ This plan is not meant to exclude the possibility of minor releases *more* frequ
 than every six months.  Six months should be considered both the maximum and the
 default timeline, but a *shorter* time between releases is allowable, if the
 community agrees it is worthwhile. When/if this happens, the subsequent major
-release should be timed such that is at least two years after the last major release, even
-though that may mean there are more than 3 releases between the major releases.
+release should be timed such that is at least one year after the last major release, even
+though that may mean there are more than one minor release between the major releases.
 
 For example, if a new feature is perceived by the Astropy developer community as
 being of great importance, and is completed only three months after the v6.0.0
 release, a v6.1.0 release could occur at that time.  Once that release has
 completed, the release clock is "reset", so v6.2.0 would be released six months
-after v6.1.0, even though it is not yet 12 months after v6.0.0. The next two
-releases would then be v6.3.0 and v6.4.0, because v6.4.0 would only be 21 months after
-v6.0.0. v7.0.0 would then either come three months or six months later, depending on
-how much work is planned to be new in v7.0.0.
+after v6.1.0, even though it is not yet 12 months after v6.0.0. The next release
+could then be v7.0.0 three or more months after v6.2.0, or could also be another
+minor release.
 
 Releases and backwards compatibility
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -114,13 +130,11 @@ The guidelines in relation to API changes and backward compatibility are as foll
 * API changes (i.e., modifications as opposed to additions) should only be
   carried out in major releases.
 * Any API changes should be preceded by deprecation warnings that should be
-  visible to users in at least one minor release cycle prior to the API change
+  visible to users for at least a year prior to the API change
   being carried out.
 * API changes should be documented, with clear guidance on what is changing,
   why the change is being made, and how to migrate existing code to the new behavior.
-* New deprecations should not be introduced in bugfix releases - this will
-  ensure that deprecations will therefore be emitted for at least 6 months
-  before a major release.
+* New deprecations should not be introduced in bugfix releases.
 * If developers wish to make an API change at a point in time where the next
   release is a major release, they should introduce the deprecation in the major
   release and carry out the change in the following major release.
--- a/APE2.rst
+++ b/APE2.rst
@@ -20,7 +20,7 @@ Abstract
 --------
 
 This APE describes the general plan for timing of releases of the core package
-and their version numbers. The basic premise follows the terminology of semantic
+and their version numbers. The basic premise follows semantic
 versioning: *major* releases (e.g., 6.0.0) every year (or longer), *minor*
 releases (e.g., 6.1.0) typically every six months in between, and *bugfix*
 releases (e.g., 6.1.2) as needed between minor releases.
--- a/APE2.rst
+++ b/APE2.rst
@@ -5,7 +5,7 @@ author: Erik Tollerud
 
 date-created: 2013 November 18
 
-date-last-revised: 2013 December 11
+date-last-revised: 2023 April 27
 
 date-accepted: 2013 December 10
 
@@ -13,74 +13,62 @@ type: Process
 
 status: Accepted
 
+revised-by: Thomas Robitaille, Pey Lian Lim, Simon Conseil - 2023 April 27 - Rewritten following the approval of APE 21 to no longer have Long Term Support (LTS) versions
+
 
 Abstract
 --------
 
-This APE describes the general plan for timing of releases and their version
-numbers.  The basic premise is a 6-month cycle for new feature releases, with
-releases designated "long-term support" (LTS) releases every two years.  Bugfix
-releases occur as needed, on both the current feature release and the LTS
-releases.
-
+This APE describes the general plan for timing of releases of the core package
+and their version numbers. The basic premise follows the terminology of semantic
+versioning: *major* releases (e.g. 6.0.0) every two years, *minor* releases (e.g.
+6.1.0) every six months in between, and *bugfix* releases (e.g. 6.1.2) as
+needed within the six months between minor releases.
 
 Detailed description
 --------------------
 
-The release plan through version 0.3 has been somewhat haphazard, with releases
-when a certain set of features are judged "ready", but without a clear sense of
-how long those features will take.  This has lead to concerns that new features
-are spending a long time in the development version because *other* features are
-not yet complete.  At the same time, some larger institutions and facilities
-have expressed concern that there is not a clearly-stated support cycle for
-Astropy versions.  That is, it is not clear if a given version will be supported
-via critical bug fixes if it is not the latest version.  This was voiced as a
+The release plan prior to v0.3 was somewhat haphazard, with releases carried out when a
+certain set of features were judged "ready", but without a clear sense of how
+long those features would take.  This led to concerns that new features were
+spending a long time in the development version because *other* features were
+not yet complete.  At the same time, some larger institutions and facilities had
+back then expressed concern that there was not a clearly-stated support cycle for
+Astropy versions. That is, it was not clear if a given version would be supported
+via critical bug fixes if it was not the latest version.  This was voiced as a
 concern because facilities will not always want to upgrade to the latest
-version, but do want to be sure major bugs will get addressed.
+version, but do want to be sure major bugs will get addressed. This led to the
+initial version of this APE following discussions at the 2013 Coordination
+Meeting. This APE is a living document and is meant to reflect the latest
+release guidelines (prior versions can be found in `Previous versions of this
+APE`_).
 
-Release scheduling and LTS
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+Release versioning and scheduling
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-These issues were discussed in the 2013 Coordination Meeting, and the following
-proposal represents a consensus from that meeting on the solution, with some
-modifications based on community feedback after the meeting.  Starting with
-version 0.4 , Astropy will have a feature release at least as often as every six
-months.  Releases between feature releases will only contain bugfixes, not new
-features. (See the "Releases and backwards compatibility" section below for more
-on what is considered a bug fix).
+The release cycle for the core package should be as follows:
 
-Additionally, some feature releases will be designated "long-term support" (LTS)
-releases, and these will continue to receive bugfixes for at least two years
-(i.e., typically 3 regular feature releases will follow an LTS release).  This
-addresses the first issue by applying a time-based release cycle, meaning no
-feature will need to wait more than six months before reaching users.
-Simultaneously, it addresses the second concern by providing a guarantee to
-facilities that if they use an LTS, they can count on releases fixing critical
-bugs for at least two years.  Finally, the LTS releases also act as a convenient
-time to submit new Astropy journal papers.
+* Major releases every two years. These are releases that can introduce breaking
+  API changes, and should be numbered as x.0.0 (e.g. 6.0.0).
+* Minor releases every six months in between. These releases should minimize any
+  API changes and focus on adding new features.
+* Bugfix releases as needed between minor releases.
 
-Version numbering
-^^^^^^^^^^^^^^^^^
+As an example, starting from the 6.0.0 release, the version number schedule might be:
 
-To match this release plan, Astropy will adopt a version numbering scheme of the
-form ``x.y.z``, where ``z`` is advanced on a bugfix release, ``y`` is advanced
-for a feature release that is *not* an LTS release, and ``x`` is advanced on an
-LTS (which also resets ``y`` to 0).  So starting from the first LTS release, the
-version number schedule might be::
-
-* 1.0.0 (LTS release)
-* 1.0.1
-* 1.0.2
-* 1.1.0 (six months after 1.0.0)
-* 1.1.1
-* 1.0.3
-* 1.1.2
-* 1.2.0 (six months after 1.1.0)
-* 1.2.1
-* 1.3.0 (six months after 1.2.0)
-* 1.0.4
-* 1.3.1
-* 2.0.0 (LTS release, six months after 1.3.0)
+* 6.0.0
+* 6.0.1
+* 6.0.2
+* 6.1.0 (six months after 6.0.0)
+* 6.1.1
+* 6.1.2
+* 6.2.0 (six months after 6.1.0)
+* 6.2.1
+* 6.3.0 (six months after 6.2.0)
+* 6.3.1
+* 6.3.2
+* 6.3.3
+* 7.0.0 (six months after 6.3.0, two years after 6.0.0)
 
 Feature freeze/testing calendar
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -103,42 +91,92 @@ than actual rules.
 Rapid Feature Releases
 ^^^^^^^^^^^^^^^^^^^^^^
 
-This plan is not meant to exclude the possibility of releases *more* frequent
+This plan is not meant to exclude the possibility of minor releases *more* frequent
 than every six months.  Six months should be considered both the maximum and the
 default timeline, but a *shorter* time between releases is allowable, if the
-community agrees it is worthwhile. When/if this happens, the subsequent LTS
-release should be timed such that is at least two years after the last LTS, even
-though that may mean there are more than 3 releases between the LTSs.
+community agrees it is worthwhile. When/if this happens, the subsequent major
+release should be timed such that is at least two years after the last major release, even
+though that may mean there are more than 3 releases between the major releases.
 
 For example, if a new feature is perceived by the Astropy developer community as
-being of great importance, and is completed only three months after the v1.0
-release, a v1.1 release could occur at that time.  Once that release has
-completed, the release clock is "reset", so v1.2 would be released six months
-after v1.1, even though it is not yet 12 months after v1.0. The next two
-releases would then be v1.3 and v1.4, because v1.4 would only be 21 months after
-v1.0. v2.0 would then either come three months or six months later, depending on
-how much work is planned to be new in v2.0.
+being of great importance, and is completed only three months after the v6.0.0
+release, a v6.1.0 release could occur at that time.  Once that release has
+completed, the release clock is "reset", so v6.2.0 would be released six months
+after v6.1.0, even though it is not yet 12 months after v6.0.0. The next two
+releases would then be v6.3.0 and v6.4.0, because v6.4.0 would only be 21 months after
+v6.0.0. v7.0 would then either come three months or six months later, depending on
+how much work is planned to be new in v7.0.0.
 
 Releases and backwards compatibility
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-As defined here, a "bugfix release" must not break backwards compatibility as
-defined by the public API, and it also should not introduce new major features.
-The exception to this is cases where the API has an actual mistake that needs to
-be fixed, such as a function argument that is missing but was clearly intended
-to be there. Further, a bugfix release *can* include documentation improvements,
-and in the future there may be some sort of "sandbox" package to act as a
-preview or testbed for new additions to come in the next feature release.
+The guidelines in relation to API changes and backward compatibility are as follows:
 
-Feature releases do *not* guarantee backwards compatibility with previous
-releases.  However, all backwards incompatible changes should always be
-mentioned in the release notes for a new version.  Furthermore, when possible, a
-backwards incompatible change should generate an `AstropyDeprecationWarning` for
-at least one feature release version before actually making the change
-permanent, to allow codes to transition to the new interface. A future APE may
-add to or change this with a more concrete policy on when and for how long
-something should be deprecated.
+* API changes (i.e., modifications as opposed to additions) should only be carried out in major releases.
+* Any API changes should be preceded by deprecation warnings that should be
+  visible to users in at least one minor release cycle prior to the API change
+  being carried out.
+* New deprecations should not be introduced in bugfix releases - this will
+  ensure that deprecations will therefore be emitted for at least 6 months
+  before a major release.
+* If developers wish to make an API change at a point in time where the next
+  release is a major release, they should introduce the deprecation in the major
+  release and carry out the change in the following major release.
+* The following will not be considered to be part of the API and can therefore
+  be changed in bugfix and minor releases:
 
+    * Changes to warning messages
+    * Changes to exception messages
+    * Changes to the ``__repr__`` or ``__repr_*__`` of objects
+
+  However, exception and warning *types* will be considered to be part of the
+  API.
+
+As with all guidelines, there are exceptions where we can deviate from these guidelines:
+
+* API changes can be carried out in major releases without deprecation if it is
+  not possible to have a deprecation phase due to the nature of the change.
+* Changes breaking or changing the behavior of code may be needed in order to
+  fix bugs. Such changes can be made in bugfix or minor releases and do not need
+  to be considered API changes.
+* Changes to exception or warning types may occasionally be done in minor
+  releases.
+
+These exceptions will require judgment calls on the part of maintainers, but any
+such changes should be minimized as much as possible and clearly communicated to
+users (see `Communication with users`_).
+
+Documentation and performance improvements can be made in bugfix releases as long
+as they do not involve any changes or additions to the API.
+
+Communication with users
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is imperative that we communicate effectively with users so that they understand
+what to expect from different releases:
+
+* We should include documentation for the core package that describes the
+  release policy above, namely that in general minor releases will not
+  break/remove functionality, but that major releases might, and also explaining
+  what is considered API.
+
+* We should ensure that any changes in releases which deviate from the
+  guidelines above are clearly communicated to users - for example any API/breaking
+  changes in minor releases should be included in the "What's New" page for the
+  minor release.
+
+* Deprecations should always be mentioned in the changelog, as well as any
+  removal of deprecated code/API changes. We do not specify how exactly this
+  should be done, but note that e.g. Numpy release notes have separate sections
+  entitled *Deprecations*, *Expired Deprecations* and *Compatibility Notes* in
+  their release notes (see `here
+  <https://numpy.org/doc/stable/release/1.24.0-notes.html>`_ for an example)
+  which is clearer than having a single *API changes* section.
+
+* Maintainers should be encouraged to advertise pull requests with API changes
+  that are likely to affect users and developers of other packages to the
+  developer mailing list, and ensuring that the pull request has adequate time
+  (at least two weeks) to be reviewed properly.
 
 Branches and pull requests
 --------------------------
@@ -149,22 +187,8 @@ N/A
 Implementation
 --------------
 
-If this APE is accepted, a few organizational steps will be required:
-
-* The currently under-development version "v0.4" will be the first following
-  the release cycle.  Because v0.3 was released On the last Monday in Nov 2013,
-  the feature freeze and beta for v0.4 will be scheduled for the last Monday in
-  April 2014, With a release at the end of May 2014.  The milestone on GitHub
-  will thus need to be updated to match this.
-* The first LTS release (v1.0) will then come six months after v0.4, so an
-  appropriate milestone will need to be added to GitHub.  Following that,
-  milestones should be added as needed to reflect the versioning scheme
-  described in this APE.
-* The documentation should be updated to mention the existence of LTS releases
-  and make it clear that they come with a promise of two years of bug fixes (as
-  needed). It should also add a link to this APE so that those interested in the
-  numbering scheme can look here for details.
-
+Whenever this APE is updated, the core package documentation should be updated
+to reflect the latest guidelines described above.
 
 Backward compatibility
 ----------------------
@@ -175,20 +199,7 @@ N/A
 Alternatives
 ------------
 
-An alternative version numbering scheme mentioned on astropy-dev was to continue
-the current scheme of ``0.x.y``, and simply append `-LTS` for LTS releases. The
-disadvantage of this approach is that the first version number is then rendered
-essentially meaningless.  That is, given its nature as a research library, it is
-unlikely to ever be "feature-complete", and hence there is no clear break point
-for a 1.0 version.  The scheme proposed here simply assigns a different meaning
-to the major version as "number of LTSs since the start of the project".
-
-Another alternative is to use `semantic versioning <http://semver.org/>`_. The
-disadvantage of that approach for Astropy is that it is very likely that *all*
-versions in the foreseeable future will break backwards compatibility to some
-extent.  In semantic versioning, this would mean all versions would need to be
-of the form ``x.0.y``, rendering the minor version number meaningless.
-
+N/A
 
 Decision rationale
 ------------------
@@ -200,3 +211,8 @@ original proposal was the addition of the idea that *less than* six month
 releases were acceptable. The above content reflects those suggestions, and the
 APE was accepted 12/10/13, as there were no significant objections from the
 community.
+
+Previous versions of this APE
+-----------------------------
+
+* 2013-12-10 [`DOI <https://doi.org/10.5281/zenodo.1043887>`_] [`GitHub <https://github.com/astropy/astropy-APEs/blob/48f949c05efa4f07ed8915eccdb0cd139c57b6f6/APE2.rst>`_]
--- a/APE2.rst
+++ b/APE2.rst
@@ -5,7 +5,7 @@ author: Erik Tollerud
 
 date-created: 2013 November 18
 
-date-last-revised: 2023 April 27
+date-last-revised: 2023 May 15
 
 date-accepted: 2013 December 10
 
@@ -13,7 +13,7 @@ type: Process
 
 status: Accepted
 
-revised-by: Thomas Robitaille, Pey Lian Lim, Simon Conseil - 2023 April 27 - Rewritten following the approval of APE 21 to no longer have Long Term Support (LTS) versions
+revised-by: Thomas Robitaille, Pey Lian Lim, Simon Conseil - 2023 May 15 - Rewritten following the approval of APE 21 to no longer have Long Term Support (LTS) versions
 
 
 Abstract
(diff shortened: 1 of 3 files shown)


APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
+Ending Long Term Support Releases
+---------------------------------
+
+author: Thomas Robitaille
```
pllim, line 4: cc @saimn
```suggestion
author: Thomas Robitaille, Pey Lian Lim, Simon Conseil
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+
+date-created: 2023 February 16 <replace with the date you submit the APE>
```
pllim, line 6:
```suggestion
date-created: 2023 February 16
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+
+date-last-revised: 2023 February 16 <keep this up to date anytime something changes>
```
pllim, line 8:
```suggestion
date-last-revised: 2023 February 20 <keep this up to date anytime something changes>
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+
+
+Detailed description
+--------------------
+
+`APE 2: Astropy Release Cycle and Version Numbering <https://doi.org/10.5281/zenodo.1043887>`_ describes the release cycle
+for the core astropy package, which consists of major releases every 6 months that
+add new features (e.g., v5.2), and releases in between this that are limited to fixing bugs
+or updating documentation (e.g., v5.2.1). Every two years, major releases are designated
+as being 'long term support' (e.g., v5.0) and we have committed to releasing bugfix releases
```
pllim, line 31: The abbreviation is already introduced in Abstract above.
```suggestion
as being LTS (e.g., v5.0) and we have committed to releasing bugfix releases
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+for this version (e.g., v5.0.3) for up to two years.
+
+At the start of the project, having long term support releases was critical for acceptance by
```
pllim, line 34:
```suggestion
At the start of the project, having LTS releases was critical for acceptance by
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+large organizations/projects for which stability is paramount, as the core package went through
+significant API changes from version to version due to the youth of the project and long term
+support releases were a way to shield against these changes, or at least reduce disruption to
```
pllim, line 37:
```suggestion
significant API changes from version to version due to the youth of the project and LTS
releases were a way to shield against these changes, or at least reduce disruption to
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+
+* Even once fixes are backported to the LTS branch, carrying out releases is not
+  fully automated and requires manual steps, so having to do
+  bugfix releases for both the latest release branch and LTS branch puts an extra
+  burden on the core package release team.
+
+In addition, it is not clear that LTS releases are needed anymore - now that the
+core package is over ten years old, it has reached a much higher level of
+stability. No other main package in the scientific Python ecosystem has long
+term support releases, and there is no longer a strong argument for having this
```
pllim, line 56:
```suggestion
stability. No other main package in the scientific Python ecosystem has LTS
releases, and there is no longer a strong argument for having this
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+be every two years.
+
+LTS releases are however not free:
+
+* While we use infrastructure to automate backporting fixes to release branches,
+  in practice the LTS branch diverges more and more from the main development
+  branch and automated backports are more likely to fail as time goes on,
+  requiring human effort to manually do the backports, which can sometimes be
+  non-trivial.
+
```
pllim, line 47: No blank line to be consistent with the next set of bullet points.
```suggestion
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+In addition, it is not clear that LTS releases are needed anymore - now that the
+core package is over ten years old, it has reached a much higher level of
+stability. No other main package in the scientific Python ecosystem has long
+term support releases, and there is no longer a strong argument for having this
+in astropy either.
+
+This APE proposes the following:
+
+* No longer designate any releases as LTS
```
pllim, line 61:
```suggestion
* No longer designate any releases as LTS.
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+* Keep the 6-month release cycle and bump releases to x.0 every two years
```
pllim, line 62:
```suggestion
* Keep the 6-month release cycle and bump releases to x.0 every two years.
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+* Refer to the x.0 releases every two years (e.g. v6.0) as 'major'
```
pllim, line 63:
```suggestion
* Refer to the x.0 releases every two years (e.g., v6.0) as 'major'
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+  releases, other releases every 6 months (e.g. v6.1) as 'minor', and bug fixing
```
pllim, line 64:
```suggestion
  releases, other releases every 6 months (e.g., v6.1) as 'minor', and bug fixing
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+  releases (e.g. v6.1.2) as 'bugfix releases'
```
pllim, line 65:
```suggestion
  releases (e.g., v6.1.2) as 'bugfix releases'.
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+* Minimize as much as possible any changes that break API or change results in
+  minor releases, but being pragmatic and accepting such changes if they would
+  not affect many users
```
pllim, line 68:
```suggestion
  not affect many users.
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+in astropy either.
+
+This APE proposes the following:
+
+* No longer designate any releases as LTS
+* Keep the 6-month release cycle and bump releases to x.0 every two years
+* Refer to the x.0 releases every two years (e.g. v6.0) as 'major'
+  releases, other releases every 6 months (e.g. v6.1) as 'minor', and bug fixing
+  releases (e.g. v6.1.2) as 'bugfix releases'
+* Minimize as much as possible any changes that break API or change results in
```
pllim, line 66: This requires us to re-educate all the maintainers with write access to the repo. We wouldn't want inconsistency in how this is decided based on who happened to merge the PRs. Should this APE also lay out how these new rules should be followed?

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+  go through with the breaking changes in major releases (either the next major
+  release following when a warning was added, or subsequent ones if the changes
+  are deemed to be high impact enough to wait longer).
+
+With this in place, we would effectively be striving to maximize stability over
+two year periods but users would see deprecation warnings to prepare them for
+changes. This is in contrast to the current LTS system where users might not see
+a deprecation warning until the next LTS release as deprecations are not typically
+backported.
+
```
pllim, line 79:
```suggestion
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+Branches and pull requests
+--------------------------
+
+N/A
+
+Implementation
+--------------
+
+This APE proposes to continue supporting the 5.0.x LTS series but to not
+designate 6.0 as LTS.
```
pllim, line 90:
```suggestion
This APE proposes to continue supporting the v5.0.x LTS series but to not
designate v6.0 as LTS.
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+
```
pllim, line 91:
```suggestion
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Backward compatibility
+----------------------
+
+N/A
+
```
pllim, line 97:
```suggestion
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Alternatives
+------------
+
+If there were any alternative solutions to solving the same problem, they should
+be discussed here, along with a justification for the chosen approach.
+
```
pllim, line 104:
```suggestion
```

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+Backward compatibility
+----------------------
+
+N/A
+
+
+Alternatives
+------------
+
+If there were any alternative solutions to solving the same problem, they should
```
pllim, line 102: Is this just a placeholder text?

APE_LTS.rst
```diff
@@ -0,0 +1,109 @@
...
+be discussed here, along with a justification for the chosen approach.
+
+
+Decision rationale
+------------------
+
+<To be filled in by the coordinating committee when the APE is accepted or rejected>
```
pllim, line 109: Being the coauthor and current CoCo member, I would recuse myself from CoCo discussions regarding this APE.

pllim (commented):
I think we have to rename this file to `APE_21.rst`. Otherwise, the file listing will hurt my head going forward.

I think this is going to supersede #20 , so the proposed addenda there by @mhvk needs to be addressed here as well, if not already.

Why the preference on single-quote over double-quote in the text?

mhvk:
Had a quick look and liked this! Agreed that it will be important to be very clear about minor and major. After reading this, and my original proposed addendum to #20, I think the proposal is:
- Minor releases mostly should add things and can, if needed, deprecate things. Generally, existing code should continue to work (with possible filtering of warnings), so little if anything should be removed..
-  Major releases are the same but can also remove deprecated features. So, existing code may need to be adjusted (but would have had ample warning).

Is this correct? Perhaps good to give examples. E.g., "existing code should continue to work" probably should not include that every `repr` remains the same, or every number exactly identical. But it would include not changing API.

astrofrog:
@mhvk - I agree with your assessment, I think we basically have to decide on guidelines for those exceptions of things that can change in a minor release that might be considered as 'breaking'. Things that are borderline would include:

1. Changes in warning messages
2. Changes in exception messages or exception types
3. Removing functionality that was not really intended to ever be public and is probably not used (for example some random function deep in astropy.utils that is not technically private)
4. Fixing recent API 'bugs'/mistakes that would be easier fixed sooner rather than later if it can't be fixed via a regular deprecation

Anything else?

I think something related to this APE that we don't have to necessarily figure out here but that we could mention briefly is that if we are going to have rules like this we should also be clearer on what is public API - anything tab-discoverable? Anything in the API docs in docs.astropy.org? (those two don't necessarily match currently though I think I saw @nstarman mention something about public API in another PR). Defining what is public API could almost be a separate APE.

APE_LTS.rst
astrofrog: Agree - see my latest comment in the main discussion. I think we'll need to add a section about this.

APE_LTS.rst
astrofrog: Yes

mhvk:
I think all four are fine in minor releases -- certainly exception and warning messages (I think it is fine to break people's tests if those are so detailed they check messages, just not to break actual code!).

But very good point about public API. The most narrow (and probably best) definition is that it is only includes thinkgs exposed at the top level of astropy's submodules (with a few exceptions in `utils` like `iers` and `masked`). But you're right, this is probably another APE, with the main goal to actually document our current behaviour!

pllim:
How many will cry if I change `astropy.utils` to `astropy._utils` in v6.0? 😝

pllim:
Why are we avoiding double quotes? I thought English prefers double quotes and you only use single quotes if you have to quote inside text that is already in double quotes?

astrofrog:
Happy to change to double quotes! Can we just run black on the APE? :stuck_out_tongue: Feel free to push a commit to fix this if you have time :)

pllim:
@astrofrog , I took some liberties in 22c9073ebf4cbfe6cef21cb189fa5f4631eba2f1 , feel free to revert the changes you don't agree with. Thanks!

nstarman:
> I think something related to this APE that we don't have to necessarily figure out here but that we could mention briefly is that if we are going to have rules like this we should also be clearer on what is public API - anything tab-discoverable? Anything in the API docs in docs.astropy.org? (those two don't necessarily match currently though I think I saw @nstarman mention something about public API in another PR). Defining what is public API could almost be a separate APE.

I would propose we adopt the PEP8 standard (https://peps.python.org/pep-0008/#public-and-internal-interfaces) and the typing guidelines (https://github.com/python/typing/blob/master/docs/source/libraries.rst#how-much-of-my-library-needs-types). The latter offers some small refinements to the former, namely to ensure that analyzers, static and dynamic, can easily understand what is public and help users with better tab autocomplete / code snippets / generative code, etc.

mhvk:
Hmm, those texts are helpful but not quite conclusive -- or at least I can read it in ways that I'm fairly sure are different from other's! Anyway, to me it sounds like if in a submodule's `__init__` we do not import the actual subsubmodules (say, in `coordinates.__init__`, we import from `representations`, but not `representations` itself), then we basically are OK (especially if we also define an `__init__.__all__`). And that in `units`, `.si` is a public module since we import it in `.__init__` (which may not be what we want, so we do need to define `__all__` there).

I guess the only part to then clarify is that a subsubmodule should only define things in `__all__` that are actually imported in the submodule proper.

p.s. Probably should move this to a different thread; sorry!

pllim:
@astrofrog , is your plan to discuss this APE at the Coordination Meeting?

astrofrog:
Yes but only if we can't agree on it sooner!

saimn:
I like the proposal for several reasons:
- the meaning of version numbers should be easier to understand for users (in contrast to LTS where people may hesitate on which version to use)
- users are more likely to update to the latest minor version since it should not break their code
- for people that were using LTS, they can update to the latest minor version and get new features more rapidly

The difficult point is how to define an API break. Tom gave some examples in https://github.com/astropy/astropy-APEs/pull/82#issuecomment-1438161547, are changes to exception or warning messages API breaks ? IMO It's hard to define strict rules here, it depends on the impact of a change, if the code that is changed is widely used or not etc. So when reviewing PRs maintainers will need to think if something is likely to break other people's code. 

There are cases where breaking the API is needed to fix a bug, depending on the impact some may be fixable in minor versions, other may have to wait for major versions with a deprecation path. I like pandas' note about this: 
https://pandas.pydata.org/docs/development/policies.html#policies-version
> pandas will sometimes make behavior changing bug fixes, as part of minor or patch releases. Whether or not a change is a bug fix or an API-breaking change is a judgement call

(And I like the way they have written their policy, it's simple so easily understandable by users)

Also we probably need a way to track deprecations so we know what changes need to be done in major versions ?
And I think the discussion about public API is not relevant here, it's an interesting one but it should be discussed somewhere else. Here the focus is on how to handle deprecations, and how to provide to users new versions with some guarantees of stability.

APE21.rst
```diff
@@ -0,0 +1,105 @@
...
+In addition, it is not clear that LTS releases are needed anymore - now that the
+core package is over ten years old, it has reached a much higher level of
+stability. No other main package in the scientific Python ecosystem has LTS
+releases, and there is no longer a strong argument for having this
+in astropy either.
+
+This APE proposes the following:
+
+* No longer designate any releases as LTS.
+* Keep the 6-month release cycle and bump releases to x.0 every 2 years.
```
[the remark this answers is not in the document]
pllim: But this also means that if someone wants a new thing that will totally break API, they have to wait 2 years? Do we not merge that PR for 2 years?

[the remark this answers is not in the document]
saimn: Yeah indeed. This case should hopefully not happen too often, but we could imagine releasing a major version only 1/1.5 year after the previous one on exceptional circumstances ? But in general if there is a need to totally break the API there should be a long enough period of time before that with deprecation warnings, so the 2 years cycle seem realistic.

mhvk:
Agreed that the pandas wording is very nice & clear. I suggest we use it! I think the main trick is in defining what API is. In my opinion, how something is typeset (`repr`) or worded is *not* included in the API (for a minor release), so the wording of an error message is not, but the structure/type/behaviour of an object or the exception type is part of the API.

APE21.rst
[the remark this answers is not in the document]
taldcroft: It's worth noting that with a fixed 2-year cycle, some API changes might wait 1.9 years, and others might be merged 1 week before the deadline.

There would tend to be a bias (maybe strong?) toward the major version feature freeze deadline, because folks might have an idea and want to get it in knowing that it will be another 2 years otherwise. Something to consider!

WilliamJamieson:
> The only way out of this I can think of is to have some policy for advertising API changes and getting a wider review. For instance, if a PR has a change that has the potential to break application code or break other package tests, then it could be brought to the attention of `astropy-dev`. E.g. an email with a subject like `Astropy PR #yyy API change for review`, with some text describing the potential impact. Allow a minimum 2-week period for review prior to accepting the PR. That extra layer of review would also serve as a reminder to developers that API changes should be avoided when possible.

On top of this I think we should explicitly encourage maintainers of downstream packages to run their tests against the development version of `astropy` regularly (with some suggestion of what regularly means).

mhvk:
Agreed that we should encourage people to test against `astropy-dev`. 

I do want to push a bit more to keep the freedom to change formats or error messages in minor releases. If people rely on details like that (presumably mostly in doctests), then I think it is fair to ask them to adjust, just like we adjust when numpy changes its output format (and, yes, sometimes that is painful, but definitely OK for a minor release).

On error and warning messages specifically, in astropy we check those internally just to ensure we actually raise the right error, but really why would anyone rely on those outside of astropy? I don't think we have any place where we specifically look for the text of an external error or warning. And it does seem one should be free to correct or clarify exception texts, certainly in a minor release, arguably even in a bug-fix release.

Anyway, I really hope we can keep things more or less like they are for minor releases!

WilliamJamieson:
> On error and warning messages specifically, in astropy we check those internally just to ensure we actually raise the right error, but really why would anyone rely on those outside of astropy? I don't think we have any place where we specifically look for the text of an external error or warning. And it does seem one should be free to correct or clarify exception texts, certainly in a minor release, arguably even in a bug-fix release.

Hypothetically, someone could be using error messages to distinguish between errors they can handle and those they cannot. Although in that case we really ought to be using a custom error class to make it easier on those use cases (they should report that behavior to us anyways).

APE21.rst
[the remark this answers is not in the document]
saimn: Yes, that's why I think we should keep some flexibility in the 2 years cycle. 
But also before a breaking change appears in a major release, there should be deprecations in a previous minor release. So major release are not the time to merge anything just because we can break things in this release, we should plan beforehand what major changes are waiting, if deprecations have been released, for how long, etc.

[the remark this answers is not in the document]
mhvk: Agree with @taldcroft here. The most annoying numpy changes we have had to deal with are things that are merged just before the release and hence we didn't catch them in `-dev`.

Would it be an idea to freeze further ahead of a major release? With perhaps more advertising to test the corresponding -rc?

Alternatively (or additionally), maybe the way to do this is to be more flexible about when major releases happen, and decide on whether the next release will be major ~mid-way between releases, based on what is already ready there and what is being planned (with perhaps an astropy-dev mailing to ask what people are planning).

Part of this might also be to follow pandas' example and really make sure API changes are explained and work-arounds given.

One practical question: how is it going to work with major items? Are they just going to sit on as PR requests for ~1 year? Or is there a separate branch? As long as new features are OK for minor releases, I don't think this is a super-big issue, but probably should be discussed in the document!

mhvk:
Agreed that for that case we should have different error types (like we introduced `UnitTypeError` and `UnitConversionError` at some point -- though I think that was mostly for internal use).

One conclusion from that is that it is good to strongly encourage people to give feedback when things break!

APE21.rst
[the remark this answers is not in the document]
mhvk: Should have read @saimn's message - agreed that if the cycle is flexible and we insist (as we should) that breaking changes have deprecation periods, that solves most of the problems.

taldcroft:
> On error and warning messages specifically, in astropy we check those internally just to ensure we actually raise the right error, but really why would anyone rely on those outside of astropy?

The problem is that astropy very commonly uses generic exception types like `ValueError` and `TypeError`, so that the only way to catch a specific error is by using the exception message text.

We certainly could embark on a mission to improve that with custom error classes, but that is a different story.

taldcroft:
> Anyway, I really hope we can keep things more or less like they are for minor releases!

I agree 100% (as a culprit for some of the changes in question). I just think we need to ensure that such changes are appropriately *communicated* to stakeholders (putting on a manager's hat).

APE21.rst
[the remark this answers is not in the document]
saimn: > One practical question: how is it going to work with major items? Are they just going to sit on as PR requests for ~1 year? Or is there a separate branch? As long as new features are OK for minor releases, I don't think this is a super-big issue, but probably should be discussed in the document!

Yes new features are OK for minor releases. It's just the ones that break the API that should wait, with deprecation added in between. And that should be a minority of the PRs hopefully, so I guess we can just milestone them for the next major release and rebase when we start the major release cycle. Another way would be to have a "future" branch but if there are conflicts it will be more painful to resolve them compared to doing that for each PR.

astrofrog:
In terms of communication, what if we required any API change in minor releases to be mentioned in the What's new? (For major releases we could just link to the API changes section in the changelog)

saimn:
Yes I think we need to revisit the presentation of those changes in the changelog/whatsnew.
E.g. for major releases Numpy has several sections: "Deprecations", "Expired deprecations", "Compatibility notes" (and "Performance improvements", cf another discussion on where to put those):
https://numpy.org/doc/stable/release/1.24.0-notes.html
Pandas has "Backwards incompatible API changes", "Deprecations" (and "Performance improvements"):
https://pandas.pydata.org/docs/whatsnew/v1.5.0.html

astrofrog:
Nice examples! I'll try and work something into this APE.

astrofrog:
@pllim @saimn @taldcroft @mhvk - I've updated the APE to try and add some text that encapsulates some of the points above. Would you be able to take a look at it? Feel free to leave specific suggestions if you see text that should be added/modified/removed.

APE21.rst
```diff
@@ -0,0 +1,170 @@
...
+  and carry out the change in the following major release.
+
+As with all guidelines, there are exceptions where we can deviate from these guidelines:
+
+* API changes can be carried out in major releases without deprecation if it is
+  not possible to have a deprecation phase due to the nature of the change.
+* Changes breaking or changing the behavior of code may be needed in order to
+  fix bugs. Such changes can be made in bugfix or minor releases and do not need
+  to be considered API changes.
+* Changes to exception types or warning/error messages may occasionally be
```
mhvk, line 110: I think this is OK for exception types but disagree with the need to limit changes to warning/error messages. I thought the consensus for this was that if people in non-test code check messages, they should contact us so we can have a proper exception subtype.

APE21.rst
```diff
@@ -0,0 +1,170 @@
...
+what to expect from different releases:
+
+* We should include documentation for the core package that describes the
+  release policy above, namely that in general minor releases will not
+  break/remove functionality, but that major releases might.
+
+* We should ensure that any changes in releases which deviate from the
+  guidelines above are clearly communicated to users - for example any API/breaking
+  changes in minor releases should be included in the "What's New" page for the
+  minor release.
```
mhvk, line 130: Yes to this, but this to me makes even more clear that we really don't want messages like "the exception text for so-and-so has been reworded to ..." (nor should we start adding that to the changelog). This has to be *useful* information to people who are worried about results changing.

mhvk (commented):
Overall, reads very well, but I would really like to avoid including in API changes modifications to warnings/exception text and, e.g., minor changes to `repr` formatting. Indeed, while we at least can ensure that we don't introduce them in bugfix releases, we do not have full control. E.g., when numpy rationalized how arrays were printed and with how many significant figures, lots of our doctests had to be changed, and the ~output~ `repr` of virtually all classes changed too. 

Anyway, I guess my bottom line would be that API means input and output with the same meaning, not necessarily identical bits.

APE21.rst
astrofrog: Ah right I see what you mean - yes this is fine with me. I've updated the APE.

APE21.rst
astrofrog: Sounds good!

astrofrog:
@mhvk - I've updated the APE to now explicitly state that messages and __repr__s are not considered API. Let me know if you are happy with the text now!

APE21.rst
```diff
@@ -0,0 +1,187 @@
...
+  visible to users in at least one minor release cycle prior to the API change
+  being carried out.
+* New deprecations should not be introduced in bugfix releases - this will
+  ensure that deprecations will therefore be emitted for at least 6 months
+  before a major release.
+* If developers wish to make an API change at a point in time where the next
+  release is a major release, they should introduce the deprecation in the major
+  release and carry out the change in the following major release.
+* The following will not be considered to be part of the API and can therefore
+  be changed in bugfix and minor releases:
```
mhvk, line 103: Maybe slightly tighter: "... changed in minor releases (and spelling corrections, etc., also in bugfix releases)."

APE21.rst
```diff
@@ -0,0 +1,187 @@
...
+  API.
+
+As with all guidelines, there are exceptions where we can deviate from these guidelines:
+
+* API changes can be carried out in major releases without deprecation if it is
+  not possible to have a deprecation phase due to the nature of the change.
+* Changes breaking or changing the behavior of code may be needed in order to
+  fix bugs. Such changes can be made in bugfix or minor releases and do not need
+  to be considered API changes.
+* Changes to exception or warning types may occasionally be done in minor
```
mhvk, line 119: Maybe generalize slightly, to make it more consistent with what you had on top: "Small-impact changes such as modifications in exception or warning types..."

mhvk (commented):
Looks good to me, with two small suggestions.

One slightly larger point: right now, it is by omission allowed to have new features in a minor release. Should this be made explicit?

Also, perhaps an implementation detail, but thinking about new features, should the "What's new" for a major release be relative to that last major release?

taldcroft:
> should the "What's new" for a major release be relative to that last major release?

In terms of features and bug fixes, I think that most people would expect it to be relative the last minor release. Since the major release no longer corresponds to an LTS there is no particular reason that many people would still be using it.

APE21.rst
taldcroft: At this point (10 years in) I would argue that changing exception or warning type base classes should be very rare. What might be more common and even encouraged would be moving to more specific exceptions/warnings that are derived from the original base class. I.e. `class IoAsciiSomeSpecificError(ValueError)`. This seems like a non-breaking change that should be encouraged, but changing `ValueError` to `TypeError` should be strongly discouraged.

APE21.rst
taldcroft: I disagree for reasons stated above. In practice this doesn't (or shouldn't) happen so frequently that it will generate too much noise, and I specifically think that there should be a barrier to making changes like this.

Especially the changelog should be complete and any exception/warning message format change should be noted as an API change. There should not be any hidden changes in releases, esp. ones that *might* break downstream code/tests.

The What's New perhaps doesn't need this however.

taldcroft:
Now a big-picture comment that I don't see here yet. This is really superceding a big component of APE-2 which is the fundamental document on Astropy Release Cycle and Version Numbering. The second sentence in the abstract there mentions the LTS and will now be wrong.

To that end I propose integrating much of the policy-oriented content into APE-2 and following on with a substantial update to APE-2. That would be instead of an update to APE-2 which just points to APE-21.  When people go to the Astropy APEs and open any of these documents they should reflect the current policy.

During the governance discussions we had a vision of policy documents, but I'm not sure that has really become reality. One test of that is that I don't actually know where I would find those policy documents. On the other hand, the APE's have continued as an effective and visible place for high level policy like release cycles, numbering, code formatting etc. *So I think we should fully embrace the idea of APEs as living documents that get updated as needed to reflect the current policy.*

An update to APE-2 to reflect this change would probably want to include a brief section on the prior policy so that people looking at APE-2 have an understanding of historical releases along with future ones.

mhvk:
@taldcroft - OK, that makes sense, so the what's-new for a major release would mostly stand out in being one where one has to carefully read the API-change section.

I like the idea of living documents, as old versions are really easy to look at with `git`, but probably good to spell out the alternative: APE 2 would be formally archived and essentially become invisible (or marked with cross-out or so).

astrofrog:
@taldcroft - so just to check, would APE 21 still exist to lay out the motivations etc for the change?

I'm not sure that relying on just 'git log' is really sufficient for people to browse old versions of APEs - instead I think we can point people to the versioned Zenodo entries?

APE21.rst
mhvk: I guess I disagree: I don't think there should be a barrier to making spelling corrections, textual clarifications, etc. It should be fine to do those with edit-on-github, without also having to create a changelog entry.

Conceptually, I see our API as being described in docstrings, and those (if complete) tell which type of error is raised, not what message the error contains. (Similarly, what type of object is returned, not its `repr`.)

Anyway, don't want to make more of this than it is worth. Most important is what we can agree on - that these types of API changes would not have to be in What's New.

APE21.rst
mhvk: Am fine with removing any suggestion to make spelling changes in bugfix releases - my real purpose is just to minimize the burden of making trivial changes. And I agree we should really make sure that bugfix releases do not break any code (to the extent it is under our control of course).

APE21.rst
mhvk: Agreed to rephrase, e.g., "Changes to more specific exception or ..."

mhvk:
:+1: to linking to versioned Zenodo entries.

APE21.rst
mhvk: @nstarman - I think what @taldcroft and I wrote is consistent with your wish! Please suggest rephrasing if it doesn't feel like it is!

taldcroft:
The first time I wrote the "big picture comment", I did think about just dropping APE-21. But then it seemed to me like it deserves its own APE as a substantial project change with a place to lay out the motivations and other usual APE things.

For APE-2, I think the key thing is including some *history* in that document itself. This idea is a bit against our usual standard of doing everything with version control, but in this case for the typical reader of APE-2 it would be helpful to clearly see the history of the previous version. Basically that amounts to a few sentences along the lines of:

Prior to astropy version X.0.0, astropy supported the concept of Long Term Support releases which corresponded to a major release and were maintained with bug fixes for two years. For details see [link to the previous APE-2]. This policy was changed with the acceptance of [APE-21].

APE21.rst
taldcroft: Yes, I think we're saying the same thing. An exception that is now `ValueError` would become a custom exception that inherits from `ValueError` so that existing `except ValueError` statements would typically work.

eerovaher:
Should we consider the stacklevel of a warning to be part of the public API or not?

pllim:
I got lost in these discussions. So, is the only blocker for this APE the debate on "what is public API"? Other details are ok?

pllim:
Also @mwcraig said conda-forge maintenance would be easier with this APE because there is no longer a need to maintain LTS branch in astropy feedstock, so no more issues like https://github.com/conda-forge/astropy-feedstock/issues/118 .

And we can also probably get rid of "lts" build in ReadTheDocs.

astrofrog:
@taldcroft - I started working on modifying APE 2 but the changes are going to be quite significant and it doesn't really feel like just an 'update' - there is text for instance giving the motivation etc for LTS which would need to go, and quite a bit of text would be changed, and e.g. the implementation section would no longer be relevant and so on.

While I think there are APEs that would work well with updates/additions/clarifications, APE 2 would be significantly changed beyond recognition. I wonder if it would make more sense to expand APE 21 so that it is a self-contained replacement of APE 2 and APE 2 could be updated to include 'Replaced by APE 21' near the top (PEPs have 'Status: Replaced' in some cases)

I'm happy to go with the majority opinion, but just wanted to bring up that if I have to update APE 2 it will be almost a complete re-write.

astrofrog:
Having thought about this a bit more, perhaps it does make sense to update APE 2, and I've opened a PR to do this here:

https://github.com/astropy/astropy-APEs/pull/84

I've also removed duplicated content here because I think it would be best to have things only live in one place (otherwise if APE 2 is updated in future, APE 21 may have guidelines inconsistent with APE 2).

APE2.rst
```diff
@@ -5,82 +5,70 @@ author: Erik Tollerud
...
 
-This APE describes the general plan for timing of releases and their version
-numbers.  The basic premise is a 6-month cycle for new feature releases, with
-releases designated "long-term support" (LTS) releases every two years.  Bugfix
-releases occur as needed, on both the current feature release and the LTS
-releases.
-
+This APE describes the general plan for timing of releases of the core package
+and their version numbers. The basic premise follows the terminology of semantic
+versioning: *major* releases (e.g. 6.0.0) every two years, *minor* releases (e.g.
```
pllim, line 24:
```suggestion
versioning: *major* releases (e.g., 6.0.0) every two years, *minor* releases (e.g.,
```

APE2.rst
```diff
@@ -5,82 +5,70 @@ author: Erik Tollerud
...
+6.1.0) every six months in between, and *bugfix* releases (e.g. 6.1.2) as
```
pllim, line 25:
```suggestion
6.1.0) every six months in between, and *bugfix* releases (e.g., 6.1.2) as
```

APE2.rst
```diff
@@ -5,82 +5,70 @@ author: Erik Tollerud
...
+release guidelines (prior versions can be found in `Previous versions of this
+APE`_).
+
+Release versioning and scheduling
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The release cycle for the core package should be as follows:
+
+* Major releases every two years. These are releases that can introduce breaking
+  API changes, and should be numbered as x.0.0 (e.g. 6.0.0).
```
pllim, line 52:
```suggestion
  API changes, and should be numbered as x.0.0 (e.g., 6.0.0).
```

APE2.rst
```diff
@@ -103,42 +91,92 @@ than actual rules.
...
+  what is considered API.
+
+* We should ensure that any changes in releases which deviate from the
+  guidelines above are clearly communicated to users - for example any API/breaking
+  changes in minor releases should be included in the "What's New" page for the
+  minor release.
+
+* Deprecations should always be mentioned in the changelog, as well as any
+  removal of deprecated code/API changes. We do not specify how exactly this
+  should be done, but note that e.g. Numpy release notes have separate sections
```
pllim, line 170:
```suggestion
  should be done, but note that, e.g., Numpy release notes have separate sections
```

APE2.rst
```diff
@@ -149,22 +187,8 @@ N/A
...
-* The first LTS release (v1.0) will then come six months after v0.4, so an
-  appropriate milestone will need to be added to GitHub.  Following that,
-  milestones should be added as needed to reflect the versioning scheme
-  described in this APE.
-* The documentation should be updated to mention the existence of LTS releases
-  and make it clear that they come with a promise of two years of bug fixes (as
-  needed). It should also add a link to this APE so that those interested in the
-  numbering scheme can look here for details.
-
+Whenever this APE is updated, the core package documentation should be updated
```
pllim, line 190: I would be nice to see a draft PR now over at `astropy` that shows me how the release procedures would change to comply with these changes here. Like... Will we be down to one backport branch? Do we no longer need a `stable` branch? Do we no longer need a `lts` build on RTD?

pllim (commented):
Thanks!

APE2.rst
astrofrog: Sounds good, I can try and do this

saimn:
About changelog/communication, I really like how Pandas did it for their recent major release: https://pandas.pydata.org/docs/whatsnew/v2.0.0.html
All the useful information is there, and easy to navigate with clear sections etc.

APE2.rst
```diff
@@ -103,42 +91,92 @@ than actual rules.
...
-permanent, to allow codes to transition to the new interface. A future APE may
-add to or change this with a more concrete policy on when and for how long
-something should be deprecated.
-
+The guidelines in relation to API changes and backward compatibility are as follows:
+
+* API changes (i.e., modifications as opposed to additions) should only be carried out in major releases.
+* Any API changes should be preceded by deprecation warnings that should be
+  visible to users in at least one minor release cycle prior to the API change
+  being carried out.
```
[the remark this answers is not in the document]
astrofrog: Done

APE2.rst
astrofrog: Done

astrofrog:
@mhvk @taldcroft @nstarman - I lost track of whether we are in agreement with the current wording for exceptions/warnings. If not, please feel free to suggest specific changes to the current wording - thanks!

eteq:
These are some consensus discussions from coordination meeting discussion on this APE - including me @nden @pllim @nstarman @weaverba137 @tepickering @tomdonaldson

* need to clarify that the semantic step happens *only if* there are any pending->post-deprecaton
* (might already be clear: major step can happen in between if we break API)
* 6 months is too short for a deprecation period, 2 years is tool long.implies a minimum-deprecation of  1 years == 2 cycles 
* Change to major/minor every two releases instead of four? Or more formally: every *major* release is yearly, minor releases are by default 6 mo after but can be more frequent if we want.
* This leads to actual breaking changes occuring every either one or one-and-a-half years
* (Might need to add something to "implementation" about the above)

Some of that might be a bit hard to understand - I'll follow up later (hopefully today) with some specific wording suggestions to implement this.

One other item to note that relates to the discussion: several observatory support staff in the room said something in the direction of "LTS was good before, but we are ok with this change after seeing the motivation and because the world has changed".  So that's just more fodder of general support for this.

astrofrog:
@mhvk @taldcroft - since you aren't mentioned in @eteq's previous comment and may not have been in on that discussion, are you on board with the suggestions in @eteq's comment? (just want to double check before going ahead and implementing the changes). Thanks!

mhvk:
I agree with the deprecation period; I would not try to schedule major releases ahead of time (i.e., no need for a fixed schedule for those).

taldcroft:
This looks good to me and I agree with @eteq's summary.

@mhvk - I'm not sure I understand "I would not try to schedule major releases ahead of time". I thought the plan is for a major release on a yearly cadence. This might be done with some thought to the timing of the yearly Python major release?

mhvk:
@taldcroft - I was replying to

> Change to major/minor every two releases instead of four? Or more formally: every major release is yearly, minor releases are by default 6 mo after but can be more frequent if we want.

I was just not sure the APE needs to decide on this; astropy seems mature enough it is not clear to me we need a major version bump every year. That said, going together with python updates might make a lot of sense. So, happy to go with whatever was discussed & decided at the meeting!

taldcroft:
@mhvk - if we are strict about API policy and deprecations, I would guess that there will be a driver for a major update every year. But you are right that if by chance there are none in the changelog then there is no need for a major version bump.

saimn:
I agree with the 1 year deprecation period, but not sure about the yearly major release : one of the goals of this APE was to keep the stability promise of the LTS for a ~2 years period. A major release is needed only to introduce breaking changes, i.e. removal of deprecations, or new feature that need an API break. Is there a need for a faster cadence ?

nstarman:
> Is there a need for a faster cadence ?

As I understand it, we don't want deprecations to last for 2.5 years. If something is deprecated in v5.3 and can't be removed until v7 (because v6 is too early) that's too long.

WilliamJamieson:
I agree with @nstarman, it will become difficult to maintain multiple versions of an API for ~2 years. I also agree any deprecation should exist for at least 1 year, so it is reasonable for us to allow for having a major release every year.

Since the goal is to drop LTS version support, we will then be expecting downstream users to migrate forward as we make minor releases. Consequently, downstream users should have at least 1 year of warning prior to a breaking change.

taldcroft:
I agree with the points from @nstarman and @WilliamJamieson above, which I believe are also in agreement with the summary of the Coordination meeting outcome from @eteq.

@saimn - are you OK with the yearly major release? If so it seems like we would have achieved consensus, with the caveat that @astrofrog needs to update the APE's to reflect one year instead of two between major releases.

astrofrog:
I've now tried to implement the 'consensus' from the coordination meeting into these documents. The bottom line as far as I have understood is that we are basically following semantic versioning, and specifying that one year is the *minimum* time between major releases, six months is the default and a maximum time between non-bugfix releases, and deprecations have to be in releases for at least a year before changes are carried out.

@taldcroft @WilliamJamieson @nstarman @pllim @saimn @mhvk @eteq - could you take a look at the latest versions and leave comments/suggestions or approve if you are happy with the APEs as-is? (just to see if we are getting close to consensus)

APE2.rst
```diff
@@ -5,82 +5,87 @@ author: Erik Tollerud
 
 date-created: 2013 November 18
 
-date-last-revised: 2013 December 11
+date-last-revised: 2023 April 27
```
pllim, line 8: Or whatever date when you make more changes.
```suggestion
date-last-revised: 2023 May 13
```

pllim (commented):
I should not approve as I am technically a co-author but I am fine with this. The important thing is that we get buy-in from people very involved with pipelines, who include but not limited to @nden , @perrygreenfield , and @weaverba137 .

Thank you!

APE2.rst
```diff
@@ -5,82 +5,87 @@ author: Erik Tollerud
...
+* 6.2.1
+* 6.3.0 (four months after 6.2.0)
+* 6.3.1
+* 6.3.2
+* 6.3.3
+* 7.0.0 (five months after 6.3.0, 18 months after 6.0.0)
+
+In the above hypothetical scenario, the developers have decided to release minor
+releases more frequently than every six months, and a year after 6.0.0 there was
+no need for a major version, so this was delayed until 18 months after 6.0.0.
```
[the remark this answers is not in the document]
astrofrog: Yes we could in principle never bump the major version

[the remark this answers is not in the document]
pllim: Should this be stated clearly, or it is unnecessary?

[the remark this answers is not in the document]
weaverba137: +1 for stating it clearly.

weaverba137:
I re-read both APEs and they sound good to me. Thanx for the reminder!

saimn:
> specifying that one year is the *minimum* time between major releases

If one year between major releases is a *minimum*, then I agree, but that was not my understanding from the above (*every major release is yearly*).

That was also my point earlier in the discussion (https://github.com/astropy/astropy-APEs/pull/82#discussion_r1115566239), we just need to keep some flexibility on when we want to make a major release. It will depend on how many deprecated items we have accumulated. And if there is some new feature that is waiting for the end of the (1 year) deprecation period, we can wait 6 months more before releasing the next major version. But most of time we are just deprecating old functions or some renamed arguments and there is no urgency in removing those.

Also we will need a way to track the things that have been deprecated to know when they can be removed, and when we decide that next version will be a major one.

astrofrog:
I thought the consensus was that it was a minimum as @eteq said:

> need to clarify that the semantic step happens only if there are any pending->post-deprecaton

APE2.rst
```diff
@@ -5,82 +5,87 @@ author: Erik Tollerud
...
-date-last-revised: 2013 December 11
+date-last-revised: 2023 April 27
 
 date-accepted: 2013 December 10
 
 type: Process
 
 status: Accepted
 
+revised-by: Thomas Robitaille, Pey Lian Lim, Simon Conseil - 2023 April 27 - Rewritten following the approval of APE 21 to no longer have Long Term Support (LTS) versions
```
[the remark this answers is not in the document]
astrofrog: Well this was in case there are multiple revisions in future? (which there might well be) I was following the canonical example from APE 1.

APE2.rst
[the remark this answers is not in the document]
astrofrog: Done

taldcroft (approved):
Looks great to me, thanks for all the work and persistence!

APE2.rst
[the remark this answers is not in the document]
pllim: Oh, okay. Ignore me then!

pllim:
How many approvals are needed here? How should we proceed? Thanks!

hamogu (approved):

mhvk (approved):
Looks great, thanks for sticking with it!

dhomeier (approved):
Thanks, I see everything I recall from the discussions reflected here!

pllim:
Thank you for the approval! Should we put the final date stamp and status on this PR and merge?

hamogu:
This has been approved by the Coco through the approvals here and internal Coco communication by those Coco members that did not comment here. I will go through the process of merging and try publishing on Zenodo etc. later today (if zenodo let's me - their webinterface has been flaky recently, if not, I might have to re-try in the next few days).

APE21.rst
```diff
@@ -0,0 +1,144 @@
+Ending Long Term Support Releases
+---------------------------------
+
+author: Thomas Robitaille, Pey Lian Lim, Simon Conseil
+
+date-created: 2023 February 16
+
+date-last-revised: 2023 May 15
+
+date-accepted: <replace with accepted date>
```
hamogu, line 10:
```suggestion
date-accepted: 2023 May 31
```

APE21.rst
```diff
@@ -0,0 +1,144 @@
...
+
+type: Process
+
+status: Discussion
```
hamogu, line 14:
```suggestion
status: accepted
```

hamogu (approved):

PR #85, opened 2023-04-28
Public API

nstarman:
Up for discussion!

~Very much a work in progress. Hopefully refined by discussion at the upcoming conference.~

### DIFF
--- a/APE_public.rst
+++ b/APE_public.rst
@@ -1,0 +1,513 @@
+
+Astropy's API Policy on Modules and their Contents
+--------------------------------------------------
+
+author: Nathaniel Starkman
+
+date-created: 2013 November 5 <replace with the date you submit the APE>
+
+date-last-revised: 2013 December 17 <keep this up to date anytime something
+changes>
+
+date-accepted: 2014 January 20 <replace with accepted date>
+
+type: <one of these three: Standard Track, Informational, Process>
+
+status: Discussion
+
+
+Abstract
+--------
+
+This APE establishes Astropy's policy for public versus internal modules and
+their contents.
+
+
+Scope
+-----
+
+A library's API is the publicly accessible "surface": the set of modules,
+classes, functions, and other symbols provided to the library's users. This APE
+is narrowly focused on which modules are public and how to communicate which
+contents of those modules are public. This APE does not address any other
+questions of API design, e.g. public versus internal methods in classes, leaving
+these for future APEs or future revisions to this APE.
+
+
+.. _description_of_the_problem:
+
+Description of the Problem
+--------------------------
+
+Pure Python has no language-level means to establish and enforce separations
+between public versus internal interfaces -- all code objects can be accessed.
+This is in contrast to languages like C++ or Java, which have language-level
+constructs to establish public versus private interfaces. Differentiating
+between public and internal interfaces is important for everyone, users and
+developers alike, to know what interfaces are stable and supported, and which
+are not. Consequently, Python has developed conventions for communicating what
+is public versus internal. Nominally there are three means (discussed more
+formally later):
+
+1. **Underscore Prefix**.  By convention, names that begin with a single underscore
+   are considered internal.
+2. ``__all__``.  The dunder attribute ``__all__`` of a module is a list of
+   names that are considered public.
+3. **Documentation**.  Public interfaces are documented and internal interfaces
+   are not, or are specifically noted.
+
+These 3 means can be ambiguous and contradictory. Let's explore this point with
+an example. Consider an example library with the following layout::
+
+   src/
+      __init__.py::
+         __all__ = ["foo", "_bar"]
+
+      foo.py::
+         __all__ = ["_func"]
+         def func():
+            pass
+
+         def _func():
+            pass
+      
+      _bar.py::
+         __all__ = ['bunc']
+         def bunc():
+            pass
+
+         def _bunc():
+            pass
+
+      baz.py::
+         __all__ = ['qux']
+
+         def qux():
+            pass
+
+- Is the ``foo`` module public? Probably, but what if it's not documented?
+- Is ``foo.func`` public? It is not in ``foo.__all__``. 
+- Is ``foo._func`` public?  It is prefixed with an underscore, but in ``foo.__all__``. 
+- Is the ``_bar`` module public? It is prefixed with an underscore, but in ``src.__all__``.
+- Is ``_bar.bunc`` public? It is in ``_bar.__all__``, but what if ``_bar`` is private?
+- Is ``_bar._bunc`` public? Probably not, but what if it's documented?
+- Is ``baz`` public? It does not start with an underscore, but is not in ``src.__all__``.
+- Is ``baz.qux`` public? It does not start with an underscore and is in
+  ``baz.__all__``, but what if ``baz`` is private?
+
+
+Astropy partially implements all three of the conventions: some private classes
+and functions are prefixed with an underscore, while some are not; some modules
+have an ``__all__`` attribute, and some do not;  some (debatably) internal
+interfaces are in the public documentation, like ``astropy.utils``, and some are
+not. However, Astropy does not have a uniform approach to differentiating and
+communicating public vs internal symbols, which leads to confusion for users and
+developers alike. This longstanding issue has steadily become more pressing with
+the development of assistive code tools, like auto-fill in IDEs or more recently
+with generative coding tools, like GitHub CoPilot, which suggest code to users
+based on a corpus of code examples.  When users unintentionally use internal
+code, code tools will then suggest that internal code to other users. If we do
+not have a uniform and systematic approach to communicating what is public vs
+internal, then we cannot expect users and especially their tools to know what is
+public vs internal.
+
+The ambiguity between public and internal interfaces also creates problems for
+developers. When private APIs are not clearly distinguished from public ones,
+they get used downstream -- as was common with NumPy -- thereby becoming
+effectively frozen even though they were never intended to be stable. This
+happens because:
+
+1. Users (knowingly or unknowingly) start depending on private APIs.
+2. Maintainers become hesitant to refactor or improve private code for fear of
+   breaking downstream users.
+3. The cost of cleanup increases over time as more code depends on the
+   quasi-public private APIs.
+4. The boundary between public and private becomes meaningless.
+
+This creates bad incentives: the behavior is individually rational ("don't break
+anyone today") but collectively harmful ("downstream increasingly relies on what
+should never have been relied on"). The whole point of having private APIs is
+precisely so that they can be changed as needed. When private APIs are
+confusingly public, improving them becomes tortuous for developers, and the
+library suffers as a result.
+
+
+Solution Criteria
+-----------------
+
+Having established that we need a uniform and systematic approach to public vs
+internal interfaces, we now decide what that approach should be. The approach
+should fulfill a few criteria:
+
+1. Align closely with established conventions, so that it is familiar
+   to users and developers, and hopefully detected by assistive tools.
+2. Be unambiguous and easy to understand.
+3. Be straightforward to implement and maintain, so that it is not a
+   burden on developers.
+
+We start by considering established practice, for criterion 1, and how to ensure
+that we satisfy the latter two criteria and can be implemented in a uniform and
+systematic way. There are three important documents that describe established
+practice as it relates to Astropy:
+
+1. `PEP 8 <https://peps.python.org/pep-0008/#public-and-internal-interfaces>`_ :
+   the Python style guide.
+2. `SciPy
+   <https://docs.scipy.org/doc/scipy/reference/index.html#importing-from-scipy>`_:
+   the largest scientific Python project.
+3. `Python's typing guide
+   <https://github.com/python/typing/blob/master/docs/source/libraries.rst#library-interface-public-and-private-symbols>`_:
+   the official guide for how to statically analyze Python code, which is important
+   for assistive tools.
+
+Starting with `PEP 8
+<https://peps.python.org/pep-0008/#public-and-internal-interfaces>`_, it states
+that::
+
+   Documented interfaces are considered public, unless the documentation
+   explicitly declares them to be provisional or internal interfaces exempt
+   from the usual backwards compatibility guarantees. All undocumented
+   interfaces should be assumed to be internal.
+
+   To better support introspection, modules should explicitly declare the names
+   in their public API using the ``__all__`` attribute. Setting ``__all__`` to
+   an empty list indicates that the module has no public API.
+
+   Even with ``__all__`` set appropriately, internal interfaces (packages,
+   modules, classes, functions, attributes or other names) should still be
+   prefixed with a single leading underscore.
+
+   An interface is also considered internal if any containing namespace
+   (package, module or class) is considered internal.
+
+   Imported names should always be considered an implementation detail. Other
+   modules must not rely on indirect access to such imported names unless they
+   are an explicitly documented part of the containing module's API, such as
+   ``os.path`` or a package's ``__init__`` module that exposes functionality from
+   submodules.
+
+
+This is a more detailed and nuanced description of the three means mentioned in
+:ref:`description_of_the_problem`.  It is also the most authoritative source, as
+it is the official Python style guide.  However, it is not as unambiguous as we
+require: which takes precedence, the ``__all__`` attribute or the underscore
+prefix?
+
+`SciPy
+<https://docs.scipy.org/doc/scipy/reference/index.html#importing-from-scipy>`_
+tries to resolve this ambiguity by stating that ::
+
+   - Methods / functions / classes and module attributes whose names begin with
+     a leading underscore are private.
+   - If a class name begins with a leading underscore, none of its members are
+     public, whether or not they begin with a leading underscore.
+   - If a module name in a package begins with a leading underscore none of its
+     members are public, whether or not they begin with a leading underscore.
+   - If a module or package defines ``__all__``, that authoritatively defines
+     the public interface.
+   - If a module or package doesn't define ``__all__``, then all names that
+     don't start with a leading underscore are public.
+
+This is a good solution, and is consistent with PEP 8, but for our goal of
+complete unambiguity and uniformity it falls short on two counts: first, it
+does not mention the documentation; second, PEP 8 recommends "modules should
+explicitly declare the names in their public API using the ``__all__``
+attribute. Setting ``__all__`` to an empty list indicates that the module has no
+public API." SciPy allows modules to lack an ``__all__`` attribute, meaning a
+user and their tools must understand the nuances of the previous rules. Having
+an ``__all__`` attribute in every module is simpler, clearer, and better for
+introspection by both users and automated systems.
+
+Finally, let's consider what `Python typing guide
+<https://github.com/python/typing/blob/master/docs/source/libraries.rst#library-interface-public-and-private-symbols>`
+adds.:: 
+
+   - Symbols whose names begin with an underscore (but are not dunder names) are
+     considered private.
+   
+   - A module can expose an ``__all__`` symbol at the module level that provides
+     a list of names that are considered part of the interface. This overrides
+     all other rules above, allowing imported symbols or symbols whose names
+     begin with an underscore to be included in the interface.
+
+This is consistent with PEP 8 and SciPy, but it does not mention the
+documentation. The Python typing guide reinforces the idea that the ``__all__``
+attribute is the authoritative source of what is public vs internal.
+Importantly, the ``__all__`` attribute is a module-level attribute, so it only
+applies to the module in which it is defined. We will refer to this as "locally"
+public. This means that a module can define an ``__all__`` attribute but if the
+module itself is not public, then anything in the ``__all__`` attribute cannot
+be publicly accessed from outside the module.
+
+
+Solution
+--------
+
+We propose the following:
+
+1. That Astropy adopts the PEP 8 rules on public versus internal interfaces
+2. That Astropy disambiguate these rules similarly to SciPy.
+3. In addition, that Astropy ensures its documentation is consistent with its
+   code, the latter being the authoritative source.
+
+**Rules for Public Interfaces:**
+
+1. A symbol is public if and only if all containing namespaces are public.
+   A symbol is internal if any containing namespace is internal.
+
+   For example, `astropy/path/to/code` is public if and only if all `astropy`,
+   `astropy/path`, `astropy/path/to`, and `astropy/path/to/code` are all public.
+   If any of these namespaces is internal, then `astropy/path/to/code` is
+   internal.
+
+2. All modules must have an ``__all__`` attribute, even if it is empty. The
+   ``__all__`` attribute specifies the public and internal interface of the
+   module in which it is defined. Anything in ``__all__`` is *locally* public,
+   including underscore-prefixed symbols. Anything not in ``__all__`` is
+   internal in that module. The exception to this rule are modules that cannot
+   have an ``__all__`` attribute, for example implicit namespace packages, which
+   lack an ``__init__.py``. In these cases, the public interface is defined by
+   the sub-rules:
+
+   - A symbol is *locally* public if it does not start with an underscore, and
+     internal otherwise, unless it is a dunder symbol (``__<...>__``).
+   - Dunder symbols (``__<...>__``) have their own public vs internal rules,
+     which are not determined in this APE.
+
+3. Public symbols must not be prefixed with an underscore.
+
+   - Internal symbols need not be prefixed with an underscore, but it is often
+     recommended. An example of when it is not needed are symbols defined in
+     internal modules, making the prefix redundant.
+
+4. Public symbols must be documented.
+
+   - If an internal symbol is documented (which is not recommended), it must be
+     unambiguously, explicitly, (and preferably repeatedly) noted as internal.
+
+
+Let's consider an example::
+
+   src/
+      __init__.py::
+         __all__ = ["foo", "spam"]
+
+      foo.py::
+         __all__ = ["func"]
+         def func(): pass
+         def _func(): pass
+      
+      _bar.py::
+         __all__ = ['bunc']
+         def bunc(): pass
+         def _bunc(): pass
+
+      baz.py::
+         __all__ = ['qux']
+         def qux(): pass
+
+      spam/  # implicit namespace package
+
+         ham.py::
+            __all__ = ['eggs']
+            def eggs(): pass
+
+In this case, ``func`` is public in ``foo``, and  ``foo`` is public in ``src``,
+so ``src.foo.func`` is public. In contrast, while ``bunc`` is public in
+``_bar``, ``src._bar`` is not public, so ``src._bar.bunc`` is not public either.
+This is ''locally'' public, as in internal. Both ``_func`` and ``_bunc`` are
+internal.  Note that ``src.baz`` is internal because it is not in
+``src.__all__``, even though it is not prefixed with an underscore. This follows
+the strict rules for public interfaces established above, but as is also noted,
+it is a very bad idea to name an easily accessed module in such a way that it
+appears public but is not.
+
+=======  =========  ==========  ==========  =========================  =========
+Symbol                                      Status                             Good Idea
+=======  =========  ==========  ==========  =========================  =========
+``src``                                     public                     ✅
+         ``.foo``                           public                     ✅
+                    ``.func``               public                     ✅
+                    ``._func``              internal                   ✅
+         ``._bar``                          public                     ✅
+                    ``.bunc``               internal (locally public)  ✅
+                    ``._bunc``              internal                   ✅
+         ``.baz``                           internal                   ❌
+                    ``.qux``                internal (locally public)  ✅
+         ``spam``                           public                     ✅
+                    ``.ham``                public                     ✅
+                                ``.eggs``   public                     ✅
+=======  =========  ==========  ==========  =========================  =========
+
+
+Branches and pull requests
+--------------------------
+
+N/A
+
+
+Implementation
+--------------
+
+The process of adopting this APE will change Astropy's API. This APE formally
+establishes Astropy's API, so what is meant by "change"?  There is already a *de
+facto* API, which is what users *think* is the API.  This is determined
+primarily by the documentation, but also by tab-completion within an interactive
+session.  In the implementation of this APE to establish a *formal* API, we aim
+to minimize changes to the *de facto* API. This will be accomplished in the
+following 5 phases:
+
+1. **Snapshot the documentation**. As of the adoption of the APE a snapshot of
+   the documentation will be saved. This will be used to determine what is
+   currently public and what is not. On a per-sub-package basis, until the APE's
+   adoption is complete, this snapshot is authoritative, e.g. dictating what
+   must be added and removed from ``__all__``, for deciding what must undergo a
+   deprecation process, etc.
+
+2. **Add / update** ``__all__``. The ``__all__`` in each module will be updated
+   to reflect phase 1. Any modules' missing ``__all__`` will have one added.
+
+3. **Update the documentation** and **implement deprecations**.
+
+   - Make sure every public symbol is in the documentation, unless it is being
+     deprecated.
+   - Deprecate any public symbol that is not intended to be public, but was made
+     public as part of steps 1 and 2.
+   - Add a SciPy-like section to the developer documentation explaining the
+     public vs internal rules.
+   - Fix links to always point to the public interface, not the internal
+     interface. (This will presumably require a similar implementation as used
+     in NumPy to change the ``__module__`` attribute of many objects.)
+   - Add a reminder to the maintainer checklist that to be public a symbol must be
+     in ``__all__`` and documented.
+   - Clearly state if a documented object is actually private.
+
+4. **Add a pre-commit hook**. This will ensure that the ``__all__`` attribute is
+   always present and up-to-date. It will also ensure that the documentation is
+   always up-to-date with the ``__all__`` attribute. This can be accomplished
+   with a pre-commit hook that runs a script that checks the ``__all__``
+   attribute and, if the module is public, searches in the ``docs/api``
+   directory for symbols in ``__all__``.
+
+5. **Add prefixes**: Add prefixes to the top-most private symbols. For modules
+   this makes all their contents private.
+
+In Astropy core each phase will be implemented on a per-sub-package basis with
+individual pull requests.
+
+
+Implementation Examples
+-----------------------
+
+1. Astropy's cosmology sub-package
+
+The ``astropy.cosmology`` sub-package serves as an example implementation of
+this APE. In ``cosmology``, all internal code was moved into a ``_src`` folder,
+making everything within it private. For backwards-compatibility, the old public
+API was then exposed through backwards-compatibility modules that use
+``__getattr__`` to import and deprecate the contents of ``_src``. For example,
+``flrw.py`` was moved to ``_src/flrw.py``, and a new ``flrw.py`` was created
+with a ``__getattr__`` that imports and deprecates the contents of
+``_src/flrw.py``. This deprecation process was completed and the
+backwards-compatibility modules removed after a full year for version 8.0, with
+no user complaints. The result is a clear and robust separation between the
+public API and the private implementation (anything in ``_src``).
+
+Other sub-packages may prefer to underscore-prefix individual modules rather
+than use the ``_src/module`` pattern. Both approaches provide clear
+communication to users and developers about what is public and what is internal.
+
+2. NumPy
+
+NumPy has adopted a similar approach to this proposal, by moving most of their
+internal code into a ``_core`` folder.
+
+3. SciPy
+
+SciPy has adopted a similar approach to this proposal, by moving most of their
+internal code into underscore-prefixed modules and packages.
+For example, see the `scipy.fft` package.
+
+
+Teaching This
+-------------
+
+This APE requires very little teaching. For users, the rules are simple:
+
+1. Check the tree of ``__all__`` to see if a symbol is public (Rules 1 and 2). This is the normative definition of what is public.
+2. Everything conforms to normal expectations, e.g. public modules don't start
+   with underscores (Rule 3).
+3. The documentation follows the code (Rule 4).
+
+So in practice users can just tab-complete in an IDE or check the documentation
+to see what is public.
+
+For most developers using modern code tools, this APE requires no teaching at
+all. Code tools will automatically pick up public vs private API.
+
+For developers using code tools that don't autocomplete underscore-prefixed
+files (typically only old editors) but want this functionality, there is an easy
+solution: create a symlink from the private file to a public name and then
+git-exclude it. For example, create ``cosmology/src`` as a symlink to
+``cosmology/_src``. What's nice about this solution is that it's permanent -- do
+it once and forget about it.
+
+
+Backward compatibility
+----------------------
+
+This APE breaks backward compatibility in two ways:
+
+1. Changes ``__all__`` in many modules. Many modules define an ``__all__`` but
+   include symbols that are intended to be private because they should be
+   imported from other modules (generally the top of the module). This will not
+   break code that directly imports the symbols (as Python does not use
+   ``__all__`` for this purpose), but it will break code that expects private
+   symbols and uses ``import *``.
+2. Adds prefixes to many symbols that are not public. This can be done in a
+   backward compatible way by adding a ``__getattr__`` method to the module that
+   raises a warning for any object that is not public. This will allow existing
+   code to continue to work, while encouraging people to fix their code to use
+   the public interface.
+
+
+Alternatives
+------------
+
+**We do nothing:**
+
+This is the status quo.  It is not a good option because it does not solve the
+issue.  The aforementioned problems of not knowing what is stable and
+supported, and what is not, remain.
+
+**We allow** ``__all__`` **to be optional:**
+
+This is not great.
+
+The only time this might be good is when a module has dynamic symbols from a
+`PEP 562 <https://peps.python.org/pep-0562/>`_ module-level ``__getattr__``
+method. However, if it is truly dynamic then it cannot be statically analyzed, which
+is undesirable for other reasons. If it is not truly dynamic, then the "dynamic"
+symbols can be added to ``__all__`` and an ``__init__.pyi`` file used to
+communicate the public interface to static analyzers.
+
+**We make the documentation the authoritative definition of the API.**
+
+Good code is well documented in the code. Beyond code comments and docstrings, what is
+public versus internal is an important aspect of the code, and should be
+communicated in the code itself. It is vitally important that the code and
+user-facing documentation are consistent, and the user-facing documentation
+should reflect the code.
+
+**We make tab-completion the authoritative definition of the API.**
+
+This is not a good option because it does not work with implicit namespace
+packages nor `PEP 562 <https://peps.python.org/pep-0562/>`_ module-level
+``__getattr__`` methods.
+
+
+Decision rationale
+------------------
+
+<To be filled in by the coordinating committee when the APE is accepted or
+rejected>


APE_public.rst
```diff
@@ -0,0 +1,284 @@
...
+
+This is a good solution, and is consistent with PEP 8, but for our goal of
+complete unambiguity and uniformity it falls short on two counts:  first, it
+does not mention the documentation; second, PEP 8 recommends "modules should
+explicitly declare the names in their public API using the ``__all__``
+attribute. Setting ``__all__`` to an empty list indicates that the module has no
+public API." SciPy allows modules to lack an ``__all__`` attribute, meaning a
+user and their tools must understand the nuances of the previous rules. Having
+an ``__all__`` attribute in every module is simpler, unambiguous, and better for
+introspection by both users and automated systems.
```
astrofrog, line 136: It might be worth mentioning at this point how a user would know that a function/class is in ``__all__``. Is it not possible to explicitly import functions/classes in a module that are not in ``__all__``? If so then having something in ``__all__`` does not stop e.g. CoPilot from suggesting code that uses a function not in ``__all__``.

APE_public.rst
```diff
@@ -0,0 +1,284 @@
...
+
+Rules:
+
+1. A symbol is public if all containing namespaces (packages, modules, classes,
+   functions, attributes or other names) are public. A symbol is private if any
+   containing namespace is private.
+2. All modules must have an ``__all__`` attribute, even if it is empty. The
+   ``__all__`` attribute defines the public and private interface of the module
+   in which it is defined. Anything in ``__all__`` is public, including
+   underscore-prefixed symbols. Anything not in ``__all__`` is private in that
```
astrofrog, line 204: although I think we should probably disallow underscore-prefixed symbols in ``__all__``?

APE_public.rst
```diff
@@ -0,0 +1,284 @@
...
+3. That Astropy ensures its documentation is consistent with its code, the
+   latter being the authoritative source.
+
+   - That Astropy explains these rules in its documentation
+4. That Astropy also adopts these rules for coordinated packages, and recommends
+   that affiliated packages follow these rules as well.
+
+Rules:
+
+1. A symbol is public if all containing namespaces (packages, modules, classes,
```
astrofrog, line 198: Is that *if and only if*?

astrofrog (commented):
Just a few comments/questions so far - one of the main things that bothers me in general and that I don't think there is a solution to is that I can do e.g.:

```
from astropy.cosmology.realizations import ScienceState
```

Obviously this isn't public API, and ``ScienceState`` isn't in ``__all__``, but there's not real way to prevent users from relying on it, and we can't rename all imports in a module to include a ``_`` prefix. So I think that we should probably also make it so that our API docs are *also* an authoritative source of public API and ensure that it's consistent with the other rules. We should also make sure that *all* public APIs are documented in the docs (we don't check that this is the case right now).

APE_public.rst
```diff
@@ -0,0 +1,284 @@
...
+N/A
+
+
+Implementation
+--------------
+
+This APE will be implemented in 3 phases:
+
+1. **Add** ``__all__``: An ``__all__`` dunder attribute will be added to all
+   modules that do not have it.
```
astrofrog, line 226: What about also having a pre-commit step to enforce this?

APE_public.rst
```diff
@@ -0,0 +1,284 @@
...
+2. **Update the documentation** to reflect the new rules.
+
+   - Add a scipy-like section to the developer documentation.
+   - Make sure the developer documentation is consistent with the new rules.
+   - Fix links to always point to the public interface, not the internal
+     interface.
+   - Clearly state if a documented object is actually private.
+3. **Add prefixes**: 1. Add prefixes to all modules that are not public. 2. Add
+   prefixes to all classes, functions, and attributes that are not public.
+      :note:`I'm less enthusiastic on this point.`
```
astrofrog, line 236: I'm in favor of this as long as - as described above - we don't need to add underscore to symbols that are already in an underscore module for instance (so e.g. nothing in ``astropy.io.fits._tiled_compression`` needs to have an underscore prefix.

astrofrog:
After further thought, I think what is going to be really important here is to define what public API is *from the perspective of a user* - that is, a user won't know what ``__all__`` is, so when we communicate with a user, should we tell them that public API is anything in the API docs, or anything that can be accessed through tab completion in IPython and which does not have ``_`` prefixes anywhere?

saimn:
The problem with `__all__` is that it only affects `import *` (and for a long time it was it's only meaning, the pep8 section about public/private interface was added later, https://github.com/python/peps/commit/7dba60ceea8842f00641893295d22338d99cc958). So it doesn't prevent importing from a module, and autocompletion doesn't use it.
So the only way to enforce public/private API is by renaming private modules with an underscore. That's what Scipy did.

saimn:
What do you propose for submodules that currently define `__all__` but from which people should not import directly, e.g. `astropy.convolution.core` (and many others). This is the scheme that is mostly used currently in Astropy.

astrofrog:
Maybe in that case .core should technically be private?

nstarman:
> Maybe in that case .core should technically be private? (._core)

That is the suggestion of PEP 8.

nstarman:
> What do you propose for submodules that currently define `__all__` but from which people should not import directly, e.g. `astropy.convolution.core` (and many others). This is the scheme that is mostly used currently in Astropy.

So this would be one of the largest changes to Astropy. PEP8, Scipy, and static typing all say that `__all__` refers to what is public *in that module*. With this in mind, doing
```python
# __init__.py
# No __all__ is defined
from .core import *
``` 
```python
# core.py
__all__ = ["Foo"]

class Foo: ...
```

Means that `Foo` is public in `astropy.convolution.core` and private in `astropy.convolution`. This is contrary to how Astropy currently does it, where `__all__`in `astropy.convolution.core` means that it is actually private in `astropy.convolution.core` and public in `astropy.convolution`. This is confusing for many reasons. If we were to adopt PEP8 (as suggested in this draft APE) then the previous example would look like

```python
# __init__.py
__all__ = ["Foo"]  # Foo is public in this module, even if it is defined elsewhere.
from .core import Foo
``` 
```python
# core.py
__all__ = []  # nothing is public in this module. Please look elsewhere.

class Foo: ...
```

Essentially we need to move the contents of various ``__all__`` to where the code is actually public, leaving behind empty `__all__` to indicate where no code is public.

APE_public.rst
nstarman: I am fine with that. I have reservations about going all in on underscore prefixing everything. Not needing to underscore symbols in private modules (which still have an ``__all__``) SGTM.

nstarman:
> After further thought, I think what is going to be really important here is to define what public API is from the perspective of a user - that is, a user won't know what __all__ is, so when we communicate with a user, should we tell them that public API is anything in the API docs, or e.g. anything that can be accessed through tab completion in IPython and which does not have _ prefixes anywhere?

I think Yes. To make sure we're on the same page, I think communicating it this way to users should be the logical consequence of the deeper rules:

- That ``__all__`` is authoritative
- That we follow PEP 8, e.g. with undercore prefixes
- That the docs are up-to-date.

Having all these rules means that a user can only get to public symbols though other public symbols -- that "anything that can be accessed through tab completion in IPython and which does not have _ prefixes anywhere" is public. It's not the source or our definition of public API, it's the consequence.

nstarman:
> The problem with __all__ is that it only affects import * (and for a long time it was it's only meaning, the pep8 section about public/private interface was added later, https://github.com/python/peps/commit/7dba60ceea8842f00641893295d22338d99cc958). So it doesn't prevent importing from a module, and autocompletion doesn't use it.
So the only way to enforce public/private API is by renaming private modules with an underscore. That's what Scipy did.

I agree, ``__all__`` is not enough to prevent autocomplete, though some autocomplete does use ``__all__``: see https://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-IPCompleter.limit_to__all__.
I think we should rename modules with an underscore, as part of adhering to PEP 8.
It should be noted that adding underscores does not actually "enforce"  public/private API as Python does not have true language-level features for public vs internal interfaces. Like ``__all__``, single underscores are convention and Scipy says that ``__all__`` takes precedence over underscores. In this APE I propose that we adopt a Scipy-like rule set where ``__all__`` takes precedence over underscores and we use both according to PEP 8.

APE_public.rst
nstarman: I agree that it practically should not, but I strongly think `__all__` should be definitive, so if we have an underscore-prefixed object that we want to make public the mandatory steps are (in this order):
1. put it in `__all__`.
2. update the docs to reflect `__all__`.
3. super strongly encouraged to remove the underscore prefix.

This is adopting the Scipy disambiguation of PEP 8, adding only the mandate that empty `__all__` be included in modules with no Public API.

APE_public.rst
nstarman: Yes. A connected graph through Public API. No private islands for non-dev users.

APE_public.rst
nstarman: As in it checks that `__all__` exists?

eteq:
(Some of this developed from discussion at the coordination meeting (including @nstarman, @saimn , @astrofrog , @tepickering, @pllim, @nden, @williamjamieson), although I don't think I can say that all of my points above are consensus of those folks, it's some mix of that and just straight up my own opinion.)

I very much agree with @astrofrog's point here:

>  After further thought, I think what is going to be really important here is to define what public API is from the perspective of a user - that is, a user won't know what __all__ is, so when we communicate with a user, should we tell them that public API is anything in the API docs, or e.g. anything that can be accessed through tab completion in IPython and which does not have _ prefixes anywhere?

Which I think has up until this point (in an uncodified way) is that whatever the docs say is the public API. So I think it makes sense to codify that as the "true answer".  @nstarman's point was that if we follow the rules here, it's the *same* between those, and that it's only an aberration if these are different. But I was/am concerned about the inevitable state when something isn't working right.  So I think what we settled on is that we start by saying as of when this APE is accepted, the docs are the "true" public API, but this APE presents a plan to get to a state where the rules highlighted in this APE lead naturaly to the docs just reflecting the same thing as these rules.

I still personally think we should have it true that the *final* source of authority is the documentation, because that's more user-facing of a contract, as @astrofrog says. But I think if we say that's the reality now, and we might re-visit it after this APEs plan is implemented, that's a reasonable compromise.

Two more opinions to offer:

* A consequence of this is that modules like `astropy/quantity/quantity.py` become `astropy/quantity/_quantity.py`.  I think leading underscore module names are ugly. That's subjective, but still.
* Another consequence is that the actual public API `__all__` s are in different files than the thing-to-be-made-public E.g., `Quantity` would be in `astropy/quantity/__init__.py` instead of `astropy/quantity/_quantity.py`. I don't like that because it means a small change in one file requires one to understand the full API structure to know which `__all__` to add it to.  I'm not sure that's annoying enough to justify changing anything, but it's a complaint I want to register and think about how we might get around it.

And one question:

* Does this apply to coordinated packages in addition to the core?  In principle it should, but that might be signing up for a *lot* more work because they are probably more of a mess than the core...

pllim:
> Does this apply to coordinated packages 

I would say not. Even for things like removing astropy-helpers, it was a tedious campaign with writing up a transition guide and opening some PRs downstream. For things like this that would break API, it is a non-starter.

nstarman:
@astrofrog @eteq @saimn @pllim, I've updated this APE based on the discussions we had at the conference an in this thread. LMK what you think!

nstarman:
> What do you propose for submodules that currently define `__all__` but from which people should not import directly, e.g. `astropy.convolution.core` (and many others). This is the scheme that is mostly used currently in Astropy.

@saimn I've been working on this over at `astropy.cosmology`. We've successfully transitioned `.utils` -> `._utils`, `io` -> `._io`, and I'm working on the rest. The code is clearer from a user's perspective since there's only one obvious place to import thing from and all the hidden modules aren't tab-completion discoverable.

nstarman:
> * Option 2: rename submodules with underscore, keep their list of public functions/classes in `__all__` (a lot of them already have it) and just change the import in subpackages' `__init__.py` (`from .module import *` → `from ._module import *`). That's what Scipy did. [1]

I also like Option 2 a lot. It works *because* `_module` is not made public in `__init__`, so even though `_module` defines an `__all__` and makes it's contents **locally / contextually** public that is within a **private** module. It's impossible to publicly navigate to the contents of `_module`, only what is exported to `__init__`.

nstarman:
> So I think what we settled on is that we start by saying as of when this APE is accepted, the docs are the "true" public API, but this APE presents a plan to get to a state where the rules highlighted in this APE lead naturaly to the docs just reflecting the same thing as these rules.

@eteq, I agree. 

> But I was/am concerned about the inevitable state when something isn't working right. 

I added a section on a pre-commit CI check. It actually looks to be fairly simple to check that a public module has corresponding documentation since we have ``docs/api`` that collects our documented objects.

APE22.rst
```diff
@@ -0,0 +1,439 @@
...
+communicating public vs internal symbols, which leads to confusion for users and
+developers alike. This longstanding issue has steadily become more pressing with
+the development of assistive code tools, like auto-fill in IDEs or more recently
+with generative coding tools, like GitHub's CoPilot, which suggest code to users
+based on a corpus of code examples.  When users unintentionally use internal
+code, code tools will then suggest that internal code to other users. If we do
+not have a uniform and systematic approach to communicating what is public vs
+internal, then we cannot expect users and especially their tools to know what is
+public vs internal.
+
```
nstarman, line 112: From https://github.com/astropy/astropy/issues/15169, I recall another source of issues that has bitten astropy in the past: I/O registration. When deep sub-modules look public, e.g. `astropy.cosmology.flrw.lambdacdm.LambdaCDM` then that is generally used as the class pathway in I/O rather than the better `astropy.cosmology.LambdaCDM`. When the class is moved it breaks the I/O and needs a whole backwards-compatibility logic to support files using the previous path. Cleaning up public vs internal, e.g. to `astropy.cosmology._flrw.lambdacdm.LambdaCDM` would make this an obviously bad way to serialize the class and code authors would get it "right" (`astropy.cosmology.LambdaCDM`) the first time. Clear public API makes for more stable code.

probably deserves a few sentences.

APE22.rst
```diff
@@ -0,0 +1,439 @@
...
+This APE breaks backward compatibility in two ways:
+
+1. Changes ``__all__`` in many modules. Many modules define an ``__all__`` but
+   include symbols that are intended to be private because they should be
+   imported from other modules (generally the top of the module). This will not
+   break code that directly imports the symbols (as Python does not use
+   ``__all__`` for this purpose), but it will break code that expects private
+   symbols and uses ``import *``.
+2. Adds prefixes to many objects that are not public. This can be done in a
```
nstarman, line 393:
```suggestion
2. Adds prefixes to many symbols that are not public. This can be done in a
```

APE22.rst
```diff
@@ -0,0 +1,439 @@
...
+The only time this might be good is when a module has dynamic symbols from a
+`PEP 562 <https://peps.python.org/pep-0562/>`_ module-level ``__getattr__``
+method. However, if it truly dynamic then it cannot be statically analyzed, which
+is undesirable for other reasons. If it is not truly dynamic, then the "dynamic"
+symbols can be added to ``__all__`` and an ``__init__.pyi`` file used to
+communicate the public interface to static analyzers.
+
+**We make the documentation the authoritative definition of the API.**
+
+Good code is well documented. Beyond code comments and docstrings, what is
```
nstarman, line 422:
```suggestion
Good code is well documented in the code. Beyond code comments and docstrings, what is
```

APE22.rst
```diff
@@ -0,0 +1,439 @@
+Astropy's Public API Policy
```
nstarman, line 1:
```suggestion
Astropy's API Policy on Modules and their Contents
```

APE22.rst
```diff
@@ -0,0 +1,439 @@
...
+
+Scope
+-----
+
+A library's API is the publicly accessible "surface": the set of modules,
+classes, functions, and other symbols provided to the library's users. This APE
+is narrowly focused on the module level of the API -- what modules are public
+and what modules are internal -- and the objects within those modules -- which
+are public and how this is communicated. This APE does not address any other
+questions of API design, e.g. public versus internal methods in classes.
```
nstarman, line 33:
```suggestion
are public and how this is communicated. This APE does not address any other
questions of API design, e.g. public versus internal methods in classes, leaving
these for future APEs or future revisions to this APE.
```

nstarman (commented):
some comments.

APE_public.rst
nstarman:
```suggestion
1. A symbol is public if and only if all containing namespaces (packages, modules, classes,
```

APE22.rst
mhvk: This is slightly orthogonal to the issue here: really, one needs to set `__module__` to avoid this - registration just followed that, not a choice made by coders. Note that `numpy` in fact does this - which partially is annoying, since one no longer knows where the function is defined.

nstarman: For me the pain point was YAML serialization because I used ``f"{class.__module__}.{class.__qualname__}"``, which looks reasonable as code, but led to the problems described above. I would not have pushed the code as I originally did if, when making the tests, I had noticed a private path in the way.

APE_public.rst
nstarman: @astrofrog Done. Resolve?

APE_public.rst
nstarman: @astrofrog, resolve?

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+   the sub-rules:
+
+   - A symbol is *locally* public if it does not start with an underscore, and
+     internal otherwise, unless it is a dunder symbol (``__<...>__``).
+   - Dunder symbols (``__<...>__``) have their own public vs internal rules,
+     which are not determined in this APE.
+
+3. Public symbols must not be prefixed with an underscore.
+
+   - Internal symbols need not be prefixed with an underscore, but it is often
```
nstarman, line 257: Perhaps this should be strengthened.
"Internal symbols must be prefixed with an underscore, unless the modules in which the symbol is used are all private, making the prefix redundant"

mhvk:
Thanks for writing this! I think it is good to describe what formalizing the status quo would entail:
1. Any module from which users can import *must* have an `__all__` (i.e., all subpackages, and some of the sub-modules of `astropy.utils` [which are all documented to be only semi-public]).
2. Any module not listed in the `__all__` above is implicitly private.

Implementing this is very little work and would give consistency, without breaking anything.

A few more general points: 
1. One should not needlessly break user code, even if that code "incorrectly" imports from nominally private files. I try to heed @taldcroft's advice to really avoid that. Does being  "more correct" outweigh this?
2. For developers there is also considerable value to not changing things, because one knows by heart how to get to given files -- I open them by typing (of course) and what is the benefit of me having to retrain and add underscores that do not help tab-completion? 
3. If I think of subpackages I maintain, like `astropy.time` and `astropy.table`, essentially all files would start with underscores. In `astropy.units`, very little except for the unit-defining modules would be public. I find `astropy/units/_quantity.py`, `astropy/units/_function/_logarithmic.py` and `astropy/units/_quantity_helpers/_function_helpers.py` needlessly complicated.
4. We need to be realistic of the work involved. Conservatively, including reviews, I'd estimate 1 full month to actually do it (excluding time for this discussion, etc.). Plus an unknown amount of time of users dealing with broken scripts/packages that used to work.
5. While scipy and numpy may be moving (partially because they had real historical baggage), others do not seem to (e.g., pandas).

Overall, I'm fairly strongly against this. But being consistent within `astropy` is more important, so if the consensus is to move forward, I'll do my bit for the transition.

APE22.rst
mhvk: Yes, that one is an issue for examples and `sphinx` too - I really dislike that when we made `representations` a module (which was definitely a good idea!), we had to make so many updates to the docs. It would also be nice to use `regex` for things like [`__construct_mixin_classes`](https://github.com/astropy/astropy/blob/291414a3b600a12738762b7c50bfc6709f3cbf13/astropy/table/serialize.py#L25)

For me, this is still separate from the APE, since this has annoyed me amply already! But I can see that coming from an environment where files with leading underscores have more meaning, you could have avoided the issues.

nstarman:
> 1. Any module from which users can import _must_ have an `__all__` (i.e., all subpackages, and some of the sub-modules of `astropy.utils` [which are all documented to be only semi-public]).
> 2. Any module not listed in the `__all__` above is implicitly private.
> Implementing this is very little work and would give consistency, without breaking anything.

If I understand this point correctly, this is similar to point 2 of the implementation section.

```
2. **Add / update** ``__all__``. The ``__all__`` in each module will be updated
   to reflect phase 1. Any modules' missing ``__all__`` will have one added.
```

I think where this APE differs is that it aims to be **explicit** everywhere and not have anything be **implicitly** private (or public).
For both users and maintainers IMO [explicit is better than implicit](https://zen-of-python.info/explicit-is-better-than-implicit.html#2).
While the transition to this explicit state is somewhat arduous, as you noted in https://github.com/astropy/astropy/issues/15169, once accomplished, this APE attempts to make *remaining* in that state vey easy: through clear rules and CI checks. (Comments appreciated to make this this APE more clear / have better CI checks.)


> A few more general points:
> 
> 1. One should not needlessly break user code, even if that code "incorrectly" imports from nominally private files. I try to heed @taldcroft's advice to really avoid that. Does being  "more correct" outweigh this?

This is definitely true in general. For me, the benefits outweigh the costs for 3 reasons:

1. In the user code it's easy to switch the import from the incorrect private file to the correct public location since the public location is guaranteed to exist (it's guaranteed to exist because the user is using a public import, just importing it from the wrong location), e.g. `units.quantity.Quantity -> units.Quantity`.
2. For select items, or for everything, we can support the nominally private files using ``__getattr__``. See https://github.com/astropy/astropy/blob/main/astropy/cosmology/utils.py for an example. This allows for a deprecation period.
3. Maintainers *do* move around private files. Sometimes that breaks people's workflows. That sucks, so we try to minimize the damage. What if public vs private were obvious? **Then we'd never break anyone's workflow when we refactored private code** (so long as they used public API). No more damage. The very point that this APE would mildly break people's code is proof positive that we should make the change.

> 2. For developers there is also considerable value to not changing things, because one knows by heart how to get to given files -- I open them by typing (of course) and what is the benefit of me having to retrain and add underscores that do not help tab-completion?

True. Python's PEP-8 recommendations for how to structure a module to signal public vs private API is not wonderful for people using tab-completion in a terminal-based text editor. However, the tab-completion *problem* for developers is actually a *feature* for users, because they won't see private API.
Also, while VIM, nano, emacs, etc. are great, the limitation you mention does not apply to IDEs like Sublime Text, Nova, VSCode, etc. We should aim to support many ways to develop, of course, but...

> 3. If I think of subpackages I maintain, like `astropy.time` and `astropy.table`, essentially all files would start with underscores. In `astropy.units`, very little except for the unit-defining modules would be public. I find `astropy/units/_quantity.py`, `astropy/units/_function/_logarithmic.py` and `astropy/units/_quantity_helpers/_function_helpers.py` needlessly complicated.

So `astropy/units/_function/_logarithmic.py` would actually be `astropy/units/_function/logarithmic.py` (note `logarithmic` is not underscored).
Likewise `astropy/units/_quantity_helpers/_function_helpers.py` -> `astropy/units/_quantity_helpers/function_helpers.py`.

To the broader point, top-level modules, e.g. `astropy.time` with flat structures would have lots of underscores.

```
astropy/module/
    file1.py
    private_submodule/
        subfile1.py
```
becomes
```
astropy/module/
    _file1.py
    _private_submodule
```



> 4. We need to be realistic of the work involved. Conservatively, including reviews, I'd estimate 1 full month to actually do it (excluding time for this discussion, etc.). Plus an unknown amount of time of users dealing with broken scripts/packages that used to work.
> 5. While scipy and numpy may be moving (partially because they had real historical baggage), others do not seem to (e.g., pandas).
> 
> Overall, I'm fairly strongly against this. But being consistent within `astropy` is more important, so if the consensus is to move forward, I'll do my bit for the transition.

astrofrog:
So just to put a radically different idea out there, at least to consider as an alternative, what if we considered that public API was defined solely as objects defined in ``__all__`` and only in a subset of the ``__init__.py`` files - we could then add something to the effect of:

```python
warnings.warn('This module is private API and should not be relied on', PrivateAPIWarning)
```

at the top of most files in the code base, while in ``__init__.py`` files defining any public API we would not emit this warning and instead do e.g.:

```python
# This file defines public API

__all__ = ['SomeClass', 'a_public_utility']

with ignore_private_api_warnings():
    from .core import SomeClass
    from .utils import a_public_utility
```

The advantage of this approach is that

* we wouldn't need to worry about cluttering the code base with underscores
* we wouldn't need to break things for users at any point, but they would know they are using API that can break at anytime (in a sense it is a bit like a permanent deprecation phase)
* we wouldn't need to have a deprecation phase where we need to have both modules with underscores and compatibility ones without underscores, which will temporarily cause a lot of clutter in the code base
* this approach would arguably be more explicit and easier to learn for contributors rather than learning the rules about underscores and so on
* permanent warnings for users relying on private API

We could likely use pre-commit to ensure that modules either include a comment such as the one above (This module defines public API) or include a PrivateAPIWarning.

The main disadvantages I can see with this approach are:

* a potential small performance impact related to emitting and catching warnings at import time, but this may not be significant, and would have to be assessed.
* this would be a different approach to other packages in the scientific python ecosystem

I'm not strongly attached to this idea, but I do think it merits consideration.

mhvk:
@astrofrog - I certainly like the simplicity and the much smaller amount of work involved!

An even less intrusive version would of course be to have warning in the module docstring. But the advantage of your version is that the warning could include the suggestion to raise an issue if a user believes the piece of code they are importing should really be part of the public API.

If we edit docstrings, one advantage is that it would show up in `sphinx`, for those modules that we have typeset just because it keeps the organization more logical (e.g., https://docs.astropy.org/en/latest/units/ref_api.html#astropy-units-equivalencies-module). Obviously, though, this does not preclude adding the warning! (Or, indeed, having module names starting with an underscore)

pllim:
I am a little 👎 on module level warning. It is going to be a pain to filter it out everywhere internally.

astrofrog:
> this would mean I would have to catch the warning every time I used something from _utils because there would be no reason to ever add its functionality to an __init__.py.

I'm happy to also follow the numpy/scipy route but I just want to answer the above statement - it isn't correct that you would have to filter the warning everywhere internally. It doesn't matter what private methods/functions call what private methods/functions, at the end of the day all a user will do is import public API and we just need to filter the warning in the ``__init__.py`` file exposing that public API. Everything that happens internally is irrelevant. The only exception to this is if functions/methods dynamically import private modules at run-time rather than import time (which arguably we should minimize the use of).

nstarman:
> I do think it merits consideration.

It's good to circle back to this, as it was where this proposal started before the Astropy conference!
I guess it's a sign of how complex this issue can be. Were we to start writing Astropy from scratch today, I don't believe we'd be having this debate as we'd follow PEP-8 and typing practices and protect user's tab-autocomplete, etc. The issue, is that we are not starting tabula rasa and given our current muddled situation, how do we proceed? I guess philosophically I'm always inclined to go where we would if starting tabular rasa, and to try to chart a path, even if there is some pain.

Here are some of the reasons the proposal evolved from the conference.

> what if we considered that public API was defined solely as objects defined in ``__all__``

That is the central gist of this APE. ``__all__`` is the "ground truth" from which we derive our API.
However, while ``__all__`` has the immense benefits of being very explicit and instrospectable by developers, it is severely lacking for the needs of users. Users have 2 primary means to discover functionality within a library: by external channels — documentation, examples, tutorials, word of mouth — and by inspecting the code, most easily accomplished by tab autocomplete. I don't think it controversial to say that many of us tab-filter in IDEs to find a relevant function or even module. One notable issue with ``__all__`` is that since it starts with an underscore it is *not* found by tab completion. And even if it were, how many would consult it when tab-autocomplete gives a full list of available objects.

So while for developers ``__all__`` is excellent, for users we need to make their means of finding and understanding public API match ``__all__``. How do we do that? I'm happy we agree on the documentation, that we can create CI checks that ensure ``__all__`` matches the docs. But what about simple exploration of the library? Is `astropy.units.quantity` public? To a user it very reasonably appears to be. Like us, `numpy` and `scipy` have this problem in spades (`numpy.core` is ostensibly private, but if it's so visibly public, is it really private?).

Using ``__all__`` is important. It's step 2 of this APE! But alone it is a half measure.

> at the top of most files in the code base,

I'm unsure of this. Any use of ``warnings.simplefilter`` is liable to trip every single file in Astropy.

> we wouldn't need to worry about cluttering the code base with underscores

We'd have the same number of files, just with a single underscore added. I guess I don't agree that qualifies as clutter.

> - we wouldn't need to break things for users at any point, but they would know they are using API that can break at anytime (in a sense it is a bit like a permanent deprecation phase)
> -we wouldn't need to have a deprecation phase where we need to have both modules with underscores and compatibility ones without underscores, which will temporarily cause a lot of clutter in the code base

To address two points simultaneously. With deprecated compatibility modules we also wouldn't need to break things for users at any point during the deprecation period.
I agree that deprecation modules add clutter. It would be nice to avoid that, but at least the cleanup when the deprecation period elapse will be easy and only require deleting whole files.

> this approach would arguably be more explicit and easier to learn for contributors rather than learning the rules about underscores and so on

For three reasons, I actually think the opposite.
1. The first is that contributors rarely make new modules, so will not themselves need to deal with underscores.
2. The second is that contributors overwhelmingly will have used terminal editors or IDEs with tab-autocomplete. On all systems I'm aware of tab-autocomplete filters underscore-prefixes (unless the user specifically starts with an underscore before pressing tab). In effect contributors will have had years of practical experience. (not that they'll need it, because of point 1).
3. The third reason isn't a new point as you brought it up in the list of disadvantages. To reiterate then,  in this APE we would be adopting a common standard, not making a new one. With a common standard *everyone* who has contributed to any other code base with will already know the rules. With a new standard, *no one* will have prior familiarity.

> permanent warnings for users relying on private API.

This part I like! But from an above point, any use of ``warnings.simplefilter`` is liable to trip every single file in Astropy.

mhvk:
> But what about simple exploration of the library? Is `astropy.units.quantity` public? To a user it very reasonably appears to be. Like us, `numpy` and `scipy` have this problem in spades (`numpy.core` is ostensibly private, but if it's so visibly public, is it really private?).

This to me is the strongest argument for change, but it would seem to require only one part of what is in this APE: have `__all__` in `__init__.py`. I checked by comparing `astropy/units/__init__.py` (which does not have it) and `astropy/coordinates/representations/__init__.py` (which does have it -- thanks to you!): I find that (in ipython):
```
import astropy.units as u, astropy.coordinates as coord
u.qu<TAB>  # Indeed, I see `quantity`, which a user does not have to see
coord.representations.s<TAB>  # I do not see `spherical`
```

So, we can use `__all__` in `__init__.py` to make clear to the user what is private; this part I'm fine with (even though for units it may be a bit of a pain, given how many entries it has -- arguably making the probability really small that someone will actually discover things by chance by tab-completion *and* then import something that should be private and which we actually change).

But this lessens the argument for the rename: We're now left with people exploring in editors, and those we can just as easily get with documentation (module-level docstrings would be good to have regardless, and now that the module is already "private", the docstring can actually be aimed at developers, as I've hoped would happen a *long* time ago: #8930).

So, I remain unconvinced about the APE as a whole, a lot of effort for a problem that, in practice, just has not occurred much (for numpy, it also has not happened all that much -- though more than for us; I think the driver there is mostly that they really wanted to reorganize their structure, which has lots of historical baggage, and thus in a way they are starting from a more or less blank slate -- note this is happening for numpy 2.0, where they are allowing API breaks!).

I am much happier with just formalizing the existing policy, by documenting it and adding `__all__` to every subpackage's `__init__.py`, and adding notes to modules. This seems to get us 99% of the way, will cost substantially less review time (the real effort in any of this) and will have no cost to finger memory of existing maintainers.

Since making `__all__` in each `__init__.py` seems less controversial, perhaps we can just get started with that?

nstarman:
>  I find that (in ipython):

Depends strongly on the editor. PEP-8 offers multiple means by which to define public versus private API. IPython command line appears to use `__all__`, Jupyter notebook also uses underscores for modules, which is also in PEP-8. 
This original proposal was to do just `__all__`, which is still an improvement over our current situation. But in discussion we found, as this screenshot shows, that `__all__` is only a partial measure. There should be one "truth", but that requires `__all__`, editing the docs, and module names. I'm happy that we `__all__` agree 😄 on the first two. The module names is clearly more contentious.

CleanShot 2023-08-24 at 16 48 55@2x

mhvk:
Indeed, we agreed on `__all__` for `__init__.py`, happy to just get going with that (maybe `units` last? The rest is *far* more obvious). 

I'm still less sure about what we do for the modules. I'm afraid it's gotten too late here to actually look at the APE, so if this is nonsense, just ignore, but I still really dislike `__all__ = []` too. Instead, as we have right now, `__all__` can give immediate information for someone opening the file about what in it is actually exposed (even if at a different level for the user). It also makes constructing `__all__` in `__init__.py` easy (or a good auto-sanity check if imports become explicit), and helps enable generation of documention for modules like `units.equivalencies`.

p.s. Will be on holidays for the next bit...  At least, that will give others a chance to pipe in!

APE22.rst
```diff
@@ -0,0 +1,440 @@
+Astropy's API Policy on Modules and their Contents
+---------------------------
+
+author: Nathaniel Starkman
+
+date-created: 2013 November 5 <replace with the date you submit the APE>
```
neutrinoceros, line 6: I guess this date can be set already ?

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+
+
+.. _description_of_the_problem:
+
+Description of the Problem
+--------------------------
+
+Pure Python has no language-level means to establish and enforce separations
+between public versus internal interfaces -- all code objects can be accessed.
+This is in contrast to languages like C++ and Java, which have language-level
```
neutrinoceros, line 44: very minor tweak: the "and" makes it sound like those are the only two languages with a private VS public boundary :)
```suggestion
This is in contrast to languages like C++ or Java, which have language-level
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+constructs to establish public versus private interfaces. Differentiating
+between public and internal interfaces is important for everyone, users and
+developers alike, to know what interfaces are stable and supported, and which
+are not. Consequently, Python has developed conventions for communicating what
+is public versus internal. Nominally there are three means (discussed more
+formally later):
+
+1. **Underscore Prefix**.  By convention, names that begin with an underscore
```
neutrinoceros, line 52: technicality: name mangling rules are confusing, let's stay away from them if we can (see https://docs.python.org/3/tutorial/classes.html#private-variables)
```suggestion
1. **Underscore Prefix**.  By convention, names that begin with a single underscore
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+internal interfaces, we now decide what that approach should be. The approach
+should fulfill a few criteria:
+
+1. Align closely with established conventions, so that it is familiar
+   to users and developers, and hopefully detected by assistive tools.
+2. Be unambiguous and easy to understand.
+3. Be straightforward to implement and maintain, so that it is not a
+   burden on developers.
+
+We start by considering established practice, for criteria 1, and how to ensure
```
neutrinoceros, line 128:
```suggestion
We start by considering established practice, for criterion 1, and how to ensure
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+     begin with an underscore to be included in the interface.
+
+This is consistent with PEP 8 and SciPy, but it does not mention the
+documentation. The Python typing guide reinforces the idea that the ``__all__``
+attribute is the authoritative source of what is public vs internal.
+Importantly, the ``__all__`` attribute is a module-level attribute, so it only
+applies to the module in which it is defined. We will refer to this as "locally"
+public. This means that a module can define an ``__all__`` attribute but if the
+module itself is not public, then anything in the ``__all__`` attribute cannot
+be publicly accessed from outside the module.
```
neutrinoceros, line 220: I'm not sure I understand this sentence. Does this imply that any private module should set `__all__ = []`, or does it mean that `__all__`, within a private module, defines an interface that's meant for internal use only ? (now that I wrote it down, the latter seems way more likely, but I think there's room to improve the phrasing)

APE22.rst
neutrinoceros: I'm in favour of this rephrasing !

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+3. Public symbols must not be prefixed with an underscore.
+
+   - Internal symbols need not be prefixed with an underscore, but it is often
+     recommended. An example of when it is not needed are symbols defined in
+     internal modules, making the prefix redundant.
+
+4. Public symbols must be documented.
+
+   - If an internal symbol is documented (which is not recommended), it must be
+     obviously, explicitly, (and preferably repeatedly) noted as internal.
```
neutrinoceros, line 264: "obvious" may sometimes give off a condescending vibe to the reader, let's avoid that
```suggestion
     unambiguously, explicitly, (and preferably repeatedly) noted as internal.
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+      spam/  # implicit namespace package
+
+         ham.py::
+            __all__ = ['eggs']
+
+            def eggs():
+               pass
+
+In this case, ``bar`` is public in ``foo``, and  ``foo`` is public in ``src``,
+so ``src.foo.func`` is public. This is public. In contrast, while ``bunc`` is
```
neutrinoceros, line 304:
```suggestion
so ``src.foo.func`` is public. In contrast, while ``bunc`` is
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+primarily by the documentation, but also by tab-completion within an interactive
+session.  In the implementation of this APE to establish a *formal* API, we aim
+to minimize changes to the *de facto* API. This will be accomplished in the
+following 5 phases:
+
+1. **Snapshot the documentation**. As of the adoption of the APE a snapshot of
+   the documentation will be saved. This will be used to determine what is
+   currently public and what is not. Until the APE's adoption is complete, this
+   snapshot is authoritative, e.g. dictating what must be added and removed from
+   ``__all__``, for deciding what must undergo a deprecation process, etc.
```
neutrinoceros, line 349: I feel like we lack a prescription for what to do with symbols that are added post-snapshot here

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+   - Add a SciPy-like section to the developer documentation explaining the
+     public vs internal rules.
+   - Fix links to always point to the public interface, not the internal
+     interface. (This will presumably require a similar implementation as used
+     in NumPy to change the ``__module__`` attribute of many objects.)
+   - Add a reminder to the maintainer checklist that to be public a symbol must be
+     in ``__all__`` and documented.
+   - Clearly state if a documented object is actually private.
+
+4. **Add a pre-commit bot**. This will ensure that the ``__all__`` attribute is
```
neutrinoceros, line 369: ?
```suggestion
4. **Add a pre-commit hook**. This will ensure that the ``__all__`` attribute is
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+   code to continue to work, while encouraging people to fix their code to use
+   the public interface.
+
+
+Alternatives
+------------
+
+**We do nothing:**
+
+This is the status quo.  It is not a good option because it is not solve the
```
neutrinoceros, line 406:
```suggestion
This is the status quo.  It is not a good option because it does not solve the
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+issue.  The aforementioned problems of not i) knowing what is stable and
```
neutrinoceros, line 407:
```suggestion
issue.  The aforementioned problems of not knowing what is stable and
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+supported, and what is not, remain.
+
+**We allow** ``__all__`` **to be optional:**
+
+This is not great.
+
+The only time this might be good is when a module has dynamic symbols from a
+`PEP 562 <https://peps.python.org/pep-0562/>`_ module-level ``__getattr__``
+method. However, if it truly dynamic then it cannot be statically analyzed, which
```
neutrinoceros, line 416:
```suggestion
method. However, if it is truly dynamic then it cannot be statically analyzed, which
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+is undesirable for other reasons. If it is not truly dynamic, then the "dynamic"
+symbols can be added to ``__all__`` and an ``__init__.pyi`` file used to
+communicate the public interface to static analyzers.
+
+**We make the documentation the authoritative definition of the API.**
+
+Good code is well documented in the code. Beyond code comments and docstrings, what is
```
neutrinoceros, line 423: ?
```suggestion
Good code is well documented. Beyond code comments and docstrings, what is
```

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+public versus internal is an important aspect of the code, and should be
+communicated in the code itself. It is vitally important that the code and
+user-facing documentation are consistent, and the user-facing documentation
+should reflect the code.
+
+**We make tab-completion the authoritative definition of the API.**
+
+This is not a good option because it does not work with implicit namespace
+packages nor `PEP 562 <https://peps.python.org/pep-0562/>`_ module-level
+``__getattr__`` methods.
```
neutrinoceros, line 433: just curious, is this mechanism actually used in astropy or in any affiliated package you know of ?

neutrinoceros (commented):
Overall this looks extremely solid and a well motivated solution to a real problem, so I 100% get behind this APE. Here are a couple, mostly superficial, remarks and suggestions.

APE22.rst
nstarman: Yes, the latter. If you can't publicly navigate to a module, it doesn't matter what's "public" within that module. The full path must be public.
The issue with relying on this alone is what you discovered in that Issue (trying to find the one) where we have modules that look public (because they are not underscored) so it's possible to construct an import path that looks totally public, but actually isn't.

APE22.rst
nstarman: They should follow the rules! If it's public, make it so.
This is just saying we're not going to debate what already exists until we rectify the current ad-hoc standard with this one. But new things should follow the new standard.

APE22.rst
nstarman: 👍

APE22.rst
nstarman: No, I mean in the code itself. E.g. we build documentation from code docstrings because we want the docstrings to be available in the code itself. If that weren't important all our docstrings would just be in the docs.

APE22.rst
nstarman: - Implicit namespace packages: not any affiliated package I know, though I have some personally. Super useful. I wish tab-complete were better.
- ``__getattr__``, yeah in a couple places! E.g in `astropy.cosmology.realizations`. Also check out https://github.com/scientific-python/lazy_loader. I would very much like for Astropy to implement lazy loading and speed up initialization time.

APE22.rst
nstarman: How about...

It is recommended, though not enforced, that internal symbols should be prefixed by a single underscore. An example of when the underscore is not necessary is for internal symbols defined in an internal module, since the prefix would be redundant (e.g. ``_bar`` in ``_foo.py`` can be just ``bar``).

nstarman (commented):
Thanks @neutrinoceros!

APE22.rst
nstarman: Agreed we can leave it out of this APE. I think the point I made "If, when making the tests, I had noticed a private path I would have fixed the problem from the start." means adopting this APE will automatically force most of these I/O issues to be fixed

nstarman:
@eerovaher @WilliamJamieson @taldcroft @eteq @pllim, I would appreciate some more eyes on this APE. As we often follow the lead of NumPy I think this APE is very timely given their refactor to quite nearly follow this APE.

APE22.rst
neutrinoceros: >  If you can't publicly navigate to a module, it doesn't matter what's "public" within that module.

I think that's what confused me; since it doesn't matter, why should we care what in a private module's `__all__` ? Wouldn't it be simpler to recommend/rule that `__all__` must be set to `[]` in private modules ?

> The issue with relying on this alone is what you discovered in that Issue (trying to find the one)

For reference, I believe you meant to link https://github.com/astropy/astropy/issues/15666

APE22.rst
neutrinoceros: Agreed, but I think I meant to suggest that it should be explicitly mentioned in the APE too.

APE22.rst
neutrinoceros: Oh, now I see what you meant !

APE22.rst
neutrinoceros: Thanks ! My question was specifically about implicit namespace packages as I have first hand experience with module-level `__getattr__` and I know it's extremely useful. Nonetheless, I'm glad my question leaked to this topic too because I'm also interested in lazy loading and startup time. Actually astropy already has it a lot better than a lot of packages I've seen, but I'm sure we can improve it still. Anyway this is a digression, I just wanted to let you know that I'm interested in that topic as well. Feel free to resolve this thread !

APE22.rst
neutrinoceros: sounds good to me, thank you !

nstarman:
@pllim, it can definitely wait in the sense that nothing will break further. And also I'm not sure who would have the time to lead this effort if this APE were accepted sooner.

pllim:
@nstarman , apparently APE 22 is taken by #87 . You will have to rename your file... but at this point, I am not sure to what. Maybe @eteq can advise. See https://github.com/astropy/astropy-APEs/pull/87#issuecomment-1915413288

APE22.rst
astrofrog: It might still be useful to be able to indicate in a private sub-module which things are meant to be importable in other parts of astropy, so ``__all__`` could be useful in that way rather than always setting to ``[]``.

nstarman: Agreed! This is demonstrated in lines 281-287

nstarman: In examples of good module layouts.

nstarman: @astrofrog, see https://github.com/astropy/astropy/pull/15109/files as an example doing what you're talking about. I agree, this is super useful.

mhvk (commented):
I looked at this again and must admit I find the prospect of yet another major refactoring rather off-putting, especially as it aims to solve what seems a minor problem; certainly, at the user level we have had few if any complaints as we (re)move code, suggesting that users have no issues finding things they need and not rely (overly) on private API. 

We should really think carefully about the costs of these refactoring exercises. I'd estimate the implementation, *including reviews*, is going to be of order 10^2 hours (certainly more than 10^1, surely less than 10^3, though perhaps 10^1.5 is more realistic). Do we really want to spend the equivalent of $ 5...10k on this? (Though I guess we have already spent several 10k $ equivalent on ruff "rules"; has that been really worth it? What else could we have done with that amount of expensive developer time?)

I also think we (and thus the APE) should consider seriously alternatives that start from current practice, that anything under the basic subpackages like `astropy.coordinates` should be taken to be private unless explicitly stated otherwise, and just try for formalize that in a way is good enough, without worrying too much about perfection (PEP 8 is explicit about why that is a bad idea). Can this not be made explicit without much effort (and no renaming)? E.g., how about the following alternative, with three nicely incremental steps:
1. We explicitly document current practice that everything under subpackages is private and add a corresponding comment in all their top level `__init__.py` files (making appropriate exceptions in `io` and `utils`).
2. We add `__all__` to all submodule `__init__.py` files that include the public items, including public submodules of the subpackages.
3. We slowly add `__all__` to the rest of astropy, to indicate to ourselves which parts are meant to be used outside a given module.

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+- Is ``_func`` public?  It is prefixed with an underscore, but in ``foo.__all__``. 
+- Is ``_bar`` public? It is prefixed with an underscore, but in ``src.__all__``.
+- Is ``bunc`` public? It is in ``_bar.__all__``, but what if ``_bar`` is private?
+- Is ``_bunc`` public? Probably not, but what if it's documented?
+- Is ``baz`` public? It does not start with an underscore, but is not in ``src.__all__``.
+- Is ``qux`` public? It does not start with an underscore and is in
+  ``baz.__all__``, but what if ``baz`` is private?
+
+
+Astropy partially implements all three of the conventions: some private classes
```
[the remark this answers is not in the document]
eerovaher: >  astropy actually has a clear, basic definition of what is public, which is everything directly in a submodule like astropy.coordinates.

Do you know where this is written down?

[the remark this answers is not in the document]
mhvk: I think it pretty obvious from the documentation - see the reference sections of `time`, `coordinates`, `timeseries` - though of course the moment I look at others, I see that `nddata` is not quite as good, and `modeling` has definite public modules (though clearly indicated as such).

I do stick with my main point that we actually have had very few cases where users were confused about where to import things from.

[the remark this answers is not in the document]
eerovaher: It was not too difficult to find counter-examples from [`astroquery`](https://github.com/search?q=repo%3Aastropy%2Fastroquery+%2Ffrom+astropy.table.table+import+Table%2F&type=code), [`specutils`](https://github.com/search?q=repo%3Aastropy%2Fspecutils+%2Ffrom+astropy.nddata.nduncertainty+import+StdDevUncertainty%2F&type=code), [`photutils`](https://github.com/search?q=repo%3Aastropy%2Fphotutils+%2Ffrom+astropy.nddata%5C.%5Ba-z%5D*%2F&type=code) and [`ccdproc`](https://github.com/search?q=repo%3Aastropy%2Fccdproc+%2Ffrom+astropy.units.quantity+import+Quantity%2F&type=code), and these are all coordinated packages.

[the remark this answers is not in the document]
mhvk: Are you suggesting these should all be converted to a particular scheme too?Then we better get more feedback on this APE  :smiley_cat: I doubt I'm the only one who feels time is better spent elsewhere (and sadly this is not something that can be done without impacting everybody).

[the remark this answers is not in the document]
eerovaher: I presented a few counter-examples from coordinated packages to the claim 

> ...that we actually have had very few cases where users were confused about where to import things from.

We have more control over coordinated packages, so if those are being confused about importing then I'm guessing the confusion is even worse in non-coordinated packages.

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+
+4. Public symbols must be documented.
+
+   - If an internal symbol is documented (which is not recommended), it must be
+     unambiguously, explicitly, (and preferably repeatedly) noted as internal.
+
+
+Let's consider an example::
+
+   src/
```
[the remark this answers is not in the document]
nstarman: Note to self: difficulties in public/private API definitions in https://github.com/astropy/astropy/pull/16519

APE22.rst
```diff
@@ -0,0 +1,440 @@
...
+Alternatives
+------------
+
+**We do nothing:**
+
+This is the status quo.  It is not a good option because it does not solve the
+issue.  The aforementioned problems of not knowing what is stable and
+supported, and what is not, remain.
+
+**We allow** ``__all__`` **to be optional:**
```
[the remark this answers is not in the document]
nstarman: Note to self: IDK when they did this, but `scipy` now has underscore-prefixed module names everywhere. They fixed their public/private problem. e.g. https://github.com/scipy/scipy/tree/main/scipy/fft

nstarman:
Note: if we take queues from "upstream" packages, scipy now has underscore-prefixed names for modules.

pllim:
This was discussed as part of "State of APEs" at Coordination Meeting 2024. I think reactions were mixed and I cannot see any clear action items on how to move this forward (or if we should).

astrofrog:
One idea I raised was that at the very least if we cannot reach consensus on changing current code, we should see if we can agree on rules for any new code?

pllim:
Re: https://github.com/astropy/astropy-APEs/pull/85#issuecomment-2183183433

For completeness, my response to that idea in the meeting was that if it is only recommendation for new code, I do not think we need an APE, but rather we can modify the dev docs.

mhvk:
There was indeed no consensus on the underscore prefixes, in large part because, contrary to what I thought at least, things like `from astropy.units.quantity import Quantity` were widespread in other github repositories. Hence, changing to underscores is guaranteed to break quite a bit of downstream code, and it is not clear this is worth it.

There also seemed to be consensus that in the end the documentation should be the ultimate arbiter, since that is what users would normally see (and the mistake of documenting the `quantity` submodule is probably at least partially to blame for the wrong usages...). So, it remains a good idea to ensure we document what is public and private, but start incrementally, as suggested in https://github.com/astropy/astropy-APEs/pull/85#pullrequestreview-2012234665:

> 1. We explicitly document current practice that everything under subpackages is private and add a corresponding comment in all their top level `__init__.py` files (making appropriate exceptions in `io` and `utils`).
>  2. We add `__all__` to all subpackage `__init__.py` files that include the public items, including public submodules of the subpackages.
>  3. We slowly add `__all__` to the rest of astropy, to indicate to ourselves which parts are meant to be used outside a given module.

Finally, I'd say there was no consensus either on new vs old code, or subpackages doing different things, with the latter having the advantage of maintainers being able to set a policy they feel is best, but the disadvantage that then there is no package-wide logic anymore at all, while currently there is (with `cosmology` the only exception).

nstarman:
> Hence, changing to underscores is guaranteed to break quite a bit of downstream code, and it is not clear this is worth it.

`from astropy.units.quantity` is private. IMO we might as well break usage of private code in one fell swoop and then not (ever?) again rather than do it piecemeal as publicly-visible-private-code is changed over the years. I would find that less disruptive.

mhvk:
> `from astropy.units.quantity` is private. IMO we might as well break usage of private code in one fell swoop and then not (ever?) again rather than do it piecemeal as publicly-visible-private-code is changed over the years. I would find that less disruptive.

It is private, indeed. But the feeling at the coordination meeting was that breaking people's code for code style purity is too big a price to pay. And it is not likely we would ever define `Quantity` in another place than `astropy.units.quantity`. I also think the issue may be moot sooner or later, since I do think there will eventually be a general units/quantity package that we are going to be based on (hopefully by combining our units machinery with `Quantity 2.0`!). 

But for astropy as a whole, continuity and consistency are important too. But nothing stopping us from making it clearer what is public and not, by defining appropriate `__all__` and ensuring that, unlike for `units`, "private" modules do not appear in the documentation unless strictly necessary, and then with a clear docstring that states why they are included.

nstarman:
> since I do think there will eventually be a general units/quantity package that we are going to be based on (hopefully by combining our units machinery with Quantity 2.0!)

🎉. That would be excellent.

> But nothing stopping us from making it clearer what is public and not, by defining appropriate `__all__`

A great thing to do, no matter the outcome of this APE.

---

When first proposed, one of the counter-arguments was that "upstream" libraries haven't done this. But now both `numpy` and `scipy` have basically done this (slightly different implementations).
And they did it in one fell swoop (numpy 2, recent scipy), so that users didn't suffer multiple falls from repeated swoops.
I'm just wondering why we're different. Same problem, similar solution?

When I first wrote this APE it was to make an argument "why we should do this". Now that our upstreams have done the same thing, IMO the argument shifts to "why aren't we doing this?". Prima facie we should. 

Documentation is important, but it is most certainly not how any of our upstream libraries *define* their public API. The point of documentation is to document what is public, not to make it public. Just like we generate documentation from docstrings, prioritizing that the code is "self-documenting" so too does Python, our upstream libraries, and most everyone else who cares about the public/private distinction makes it so that public/private is a product *of* the code, not imposed *on* it. 
And this understanding is intrinsic to how we've built tooling for Astropy, like `sphinx-automodapi`: it looks at `__all__`.

neutrinoceros:
> And they did it in one fell swoop (numpy 2, recent scipy), so that users didn't suffer multiple falls from repeated swoops.
I'm just wondering why we're different. Same problem, similar solution?

I see 3 options here
a) moving private code to private modules over one swoop
b) moving private code to private modules piecewise
c) do nothing[^1]

I agree with @nstarman that `a>b`. I also agree with Marten `c>b`. The remaining question is how to compare `a` VS `c`.

Since, as you guys pointed out, we may eventualy *have to* move part the private code from `astropy.units` in response to Quantity 2.0 becoming a dependency, why not use *that* event as a pivot to switch our strategy to `a`, and keep status quo (`c`) in the mean time ?

[^1]: I'm only speaking about moving modules/members around here. Defining `__all__` is a separate discussion and one that seems more consensual anyway.

mhvk:
Numpy was in a different state, with, e.g., some parts of `np.lib` being public, while other parts were not, so there was more urgency than we have. Even so, they held off to numpy 2.0, where a lot of other stuff was broken too.

Overall, @neutrinoceros made the right list, and I guess the conclusion in the coordination meeting was that `c>a` at the present time. At a time when there is a larger API change (as Quantity 2.0 would be), the conclusion may well be different.

In the meantime, there's nothing stopping us from incrementally ensuring that docstrings and `__all__` are all consistent and clear.

pllim:
Also keep in mind that NumPy has the backing of private industry (e.g., NVidia). Astronomy does not. I have started seeing pipelines pinning `numpy<2` privately just because they have larger fish to fry and no time to deal with breaking API here and there. To them, calibration accuracy and stability is way more important than whether `astropy.units.quantity` is private or not. We have to keep our main "customers" in mind and they are not "big money".

astrofrog (commented):
One thing we could consider doing in any case would be for common cases of misuse (such as astropy.units.quantity) we could use batchpr to search through GitHub and open PRs to fix these incorrect imports.

It's also worth considering another option d, which is to do The Right Thing :tm: for new sub-packages and code, to at least not increase the problem.

pllim:
xref https://github.com/astropy/astropy/issues/17589

kbwestfall:

Your chief APE editor is Erik Tollerud (@eteq).

nstarman:
@eteq I've updated this APE.
In particular, `astropy/cosmology` now implements this APE and serves as good example. Also numpy and scipy basically implement this APE, which they hadn't yet done when this APE was written.

PR #90, opened 2023-12-19
Update APE20 for Black formatting not just the Black tool

taldcroft:
Since APE-20 was accepted in 2022, another auto-formatting tool has become available that also implements the default Black format and has gained popularity. This development highlights that APE-20 was written with a somewhat short-sighted perspective -- it focuses on using the `Black` tool instead of adopting Black formatting using whatever Black-compatible auto-formatter is appropriate.

Changing the entire PR to reflect this change is not that helpful, but instead this PR adds a few words in the abstract to reflect this broader perspective. Adoption of this PR is technically a lien on changing the astropy Black-format CI check to another tool.

### DIFF
--- a/APE20.rst
+++ b/APE20.rst
@@ -1,9 +1,11 @@
 .. _Black: https://github.com/psf/black
+.. _ruff: https://docs.astral.sh/ruff/
 .. _Django DEP-0008: https://github.com/django/deps/blob/main/final/0008-black.rst
 .. _pandas pre-commit: https://pandas.pydata.org/docs/development/contributing_codebase.html#pre-commit
 .. _coding style: https://docs.astropy.org/en/latest/development/codeguide.html#coding-style-conventions
 .. _PEP8: https://peps.python.org/pep-0008/
 .. |Black| replace:: *Black*
+.. |ruff| replace:: *ruff*
 
 Formatting Code with *Black*
 ============================
@@ -12,7 +14,7 @@ authors: Tom Aldcroft, Erik Tollerud
 
 date-created: 2022 July 13
 
-date-last-revised: 2022 September 23
+date-last-revised: 2023 December 19
 
 date-accepted: 2022 September 23
 
@@ -25,7 +27,14 @@ Abstract
 --------
 
 This APE proposes that the `astropy core package
-<https://github.com/astropy/astropy>`_ adopt |Black|_ for formatting code.
+<https://github.com/astropy/astropy>`_ adopt Black formatting for the code. Black
+formatting refers to the default code style defined by the |Black|_ code formatter.
+
+Since this APE was accepted in 2022, the Black code format has been implemented by other
+tools, most notably |ruff|_. In the text below we will use the term |Black| to refer to
+|Black| or any other tool that provides robust auto-formatting in the Black code format.
+Beyond the initial migration of the code base to black formatting, the tool(s) used
+to maintain |Black| formatting is not a requirement of this APE.
 
 Detailed description
 --------------------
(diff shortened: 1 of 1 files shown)


neutrinoceros (approved):

pllim:
Thanks! I will put this on agenda for dev telecon.

mhvk (approved):
Makes sense to me, thanks for the update.

WilliamJamieson (approved):
I agree with this change in principle.

pllim:
So far no one objects, so I added this to CoCo agenda for 2024-01-22 when we next meet. Thanks for your patience!

astrofrog:
If uncontroversial this could be put to the full voting members list for a vote instead of waiting for the coordination meeting which might only have a fraction of voting members.

dhomeier:
Fully agree with making this tool-independent. Perhaps the heading should also be changed?
> Formatting Code in *Black* Style

And since the modified text now becomes a little bit more specific about what is _the Black code style_, it would be good to note that its exact incarnation to be followed is decided per case, sort of.
In particular since Ruff has some history of pulling in whatever formatting or linting rules can be found as (optionally) available ruleset, I am wondering if its "standard" format will always stay in sync with Black itself (but even when we first adopted it, I think Black itself had somewhat different rulesets in stable and developer versions).

pllim:
> put to the full voting members list for a vote

I guess https://github.com/astropy/astropy-APEs/blob/main/APE0.rst#responsibilities-and-rights does not rule that possibility out, but isn't that slower? Someone has to set up the heliovotebot (or whatever it is called) and then go through the 2 weeks voting period, get the tallies, and so on?

astrofrog:
> I guess https://github.com/astropy/astropy-APEs/blob/main/APE0.rst#responsibilities-and-rights does not rule that possibility out, but isn't that slower? Someone has to set up the heliovotebot (or whatever it is called) and then go through the 2 weeks voting period, get the tallies, and so on?

Yeah sorry I misremembered the process and this is indeed overkill. But maybe we can even reach a consensus before the coordination meeting.

mhvk:
This really seems super uncontroversial - the APE really was about settling on a specific style, with surely the tool an implementation detail. I'd consensus has already become clear!

taldcroft:
From what I can see this has been uncontroversial and there is consensus. Jan 22 will be here before we know it, time flies! Is there a PR ready or in-work for migrating to `ruff format` in astropy?

neutrinoceros:
There isn't. Can I volunteer ?

eerovaher:
My plan (that I mentioned at a developer telecon a couple of months ago) was to have a look at the lines in `astropy` that Ruff formatter would edit and to see if they could be rewritten so that both Black and Ruff formatter would be happy with them. The goal was to reduce the size of the patch that would be required to switch from Black to Ruff formatter, but also to try to improve odd-looking code if possible.

The part of the transitioning that requires the most effort to implement and to review is updating the developer documentation, and my plan has nothing to do with that.

neutrinoceros:
Ok, then I can do it, as I think I should read that doc anyway !

eerovaher (approved):
I am in favor of `astropy` switching from using Black to using Ruff formatter. We are using Ruff for linting anyways.

dhomeier (approved):
Looks good now, thanks!

eteq (approved):
Happy with this too!  I think if a third one shows up we might want to re-word this to not have to update it every time a new tool comes out but this might be the last one so :shrug:

ceb8:
I also approve.

eteq:
The CoCo today unanimously approved  this, will merge shortly!

eteq:
Hmm, I think someone else might own the Zenodo entry, so it has *not* been updated yet.  Trying to track down who it is...

pllim:
Oh no... maybe it was me. 😅

pllim:
OK never heard back from Erik, so if it is indeed under my own Zenodo account, I am just going to update that same record. I don't have time to mess around with this stuff.

PR #93, opened 2024-01-08
Fix link to voting members

### LINKED ISSUE #88: Broken link in APE0
There are many links to "Voting Members", with the form
```
`Voting Members <votingmembers>`_
```
but the file `votingmembers` doesn't exist. Has this been moved elsewhere?

weaverba137:
This PR fixes #88.

I had to experiment with several different types of internal RST links. I suspect that GitHub may be interpreting RST somewhat differently from, *e.g.* Spinx, but I found a type of internal link that does seem to work when rendered on GitHub.

### DIFF
--- a/APE0.rst
+++ b/APE0.rst
@@ -41,10 +41,8 @@ users, and even potential participants in the broader astronomical and
 scientific software community. All have a vested interest in the success of the
 Project, so all governance in the end belongs to this community in its entirety.
 The governance Charter described below provides levels of delegation of that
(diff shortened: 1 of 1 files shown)


weaverba137:
@pllim, I think this is ready to go.

pllim (approved):
LGTM. Your fix is more elegant than what I would have done myself, so thanks!

Since this is strictly a bug fix and does not change the actual meaning of the content, I am going to just merge.

PR #94, opened 2024-01-19
APE: An initial roadmap for static type checking

### LINKED ISSUE #92: Create an APE that describes the extent to which Astropy should be typed and the process for doing so
There has been a fantastic discussion in https://github.com/astropy/astropy/issues/15170 about the adding type hint annotations to Astropy.  Given the magnitude of this change, we would probably benefit from an APE that answers questions like:

 - To what extent should Astropy be typed? Should we strive for [mypy](https://mypy-lang.org/) regular mode or aim for mypy strict mode?
 - How do we deal with existing mypy errors?
 - What is the process for adding type hint annotations to Astropy?
 - How would we go about adding mypy to the suite of continuous integration checks?
 - How do we ensure that new code is typed?
 - Should we require mypy strict mode for all new code or modules?
 - How do we deal with dynamically created content? For example, do we add a `.pyi` type stub file for the individual units in `astropy.units`?

We could also add some helpful information, like which tools we could start off using. We probably don't want to be overly prescriptive, since the tooling landscape will probably change significantly in a few years.

namurphy:
This PR will add an APE that proposes a path forward for enabling static type checking in Astropy's continuous integration suite.  If anyone is interested in being a co-author, please let me know!

The process would be more or less to:
 - Enable mypy in CI, but only check a single file at first
 - Gradually enable static type checking in one or two subpackages (`astropy.cosmology` and `astropy.units`?)
 - Add a section to the developer docs on type hint annotations and static type checking
 - Allow subpackage maintainers to gradually enable static type checking in more subpackages

It's currently in an early draft state, and might be best though of as being in the brainstorming stage 🧠 🌩️.  Big picture feedback would be especially appreciated!  In particular, we will need to address the really valid concerns and considerations that have been brought up in https://github.com/astropy/astropy/issues/15170.

I'm also hoping to avoid being over-prescriptive, so as to give us flexibility and avoid the need to amend this APE when ruff implements a full static type checker. 🙃 

Closes #92.

### DIFF
--- a/APE25.rst
+++ b/APE25.rst
@@ -23,7 +23,36 @@ type checking to Astropy.
 Detailed description
 --------------------
 
+Type hint annotations allow us to specify the expected data types of
+function arguments, return values, and variable assignments. For
+example,
 
+.. code-block::
+
+   def f(x: int) -> float:
+       ...
+
+indicates that ``x`` should be an ``int`` and the return value should
+be a ``float``. Type hints are not enforced by Python during
+runtime. Type hints can be used by static type checkers to find
+problems with the code and by Jupyter notebooks and integrated
+development environments (IDEs) to help with code completion. Type
+hints also serve as a form of documentation in function and method
+signatures.
+
+For most of its history, type hint annotations have not been applied
+within Astropy's source code (though annotations have been used to
+indicate the expected units or physical types of function arguments
+and return values). Only recently have type hint annotations begun to
+be applied to Astropy. However, no mechanism is in place to check the
+correctness and validity of type hint annotations. ...
+
+Most guides on adding type hint annotations and enabling static type
+checking recommend starting small, running mypy consistently,
+prioritize annotating widely used imports, add type hint annotations
+for new code, gradually increase the fraction of the code base that is
+checked my mypy, and gradually increase the strictness of the checks
+performed by mypy.
 
 .. This section describes the need for the APE.  It should describe
    the existing problem that it is trying to solve and why this APE
@@ -31,8 +60,8 @@ Detailed description
    new functionality would be used and perhaps some use cases.
 
 
-Benefits of adding type hint annotations with static type checking
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Benefits of static type checking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 #. Can find programming errors that would be difficult to find
    otherwise
@@ -40,8 +69,10 @@ Benefits of adding type hint annotations with static type checking
 #. Helps IDEs to make things easier for humans
 #. Helpful for downstream package maintainers
 
-Disadvantages of adding type hint annotations with static type checking
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Disadvantages of static type checking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Requiring type 
 
 1. Requiring type hint annotations that mypy doesn't complain about
    adds an extra step to contributing to Astropy...
@@ -125,9 +156,9 @@ Alternatives
    annotations have begun to be added to Astropy. At present, there is
    no way to check the validity or accuracy of type hint annotations. 
 2. **Take a more aggressive approach to static type checking.** Most
-   recommendations 
+   recommendations
 3. **Delay adding a static type checker to Astropy's continuous
-   integration suite.** 
+   integration suite.**
 
 Decision rationale
 ------------------
--- a/APE25.rst
+++ b/APE25.rst
@@ -1,0 +1,135 @@
+An initial roadmap for static type checking
+-------------------------------------------
+
+author: Nick Murphy
+
+date-created:
+
+date-last-revised:
+
+date-accepted:
+
+type: Standard Track
+
+status: Discussion
+
+
+Abstract
+--------
+
+This APE describes the initial process for gradually adding static
+type checking to Astropy.
+
+Detailed description
+--------------------
+
+
+
+.. This section describes the need for the APE.  It should describe
+   the existing problem that it is trying to solve and why this APE
+   makes the situation better.  It should include examples of how the
+   new functionality would be used and perhaps some use cases.
+
+
+Benefits of adding type hint annotations with static type checking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Can find programming errors that would be difficult to find
+   otherwise
+#. Helps introspection and tab completion in Jupyter notebooks
+#. Helps IDEs to make things easier for humans
+#. Helpful for downstream package maintainers
+
+Disadvantages of adding type hint annotations with static type checking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+1. Requiring type hint annotations that mypy doesn't complain about
+   adds an extra step to contributing to Astropy...
+
+1. An additional step is required to contribute to Astropy.
+
+1. Type hint annotations may add a barrier to entry for contributing
+   to Astropy.
+
+
+
+
+
+Branches and pull requests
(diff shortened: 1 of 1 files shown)


APE25.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Disadvantages of static type checking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Requiring type 
+
+1. Requiring type hint annotations that mypy doesn't complain about
+   adds an extra step to contributing to Astropy...
+
+1. An additional step is required to contribute to Astropy.
```
neutrinoceros, line 80: aren't those the same thing ?

APE25.rst
```diff
@@ -0,0 +1,166 @@
...
+
+1. Type hint annotations may add a barrier to entry for contributing
```
neutrinoceros, line 82: This also *feels* similar to the first two points, but I guess maybe you mean to say that existing type hints may make the code harder to read from a newcomer's perspective ? This is something I've heard in similar contexts, coming from _seasoned_ devs, and I'm honestly not convinced it's a valid point, but it's nonetheless an opinion that exists.

neutrinoceros (commented):
Thank you for starting this ! I feel that two important points may be missing at the moment:
- we ought to acknowledge and emphasis the expected gains (and costs ?) of having type-hints, and of static checking, from a downstream developers/maintainers perspective
- some tools (including mypy, last time I checked) will not attempt to discover type hints in a package unless it contains a `py.typed` marker file. This means that as long as we don't include it anywhere, our type annotations, while not statically checked, are also invisible from the perspective of someone checking downstream code. I think we'll *have* to include it to make our effort visible and useful, and that running a type checker ourselves is a necessary step that needs to happen before it. Meanwhile, as long as we don't include it, we can treat type hints as private details, which I assume will help with early implementation.

APE25.rst
namurphy: Yep!  This is still an early draft so I was trying out different wordings and ways to organize the doc.

namurphy:
It's going to be a while before I can get back to this PR, so if anyone else wants to work on this, please let me know!  

My main goal here has been to encapsulate the advice given in the mypy docs on [enabling static type checking on an existing code base](https://mypy.readthedocs.io/en/stable/existing_code.html).

kbwestfall:

Your chief APE editor is Clara Brasseur (@ceb8).

namurphy:
Thank you for the update, Kyle!  It's still likely that I won't be able to work on this for a while, so I plan to close this pull request unless anyone else wants to take the lead on this. It's also possible that this may be obsolete at this point, since I haven't been following the discussions about this that closely.

namurphy:
An additional note on this topic — for PlasmaPy, I'm hoping to switch us over to using [ty](https://docs.astral.sh/ty/) for static type checking once it becomes a bit more mature, both because of performance and also some capabilities that will likely make it more contributor-friendly than mypy.

kbwestfall:
> Thank you for the update, Kyle! It's still likely that I won't be able to work on this for a while, so I plan to close this pull request unless anyone else wants to take the lead on this. It's also possible that this may be obsolete at this point, since I haven't been following the discussions about this that closely.

No worries.  Thanks, Nick!  I'll leave it to you to decide when/if to close the PR.

neutrinoceros:
With the new editorial process I think merging this APE (even as rejected or undecided) would be useful; this topic has been controversial and the effort needed to move it forward is certainly big enough to warrant fundings. So, anytime this is brought up again in any serious capacity, having this APE/discussion as a precedent will be relevant.

taldcroft:
> With the new editorial process I think merging this APE (even as rejected or undecided) would be useful;

I somewhat disagree, based in particular on the current language in APE 1:
```
Note that there is not much point to submitting new APEs unless someone or some
group has signed up to implement it should the APE be accepted
(typically this would involve the author or authors of the APE). Just issuing
an APE in order to spur others to do work is not going to be received
well, due to the open-source nature of the Astropy Project.
```
IMO,  merging the APE constitutes the action of actively seeking approval with the intent (and resources) to implement the APE.

namurphy:
Thinking more about this, it wouldn't take me that long to get this to a point where it would be worth merging, including adding something like a rationale section for why I'm withdrawing it.

With respect to whether or not to merge this as a withdrawn APE, the main questions I have are:

 - Are these ideas worth keeping as part of the historical record?
 - Is there a reasonable chance that a future APE might want to cite the ideas here?

It's pretty common for a PEP to be written that proposes an idea and gets declined or withdrawn but merged, and another PEP to come along with improvements to the original idea.

There have already been a few PRs submitted that largely take the approach described here.  My favorite is https://github.com/astropy/astropy/pull/16312 by @nstarman, and there were also https://github.com/astropy/astropy/pull/12971 and https://github.com/astropy/astropy/pull/15815. (I also submitted https://github.com/astropy/astropy/pull/15794, but that took an approach I used in PlasmaPy that I have since come to regret. 😅)

taldcroft:
@namurphy - regardless of whether this PR gets merged into the formal list of APE's, the work you've put into it will always remain in this PR. This is citable and won't be forgotten any time soon. Nevertheless I do appreciate the two questions you posed as being entirely reasonable.

I think this is really a question that should go up to the CoCo, namely whether an APE PR should be merged **already withdrawn**.

cc: @eteq @kbwestfall @nstarman @aaryapatil @ceb8

nstarman:
I think this PEP is worth merging, withdrawn or otherwise. Some of these considerations are being addressed outside of the APE mechanism as we decided that some sub-package type checking can happen without an APE, but if (🤞 
when) we need a more package-wide solution this APE will be a valuable resource! 
Personally, I would be very happy to see this APE submitted not withdrawn :).

taldcroft:
@nstarman - Sounds OK, assuming you are speaking for the CoCo with a focus on process rather than the technical merits of the APE. In that case I might suggest making the `type` be `Informational` to convey the intent here.

namurphy:
> Personally, I would be very happy to see this APE submitted not withdrawn :).

Thank you for the encouragement!  I started working on this again, and am reconsidering the decision to withdraw it.  

One change is that I'm switching this to proposing to use ty instead of mypy. I've used mypy for a few years now and don't think it's the best tool at present, and I've been wanting to try out ty for PlasmaPy.  The performance of ty is amazing, and has a helpful feature that `ty check --add-ignore` will add `ty: ignore` comments to all lines that have errors, which can make it much easier and quicker to enable for an existing code base.  I've found the error messages in mypy to be quite confusing at times.  However, ty is in beta, so it may be premature to add. We'll see what happens! 🌱

neutrinoceros:
well it's already in much better shape than early alphas, where checking astropy would hit a runaway memory leak. I'm also much more enhusiast about ty than mypy, especially for a large code base. The performance is one thing (as you noted, it's quite a feat), but what I think takes the cake are much better gradual promises: ty is supposed to be much more resilient than mypy when checking code that contains little to no type info.

taldcroft:
> I started working on this again, and am reconsidering the decision to withdraw it.

@namurphy - without a prototype implementation or identified funding, this will may well earn you the distinction of writing the first APE to get rejected. Just saying... :smile:.

APE25.rst
```diff
@@ -35,37 +32,103 @@ Background
...
+`mypy error codes enabled by default
+<https://mypy.readthedocs.io/en/stable/error_code_list.html#error-codes-enabled-by-default>`__,
+`mypy error codes for optional checks
+<https://mypy.readthedocs.io/en/stable/error_code_list2.html>`__,
+`pyrefly error kinds <https://pyrefly.org/en/docs/error-kinds/>`__, etc.).
+
+A limitation of static type checking is that the analysis is performed
+statically (i.e., without running the code being analyzed). Dynamically
+generated objects (e.g., ``astropy.units.zettajansky``) therefore cannot
+be directly analyzed with a static type checker.
```
nstarman, line 90: Suggesting (as units maintainer, not CoCo) to say that 'directly' is strictly true, but there are ways to ameliorate this, e.g. by generating type stubs. We have mechanisms for this now in `astropy.units`, so it would be good to distinguish between code that:

1. is static so can easily be typed
2. is dynamic but fixed at installation, so can also be typed, though this is more complex
3. is dynamic at runtime, is much more challenging / impossible to statically type.

ceb8:
@namurphy What is this status of this APE at the moment? (I just want to make sure I don't accidentally let it languish when it's ready to move to the next step)

namurphy:
This PR has been stalled for a while since I haven't had the brainwidth or capacity to shepherd this through the process.  But, if anyone else would like to work on this together, please let me know! 🚀

ceb8:
Great! I will let you two work on this then, please ping me when it's ready to move on to the next step.

PR #87, opened 2023-07-16
APE 24: Astropy Affiliated Packages with pyOpenSci

pllim:
Initial authors: @pllim @lwasser @WilliamJamieson @dhomeier

* Close https://github.com/astropy/astropy-project/issues/350
* Close https://github.com/astropy/astropy-project/issues/340
* Close https://github.com/astropy/astropy-project/issues/139
* Fix https://github.com/astropy/astropy-project/issues/334

### TODO

- [x] Start a draft APE during Scipy 2023.
- [x] Finish filling in the text.
- [x] Have co-authors proof-read and address feedback.
    - ✔️ Leah
    - ✔️ Derek
    - ✔️ William
    - ✔️ Moritz
- [x] Open up to general comment period and address feedback. Done on 2023-10-23
- [x] Address feedback from general comment period.
- [x] Wait for CoCo decision. 2023-12-18: 2 abstained, 2 approved, 1 absent
- [ ] CoCo to update PR with decision and merge.
- ???
- Profit!!!

PR #95, opened 2024-01-29
Readme update for APE22 and instruction fix

eteq:
This updates the README both to include APE22 in the index (#87).  It also updates the instructions a bit, as I realized while following them that it doesn't explicitly state we should use the shared account.


pllim (approved):
Thanks!

pllim:
p.s. Caught a typo too late. Fixed it directly on main via 6bdda83 . FYI.

PR #96, opened 2024-02-05
APE 1: Proposal filename should be descriptive (naming chaos fix 1)

pllim:
Motivation: Asking people to put a number in APE proposal filename too early and then changing the number last minute can cause merge conflict (for another open APE proposal) and confusion (for everyone else).

Case study:

* #87 (advertised as APE 24 for many months but merged as APE 22)
* #85 (advertised as APE 22 that can now never be)

But why are we renaming? To prevent gaps in accepted APE listing (e.g., the still open #14 and the missing APE 11 in the list of accepted APEs).

Proposed solution: Update APE 1 to ask APE proposal authors to not use a number at all. Now, the numbering will only be assigned at merge time by a CoCo member doing the merge.

Pros: No more numbering confusing or gaps.
Cons: Renaming a file last minute makes for ugly commit history and weird `git blame`. Also cannot say "APE short-number" in discussions until after it is accepted.

### DIFF
--- a/APE1.rst
+++ b/APE1.rst
@@ -1,11 +1,11 @@
 APE Purpose and Process
 -----------------------
 
-author: Perry Greenfield
+author: Perry Greenfield, Lia Corrales, Thomas Robitaille, Erik Tollerud, Pey Lian Lim
 
 date-created: 2013 October 11
 
-date-last-revised: 2021 February 26
+date-last-revised: 2024 February 5
 
 date-accepted: 2013 November 8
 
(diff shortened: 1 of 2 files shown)


APE1.rst
```diff
@@ -1,19 +1,22 @@
 APE Purpose and Process
 -----------------------
 
-author: Perry Greenfield
+author: Perry Greenfield, Lia Corrales, Thomas Robitaille, Erik Tollerud, Pey Lian Lim
```
bsipocz, line 4: there was a long discussion about not changing the original metadata: 

https://github.com/astropy/astropy-APEs/pull/60#pullrequestreview-502468338

pllim:
No support for this one, so closing.

PR #99, opened 2024-03-11
Ape 1 numbering - Add Zenodo update

ceb8:
Added Zenodo update info and clarification in the readme.


pllim (approved):
Thanks for cleaning up my mess from #97 ! LGTM.

PR #112, opened 2024-11-04
APE26: Removing data storage (representations) from coordinate frames 

jeffjennings:
Coordinate frames in `astropy.coordinates` currently store metadata used to construct the frame (i.e. for transforming between frames) and may also store coordinate data itself. This duplicates functionality with `SkyCoord`, which acts as a container for both coordinate data and reference frame information. We propose to change the frame classes such that they only store metadata and never coordinate data. This would make the implementation more modular, remove ambiguity for users from having nearly duplicate functionality with slightly different APIs, and better satisfy the principle of Separation of Concerns.

Authors of this APE (alphabetical): Jeff Jennings @jeffjennings, Adrian Price-Whelan @adrn, Nathaniel Starkman @nstarman, Marten van Kerkwijk @mhvk

### DIFF
--- a/APE26.rst
+++ b/APE26.rst
@@ -117,10 +117,13 @@ is the more important consideration. In this arena too,
 separating frames from data storage has its advantages. Perhaps
 most importantly, documentation will be more obvious: the methods
 and attributes are defined on ``SkyCoord`` (and ``Coordinate``)
-proper, so `Sphinx <https://www.sphinx-doc.org/>`_ will know how to typeset those, while type
+proper, so `Sphinx <https://www.sphinx-doc.org/>`_ will know how 
+to typeset those, while type
 checkers can help users in finding and using them properly. It
-will also be easier: following the `Zen of Python <https://peps.python.org/pep-0020/>`_, "There should be
-one-- and preferably only one --obvious way to do it." The present overlap leads to
+will also be easier: following the 
+`Zen of Python <https://peps.python.org/pep-0020/>`_, "There should 
+be one-- and preferably only one --obvious way to do it." The 
+present overlap leads to
 confusion wherein beginner users end up creating
 ``BaseCoordinateFrame`` objects such as ``ICRS``, when the docs
 are clear that these are for more advanced users and that
@@ -254,7 +257,9 @@ substeps).
    enabling automatic conversion of the with-data frames into 
    ``SkyCoord`` objects.
 
-4. Deprecating the legacy with-data frame classes, and eventually removing them after a deprecation period that adheres to `APE 2 <https://github.com/astropy/astropy-APEs/blob/main/APE2.rst>`_.
+4. Deprecating the legacy with-data frame classes, and eventually 
+   removing them after a deprecation period that adheres to 
+   `APE 2 <https://github.com/astropy/astropy-APEs/blob/main/APE2.rst>`_.
 
    - Emitting warnings when instantiated.
 
@@ -263,7 +268,7 @@ substeps).
 
    - Remove.
 
-The third step (at stage 3a) is illustrated in the following 
+The fourth step is illustrated in the following 
 pseudocode:
 
 .. code-block:: python
--- a/APE26.rst
+++ b/APE26.rst
@@ -12,30 +12,94 @@ Authors (alphabetical): Jeff Jennings, Adrian Price-Whelan, Nathaniel Starkman, 
 
 Abstract
 --------
-Coordinate frames in astropy.coordinates currently store metadata used to construct the frame (i.e. for transforming between frames) and may also store coordinate *data* itself. This duplicates functionality with `SkyCoord`, which acts as a container for both coordinate data and reference frame information. We propose to change the frame classes such that they only store metadata and never coordinate data. This would make the implementation more modular, remove ambiguity for users from having nearly duplicate functionality with slightly different APIs, and better satisfy the principle of Separation of Concerns.
+Coordinate frames in astropy.coordinates currently store metadata used to construct the 
+frame (i.e. for transforming between frames) and may also store coordinate *data* itself. 
+This duplicates functionality with ``SkyCoord``, which acts as a container for both 
+coordinate data and reference frame information. We propose to change the frame classes 
+such that they only store metadata and never coordinate data. This would make the 
+implementation more modular, remove ambiguity for users from having nearly duplicate 
+functionality with slightly different APIs, and better satisfy the principle of 
+Separation of Concerns.
 
 Detailed description
 --------------------
-The coordinate frame classes (subclasses of `BaseCoordinateFrame`, e.g., `ICRS`, `Galactic`, `FK4`, etc.) are used to represent astronomical reference frames. These may also contain position, velocity, time, or other information about the reference frame relative to another reference frame, which is used to transform coordinates between the reference frames that exist in the `astropy.coordinates` ecosystem. For example, the `AltAz` frame class, which is used when referencing sky coordinates in altitude and azimuth, must contain a location on Earth and a time in order to transform to and from a fixed reference frame like the ICRS. In addition to reference frame information, these frame classes can **also** contain coordinate data (positions, velocities, or other differentials), which are stored internally using the `Representation` and `Differential` classes.
+The coordinate frame classes (subclasses of ``BaseCoordinateFrame``, e.g., ``ICRS``, 
+``Galactic``, ``FK4``, etc.) are used to represent astronomical reference frames. These 
+may also contain position, velocity, time, or other information about the reference frame 
+relative to another reference frame, which is used to transform coordinates between the 
+reference frames that exist in the ``astropy.coordinates`` ecosystem. For example, the 
+``AltAz`` frame class, which is used when referencing sky coordinates in altitude and 
+azimuth, must contain a location on Earth and a time in order to transform to and from a 
+fixed reference frame like the ICRS. In addition to reference frame information, these 
+frame classes can **also** contain coordinate data (positions, velocities, or other 
+differentials), which are stored internally using the ``Representation`` and 
+``Differential`` classes.
 
-The above reveals a problematic oddity with the coordinate frame implementation – an insufficient `separation of concerns <https://en.wikipedia.org/wiki/Separation_of_concerns>`_. The coordinate frame class deals with two separate issues: the definition of the frame and the storage of coordinate data within that frame. Ideally the code would be more modularly structured, where the coordinate frame class **only** defines the reference frame. It would then be the purview of another class – `SkyCoord` – to bring these concerns together and to represent data (using `Representation` and `Differential` classes) in a given reference frame (using a `CoordinateFrame` class). As a demonstration of the current state of duplicated functionality, these two initializations effectively represent the same thing, but return different objects with similar but not identical APIs:
+The above reveals a problematic oddity with the coordinate frame implementation – an 
+insufficient `separation of concerns <https://en.wikipedia.org/wiki/Separation_of_concerns>`_. 
+The coordinate frame class deals with two separate issues: the definition of the frame 
+and the storage of coordinate data within that frame. Ideally the code would be more 
+modularly structured, where the coordinate frame class **only** defines the reference 
+frame. It would then be the purview of another class – `SkyCoord` – to bring these 
+concerns together and to represent data (using `Representation` and `Differential` 
+classes) in a given reference frame (using a `CoordinateFrame` class). As a demonstration 
+of the current state of duplicated functionality, these two initializations effectively 
+represent the same thing, but return different objects with similar but not identical 
+APIs:
 
 .. code-block:: python
 
     c1 = SkyCoord(1., 2., frame="icrs", units="deg")
     c2 = ICRS(1. * u.deg, 2. * u.deg)
 
-We suggest the situation should instead be analogous to what is the case for units, which know how to transform from one to another, but are not concerned with how the values are stored (that belongs to `Quantity`). Translating to coordinates, units are like `CoordinateFrame`, the values are the coordinate data (`Representations`), and `Quantity` is like `SkyCoord`.
+We suggest the situation should instead be analogous to what is the case for units, 
+which know how to transform from one to another, but are not concerned with how the 
+values are stored (that belongs to ``Quantity``). Translating to coordinates, units are 
+like ``CoordinateFrame``, the values are the coordinate data (``Representations``), and 
+``Quantity`` is like ``SkyCoord``.
 
-Having both the frame classes as well as `SkyCoord` be able to store and handle data has resulted in a large amount of code duplication. Restructuring the `Frame` classes to remove data storage will allow for much more maintainable, de-duplicated code. It will also make it easier to contribute: if there is a problem one would like to solve in a given method, if one looks in `SkyCoord`, one will likely find that it does not exist, but instead is defined on `BaseCoordinateFrame` and gets dynamically called via `SkyCoord.__getattr__`. Indeed, the construction of `BaseCoordinateFrame` ends up complicating `SkyCoord`, which has to manage the coordinate data **through** the stored BaseCoordinateFrame.
+Having both the frame classes as well as ``SkyCoord`` be able to store and handle data 
+has resulted in a large amount of code duplication. Restructuring the ``Frame`` classes 
+to remove data storage will allow for much more maintainable, de-duplicated code. It 
+will also make it easier to contribute: if there is a problem one would like to solve 
+in a given method, if one looks in ``SkyCoord``, one will likely find that it does not 
+exist, but instead is defined on ``BaseCoordinateFrame`` and gets dynamically called via 
+``SkyCoord.__getattr__``. Indeed, the construction of ``BaseCoordinateFrame`` ends up 
+complicating ``SkyCoord``, which has to manage the coordinate data **through** the stored 
+``BaseCoordinateFrame``.
 
-Another issue with the current implementation of coordinate frames is that the optional inclusion of coordinate data makes the reference frames “multi-modal”. This creates different usage modes (with and without data), each exhibiting different behavior. For instance, some methods such as `separation` work fine with frames with data, while doing this on coordinate frames without data results in an error. This kind of multi-modal structure is considered an anti-pattern because it forces any code interacting with the frame classes to handle both cases (checking `frame.has_data`), complicating the codebase unnecessarily.
+Another issue with the current implementation of coordinate frames is that the optional 
+inclusion of coordinate data makes the reference frames “multi-modal”. This creates 
+different usage modes (with and without data), each exhibiting different behavior. For 
+instance, some methods such as ``separation`` work fine with frames with data, while 
+doing this on coordinate frames without data results in an error. This kind of 
+multi-modal structure is considered an anti-pattern because it forces any code 
+interacting with the frame classes to handle both cases (checking ``frame.has_data``), 
+complicating the codebase unnecessarily.
 
-All the points discussed thus far – separation of concerns and code duplication – concern maintainers. However user experience is the more important consideration. In this arena too, separating frames from data storage has its advantages. Perhaps most importantly, documentation will be more obvious: the methods and attributes are defined on `SkyCoord`, and sphinx will know how to typeset those. It will also be easier: following the Zen of Python, there should be one clear way to do something. The present overlap leads to confusion where beginner users end up creating `BaseCoordinateFrame` objects such as `ICRS`, when the docs are clear that these are for more advanced users and that `SkyCoord` is to be preferred. The system will also be less fragile. For example, if users manipulate the internal workings of `SkyCoord` (which is discouraged but possible), the coordinate data can become decoupled from the caching that `SkyCoord` performs for speed.
+All the points discussed thus far – separation of concerns and code duplication – 
+concern maintainers. However user experience is the more important consideration. In 
+this arena too, separating frames from data storage has its advantages. Perhaps most 
+importantly, documentation will be more obvious: the methods and attributes are defined 
+on ``SkyCoord``, and sphinx will know how to typeset those. It will also be easier: 
+following the Zen of Python, there should be one clear way to do something. The present 
+overlap leads to confusion where beginner users end up creating ``BaseCoordinateFrame`` 
+objects such as ``ICRS``, when the docs are clear that these are for more advanced users 
+and that ``SkyCoord`` is to be preferred. The system will also be less fragile. For 
+example, if users manipulate the internal workings of ``SkyCoord`` (which is discouraged 
+but possible), the coordinate data can become decoupled from the caching that ``SkyCoord`` 
+performs for speed.
 
 Finished Product
 ----------------
-The end result of the implementation of this APE will be that coordinate frame classes only hold information pertaining to the reference frame they represent and never actual coordinate data in that reference frame. This is consistent with our mathematical framework, the reference frame mediates how coordinate data is understood (e.g. distance measures) or interacts (e.g. separation from other coordinates), but the coordinate data itself is actually independent of that information. Classes like `SkyCoord` will be composed structures bringing together the reference frame (an instance of a `BaseCoordinateFrame` subclass) and the coordinate data (`Representation` objects).
+The end result of the implementation of this APE will be that coordinate frame classes 
+only hold information pertaining to the reference frame they represent and never actual 
+coordinate data in that reference frame. This is consistent with our mathematical 
+framework, the reference frame mediates how coordinate data is understood (e.g. distance 
+measures) or interacts (e.g. separation from other coordinates), but the coordinate data 
+itself is actually independent of that information. Classes like ``SkyCoord`` will be 
+composed structures bringing together the reference frame (an instance of a 
+``BaseCoordinateFrame`` subclass) and the coordinate data (``Representation`` objects).
 
 We illustrate this with the following pseudocode.
 
@@ -58,32 +122,45 @@ We illustrate this with the following pseudocode.
 
 Branches and pull requests
 --------------------------
-No direct progress on these changes has yet occurred. Discussion of these ideas has however arisen in multiple issues and pull requests, demonstrating the need for and utility of the proposed changes.
+No direct progress on these changes has yet occurred. Discussion of these ideas has 
+however arisen in multiple issues and pull requests, demonstrating the need for and 
+utility of the proposed changes.
 
-Several issues have been raised regarding topics such as confusion differentiating the use of `frame` and `SkyCoord` for data storage, and problems arising in other astropy subpackages when using frames that store data. For example:
+Several issues have been raised regarding topics such as confusion differentiating the 
+use of ``frame`` and ``SkyCoord`` for data storage, and problems arising in other astropy 
+subpackages when using frames that store data. For example:
 
 - *Comparing Frame with data and SkyCoord with same data raises exception*: `https://github.com/astropy/astropy/issues/13476`
 - *Add Frame objects without data to a Table*: `https://github.com/astropy/astropy/issues/16823`
 
-Additionally, multiple pull requests have factored out common code between frames and `SkyCoord`, showing that there is no proper separation of concern:
+Additionally, multiple pull requests have factored out common code between frames and 
+``SkyCoord``, showing that there is no proper separation of concern:
 
 - *Introduction of shared `CoordinateFrameInfo`*: `https://github.com/astropy/astropy/pull/16831`
-- *Introduction of `CoordinateSharedMethods`* (this was later removed and instead methods were duplicated): `https://github.com/astropy/astropy/pull/17016`
+- *Introduction of `CoordinateSharedMethods`* (this was later removed and instead methods 
+were duplicated): `https://github.com/astropy/astropy/pull/17016`
 
-Further, pull requests have added methods to make frames and `SkyCoord` even more similar, underscoring that frames *with* data should not be separate entities from `SkyCoord`:
+Further, pull requests have added methods to make frames and ``SkyCoord`` even more 
+similar, underscoring that frames *with* data should not be separate entities from 
+``SkyCoord``:
 
 - *Add .to_table() to frames*: `https://github.com/astropy/astropy/pull/17009`
 - *Add .frame attribute to frames*: `https://github.com/astropy/astropy/pull/16356`
 
 Implementation
 --------------
-The direct use of coordinate frames instead of `SkyCoord` is common. In particular ICRS objects are frequently created with data. Given the prevalent use, it is imperative to maintain backward compatibility and not break the API too quickly. Therefore, we propose implementing this APE through 3 steps (and substeps).
+The direct use of coordinate frames instead of ``SkyCoord`` is common. In particular 
+`ICRS`` objects are frequently created with data. Given the prevalent use, it is imperative 
+to maintain backward compatibility and not break the API too quickly. Therefore, we 
+propose implementing this APE through 3 steps (and substeps).
 
 1. Splitting the frame classes into two hierarchies: ones with and without data.
-2. Switching `SkyCoord` to use the data-less frame classes, and enabling automatic conversion of the with-data frames into `SkyCoord` objects.
+2. Switching ``SkyCoord`` to use the data-less frame classes, and enabling automatic 
+conversion of the with-data frames into ``SkyCoord`` objects.
 3. Deprecating the legacy with-data frame classes.
    1. Emitting warnings when instantiated.
-   2. Still warn, but return a `SkyCoord`, not an instance of its class type (by overriding __new__)
+   2. Still warn, but return a ``SkyCoord``, not an instance of its class type (by 
+   overriding ``__new__``)
    3. Remove.
 
 The 3 steps (at stage 3a) are illustrated in the following pseudocode:
--- a/APE26.rst
+++ b/APE26.rst
@@ -161,12 +161,12 @@ conversion of the with-data frames into ``SkyCoord`` objects.
 
 3. Deprecating the legacy with-data frame classes.
 
-   a. Emitting warnings when instantiated.
+   - Emitting warnings when instantiated.
 
-   b. Still warn, but return a ``SkyCoord``, not an instance of its class type (by 
+   - Still warn, but return a ``SkyCoord``, not an instance of its class type (by 
    overriding ``__new__``)
 
-   c. Remove.
+   - Remove.
 
 The 3 steps (at stage 3a) are illustrated in the following pseudocode:
 
--- a/APE26.rst
+++ b/APE26.rst
@@ -19,7 +19,7 @@ This duplicates functionality with ``SkyCoord``, which acts as a container for b
 coordinate data and reference frame information. We propose to change the frame classes
 such that they only store metadata and never coordinate data, superseding this aspect 
 of the implementation in `APE 5 <https://github.com/astropy/astropy-APEs/blob/main/APE5.rst>`_. 
-This would make the implementation more modular, remove ambiguity for users from having 
+This would make the implementation more modular and performant, remove ambiguity for users from having 
 nearly duplicate functionality with slightly different APIs, and better satisfy the 
 principle of Separation of Concerns.
 
@@ -44,7 +44,7 @@ insufficient `separation of concerns <https://en.wikipedia.org/wiki/Separation_o
 The coordinate frame class deals with two separate issues: the definition of the frame
 and the storage of coordinate data within that frame. Ideally the code would be more
 modularly structured, where the coordinate frame class **only** defines the reference
-frame. It would then be the purview of another class – ``SkyCoord`` – to bring these
+frame. It would then be the purview of another class – like ``SkyCoord`` or the newly-proposed ``Coordinate`` – to bring these
 concerns together and to represent data (using ``Representation`` and ``Differential``
 classes) in a given reference frame (using a ``CoordinateFrame`` class). As a demonstration
 of the current state of duplicated functionality, these two initializations effectively
@@ -81,36 +81,50 @@ doing this on coordinate frames without data results in an error. While this mul
 structure as motivated in `APE 5 <https://github.com/astropy/astropy-APEs/blob/main/APE5.rst>`_ 
 can be seen as a benefit for interpretability of the logic, we also now see it from the
 perspective of an anti-pattern. It forces any code interacting with the frame classes 
-to handle both cases (checking ``frame.has_data``), complicating the codebase. We thus 
+to handle both cases (checking ``frame.has_data``), complicating the codebase. 
+Moreover static analyzers cannot determine which case is being used; with separated frames and data static analyzers will be able to prevent this entire class of errors.
+We thus 
 find it worthwhile to separate coordinate data from reference frames at the possible 
-expense of ease of understanding for some developers.
+expense of developers having to learn this new framework.
 
 All the points discussed thus far – separation of concerns and code duplication –
 concern maintainers. However user experience is the more important consideration. In
 this arena too, separating frames from data storage has its advantages. Perhaps most
 importantly, documentation will be more obvious: the methods and attributes are defined
-on ``SkyCoord`` proper, so sphinx will know how to typeset those, while type checkers
-can help users in finding and using them propertly. It will also be easier:
+on ``SkyCoord`` (and ``Coordinate``) proper, so sphinx will know how to typeset those, while type checkers
+can help users in finding and using them properly. It will also be easier:
 following the Zen of Python, there should be one clear way to do something. The present
 overlap leads to confusion wherein beginner users end up creating ``BaseCoordinateFrame``
 objects such as ``ICRS``, when the docs are clear that these are for more advanced users
 and that ``SkyCoord`` is to be preferred.
 The system will also be less fragile. For example, if users manipulate the 
 internal workings of ``SkyCoord`` (which is discouraged but possible), the coordinate 
-data can become decoupled from the caching that ``SkyCoord`` performs for speed.
+data can become decoupled from the caching that ``SkyCoord`` performs for speed. With these proposed changes users will have a more clear, introspectable, and robust system.
 
 Finished Product
 ----------------
-The end result of the implementation of this APE will be that coordinate frame classes
-only hold information pertaining to the reference frame they represent and never actual
-coordinate data in that reference frame. This is consistent with our mathematical
-framework, the reference frame mediates how coordinate data is understood (e.g. distance
-measures) or interacts (e.g. separation from other coordinates), but the coordinate data
-itself is actually independent of that information. Classes like ``SkyCoord`` will be
-composed structures bringing together the reference frame (an instance of a
-``BaseFrame`` subclass) and the coordinate data (``BaseRepresentation`` objects).
+The end result of the implementation of this APE are be two separate hierarchies
+of classes: reference frame classes and coordinate classes which bring together
+a reference frame and coordinate data. We discuss each class type in turn.
 
-We illustrate this with the following pseudocode.
+Coordinate frame classes only hold information pertaining to the reference frame
+they represent and never actual coordinate data in that reference frame. This is
+consistent with our mathematical framework, the reference frame mediates how
+coordinate data is understood (e.g. distance measures) or interacts (e.g.
+separation from other coordinates), but the coordinate data itself is actually
+independent of that information.
+
+Classes like ``SkyCoord`` will be composed structures bringing together the
+reference frame (an instance of a ``BaseFrame`` subclass) and the coordinate
+data (``BaseRepresentation`` objects). We also introduce a new class,
+``Coordinate``, which is akin to ``SkyCoord`` (containing both frame and data),
+but without extra features like caching and flexible input parsing. In this way
+``Coordinate`` operates very similarly to the current ``BaseCoordinateFrame``
+objects when they have data, and is meant to be their direct replacement in the
+new framework as well as a more lightweight and performant alternative to
+``SkyCoord``. 
+
+We illustrate the new framework with the following pseudocode.
 
 .. code-block:: python
 
@@ -123,11 +137,18 @@ We illustrate this with the following pseudocode.
     class FK5Frame(BaseFrame):
         equinox: TimeAttribute
 
-    class SkyCoord:
+    # ------
+
+    class BaseCoordinate:
         frame: BaseFrame
         data: BaseRepresentation
 
-        def __init__(...):
+    class Coordinate(BaseCoordinate):
+        ...  # it's fast.
+
+    class SkyCoord(BaseCoordinate):
+
+        def __init__(...):  # flexible input parsing
             ...
 
 Branches and pull requests
@@ -156,24 +177,36 @@ similar, underscoring that frames *with* data should not be separate entities fr
 - `Implement BaseCoordinateFrame.to_table() #17009 <https://github.com/astropy/astropy/pull/17009>`_
 - `Implement BaseCoordinateFrame.frame property #16356 <https://github.com/astropy/astropy/pull/16356>`_
 
+In addition, many of these ideas have been developed and tested in parallal in
+the JAX-oriented library `coordinax
+<https://github.com/GalacticDynamics/coordinax>`_. Many of the developers of
+that library are also active Astropy developers and the development effort
+towards ``coordinax`` informs, tests, and validates the ideas presented in this
+APE. In short, it works.
+
+
 Implementation
 --------------
 The direct use of coordinate frames instead of ``SkyCoord`` is common. In particular
 ``ICRS`` objects are frequently created with data. Given the prevalent use, it is imperative
 to maintain backward compatibility and not break the API too quickly. Therefore, we
-propose implementing this APE through 3 steps (and substeps).
+propose implementing this APE through 4 steps (and substeps).
 
 1. Splitting the frame classes into two hierarchies: ones with and without data, with
 the data-less ones getting new names.
 
-2. Switching ``SkyCoord`` to use the data-less frame classes, and enabling automatic
+2. Adding a new ``Coordinate`` class that is similar to ``SkyCoord``, but
+   without extra features like caching and flexible input parsing. It will only
+   accept data-less frame classes.
+
+3. Switching ``SkyCoord`` to use the data-less frame classes, and enabling automatic
 conversion of the with-data frames into ``SkyCoord`` objects.
 
-3. Deprecating the legacy with-data frame classes.
+4. Deprecating the legacy with-data frame classes.
 
    - Emitting warnings when instantiated.
 
-   - Still warn, but return a ``SkyCoord``, not an instance of its class type (by overriding ``__new__``)
+   - Still warn, but return a ``Coordinate``, not an instance of its class type (by overriding ``__new__``)
 
    - Remove.
 
@@ -205,21 +238,24 @@ The third step (at stage 3a) is illustrated in the following pseudocode:
         data: BaseRepresentation
         ...
 
-    class Coordinate(BaseCoordinate):
-        """Data in a reference frame."""
-        pass
-
     class SkyCoord(BaseCoordinate):
          """Data in a reference frame, batteries included."""
 
-        def __init__(...):
+        def __init__(...):  # flexible input parsing
             # If the frame is a LegacyBaseCoordinateFrame then it is
             # split into a BaseFrame and BaseRepresentation.
             ...
 
+        _cache: dict[str, Any]  # cache
+
+    class Coordinate(BaseCoordinate):
+        """Data in a reference frame."""
+        ...  # Direct and fast.
+
     # === Legacy Coordinate Classes ===
 
     class BaseCoordinateFrame(BaseCoordinate):
+        """Reference frames (with optional data storage)."""
 
         def __new__(self):
             warnings.warn("Please use SkyCoord")
--- a/APE26.rst
+++ b/APE26.rst
@@ -1,0 +1,139 @@
+Removing data storage (representations) from coordinate frames
+==============================================================
+
+Authors (alphabetical): Jeff Jennings, Adrian Price-Whelan, Nathaniel Starkman, Marten van Kerkwijk
+---------------------------------------------------------------------------------------------------
+
+:date-created: 2024 11 04
+:date-last-revised: 2024 11 04
+:date-accepted: 202x xx xx
+:type: Standard Track
+:status: Discussion
+
+Abstract
+--------
+Coordinate frames in astropy.coordinates currently store metadata used to construct the frame (i.e. for transforming between frames) and may also store coordinate *data* itself. This duplicates functionality with `SkyCoord`, which acts as a container for both coordinate data and reference frame information. We propose to change the frame classes such that they only store metadata and never coordinate data. This would make the implementation more modular, remove ambiguity for users from having nearly duplicate functionality with slightly different APIs, and better satisfy the principle of Separation of Concerns.
+
+Detailed description
+--------------------
+The coordinate frame classes (subclasses of `BaseCoordinateFrame`, e.g., `ICRS`, `Galactic`, `FK4`, etc.) are used to represent astronomical reference frames. These may also contain position, velocity, time, or other information about the reference frame relative to another reference frame, which is used to transform coordinates between the reference frames that exist in the `astropy.coordinates` ecosystem. For example, the `AltAz` frame class, which is used when referencing sky coordinates in altitude and azimuth, must contain a location on Earth and a time in order to transform to and from a fixed reference frame like the ICRS. In addition to reference frame information, these frame classes can **also** contain coordinate data (positions, velocities, or other differentials), which are stored internally using the `Representation` and `Differential` classes.
+
+The above reveals a problematic oddity with the coordinate frame implementation – an insufficient `separation of concerns <https://en.wikipedia.org/wiki/Separation_of_concerns>`_. The coordinate frame class deals with two separate issues: the definition of the frame and the storage of coordinate data within that frame. Ideally the code would be more modularly structured, where the coordinate frame class **only** defines the reference frame. It would then be the purview of another class – `SkyCoord` – to bring these concerns together and to represent data (using `Representation` and `Differential` classes) in a given reference frame (using a `CoordinateFrame` class). As a demonstration of the current state of duplicated functionality, these two initializations effectively represent the same thing, but return different objects with similar but not identical APIs:
+
+.. code-block:: python
+
+    c1 = SkyCoord(1., 2., frame="icrs", units="deg")
+    c2 = ICRS(1. * u.deg, 2. * u.deg)
+
+We suggest the situation should instead be analogous to what is the case for units, which know how to transform from one to another, but are not concerned with how the values are stored (that belongs to `Quantity`). Translating to coordinates, units are like `CoordinateFrame`, the values are the coordinate data (`Representations`), and `Quantity` is like `SkyCoord`.
+
+Having both the frame classes as well as `SkyCoord` be able to store and handle data has resulted in a large amount of code duplication. Restructuring the `Frame` classes to remove data storage will allow for much more maintainable, de-duplicated code. It will also make it easier to contribute: if there is a problem one would like to solve in a given method, if one looks in `SkyCoord`, one will likely find that it does not exist, but instead is defined on `BaseCoordinateFrame` and gets dynamically called via `SkyCoord.__getattr__`. Indeed, the construction of `BaseCoordinateFrame` ends up complicating `SkyCoord`, which has to manage the coordinate data **through** the stored BaseCoordinateFrame.
+
+Another issue with the current implementation of coordinate frames is that the optional inclusion of coordinate data makes the reference frames “multi-modal”. This creates different usage modes (with and without data), each exhibiting different behavior. For instance, some methods such as `separation` work fine with frames with data, while doing this on coordinate frames without data results in an error. This kind of multi-modal structure is considered an anti-pattern because it forces any code interacting with the frame classes to handle both cases (checking `frame.has_data`), complicating the codebase unnecessarily.
+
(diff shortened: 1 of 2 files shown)


adrn:
Pinging @eerovaher @eteq @StuartLittlefair @Cadair @ayshih @taldcroft as other "stakeholders" who may want to take a look at this and review!

APE26.rst
```diff
@@ -0,0 +1,220 @@
...
+    c2 = ICRS(1. * u.deg, 2. * u.deg)
+
+We suggest the situation should instead be analogous to what is the case for units, 
+which know how to transform from one to another, but are not concerned with how the 
+values are stored (that belongs to ``Quantity``). Translating to coordinates, units are 
+like ``CoordinateFrame``, the values are the coordinate data (``Representations``), and 
+``Quantity`` is like ``SkyCoord``.
+
+Having both the frame classes as well as ``SkyCoord`` be able to store and handle data 
+has resulted in a large amount of code duplication. Restructuring the ``Frame`` classes
```
eerovaher, line 62: Not just code. It's even worse in tests because we have to test both `SkyCoord` and `BaseCoordinateFrame` methods with both `SkyCoord` and `BaseCoordinateFrame` inputs. Not testing all the combinations has led to bugs going unnoticed, e.g. https://github.com/astropy/astropy/pull/15659

APE26.rst
```diff
@@ -0,0 +1,220 @@
...
+doing this on coordinate frames without data results in an error. This kind of 
+multi-modal structure is considered an anti-pattern because it forces any code 
+interacting with the frame classes to handle both cases (checking ``frame.has_data``), 
+complicating the codebase unnecessarily.
+
+All the points discussed thus far – separation of concerns and code duplication – 
+concern maintainers. However user experience is the more important consideration. In 
+this arena too, separating frames from data storage has its advantages. Perhaps most 
+importantly, documentation will be more obvious: the methods and attributes are defined 
+on ``SkyCoord``, and sphinx will know how to typeset those. It will also be easier:
```
eerovaher, line 84: This is relevant not only for documentation but also for type annotations. Looking up `BaseCoordinateFrame` methods via `SkyCoord.__getattr__()` is too dynamic for static type checkers. It is possible to partially alleviate this by writing out stubs in `SkyCoord` (e.g. https://github.com/astropy/astropy/blob/24d48a9e2ef90af5383bf76e6d098e6d193d8186/astropy/coordinates/sky_coordinate.py#L178-L181), but type checkers cannot ensure that the stubs agree with the signatures of the actual methods.

APE26.rst
```diff
@@ -0,0 +1,220 @@
...
+- *Introduction of shared `CoordinateFrameInfo`*: `https://github.com/astropy/astropy/pull/16831`
+- *Introduction of `CoordinateSharedMethods`* (this was later removed and instead methods 
+were duplicated): `https://github.com/astropy/astropy/pull/17016`
+
+Further, pull requests have added methods to make frames and ``SkyCoord`` even more 
+similar, underscoring that frames *with* data should not be separate entities from 
+``SkyCoord``:
+
+- *Add .to_table() to frames*: `https://github.com/astropy/astropy/pull/17009`
+- *Add .frame attribute to frames*: `https://github.com/astropy/astropy/pull/16356`
```
eerovaher, line 148: I find it confusing that you have not used the pull request titles.

eerovaher (commented):
I am convinced that there should be exactly one type for coordinate data. This would prevent users from getting confused about which type they should be using and simplifies things for both `astropy` and downstream developers because we'll only have to support one type of coordinate data as input for our functions.

This leaves two possibilities: the single type is a class, which is what this APE proposes, or the single type is a [Protocol](https://peps.python.org/pep-0544/), which is what `Implement BaseCoordinateFrame.frame property` (astropy/astropy#16356) (that the APE mentions) proposed. Implementing a `Protocol` would be backwards-compatible and wouldn't cause disruption for downstream packages. In every other regard this APE is a better idea.

APE26.rst
jeffjennings: Good point, thank you. Fixed in [46f503e](https://github.com/astropy/astropy-APEs/pull/112/commits/46f503ee97e7ec3f1d050b29b6949a52a48272c2)

StuartLittlefair:
Yes, yes and a thousand times yes. 

The devil will be in the implementation details but this is a thoroughly good idea.

adrn:
Does anyone else have thoughts before we ask the @astropy/coordinators to review this? cc @eteq @Cadair @ayshih @taldcroft!

eteq:
I have been traveling and haven't had a chance to review this in detail, but I will try to do so this week.  I am fairly sure I will have Opinions :tm:

adrn:
@eteq OK cool. It'd be good to get your feedback ASAP, because @jeffjennings is starting to work through some of the changes we discuss here (in draft stages, but still, it is somewhat in progress).

Cadair:
I have two main thoughts about some effects from this change, I think they can both be managed, but probably worth having a think about.

1. The asdf serialisation is going to substantially different from now, managing that transition could be complex. - @braingram 
2. Many SunPy frames have an `observer` attribute which is a realised frame holding the position of the observer. This could obviously become a SkyCoord, but I worry that adds some more complexity, and will for a start make the repr even more obnoxious than it is currently.

braingram:
>     1. The asdf serialisation is going to substantially different from now, managing that transition could be complex. - @braingram

Thanks for pinging me about this change. If I'm following the end result will be that coordinate frames will no longer have `data`.

#### Reading old files with `data` when coordinate frames no longer support `data`
At that point when a user opens an ASDF file with a serialized coordinate frame (which might contain [data](https://github.com/asdf-format/asdf-coordinates-schemas/blob/b04d5ee1f2971545fab2a1f7a75314f9a8d48682/resources/schemas/frames/baseframe-1.1.0.yaml#L15)) the [converter](https://github.com/astropy/asdf-astropy/blob/8147b824d56ab24a5c3f7572e43215c14fe7c134/asdf_astropy/converters/coordinates/frame.py#L50) that handles the deserialization could see that the file contains `data` and generate a `SkyCoord` instead (possibly with a warning that the file should be updated).

#### Writing files when coordinate frames no longer support `data`
This should be fine. Technically we could keep the same schema and just never write `data` (it's not required) but it might be a good opportunity to include some other schema cleanup (at the expense of some schema churn since many schemas reference the base coordinate frame schema).

#### Implementation details
The details about how this is implemented may impact ASDF serialization. Specifically the first item under implementation:

> 1. Splitting the frame classes into two hierarchies: ones with and without data.

The converter(s) would need to be updated to support these new classes. This isn't a problem but is work that wouldn't be needed if instead the existing classes were updated (deprecating and then removing the `data` feature).

> 3. Deprecating the legacy with-data frame classes.

At this point asdf-astropy could be updated to not use the legacy classes.

eteq (commented):
Alright, I've left several specific inline comments.  I also have several bigger picture concerns.  Note that I am *not* dead-set against this - I get the motivation for this change and can see how it might be an improvement.  But I want to be sure we address these concerns before going down this road:

* This is a major change in communicating the purpose of SkyCoord.  Previously (i.e. in APE5) SkyCoord is highlighted as a "convenience class" - primarily this is 1)  it has a much smarter and easier to use initializer, and 2) it carries the frame attributes along with the data, regardless of whether that frame attribute is in the current frame.  This makes subtly different behavior for  certain transformations depending on whether you give the to frame the same frame attributes or not.  I don't really understand how this proposal will change that - i.e., if there are no frame classes with data, how do you say "do the transform that only uses the frame attributes in the to-frame, but doesn't carry them over from the from-frame"?  I can thinks of ways to do this, but this was one of the prime reasons for data and frames being tied together in the first place, to keep that case simpler and not dependent on SkyCoord.
* We should get someone in Sunpy look at this before proceeding.  *hopefully* this won't be a big change for them since I think mostly they define their own frames and transforms, which should be backwards compatible with these changes.  But it may be they use the low-level classes directly more often and hence this would be a huge negative impact on them.
* Relatedly, I'm concerned about the number of examples out in the wild that use the low-level classes directly. There are various use cases where we have *encouraged* people to use the low-level classes with data because they are much cheaper for initialization. If we are going to remove that option they will be impacted whenever the removal finally happens.
* I think it's absolutely critical we highlight the change in philosophy from what APE5 said. This sub-package is sufficiently complex that documentation (either in APEs or otherwise) are critical to helping people understand the big-picture.  This might be as simple as just making sure https://docs.astropy.org/en/stable/coordinates/index.html#astropy-coordinates-overview and related documentation spots are up-to-date, but still, that should be explicitly in the implementation plan IMHO since it's a non-trivial effort to update all the docs for this.

APE26.rst
```diff
@@ -0,0 +1,223 @@
...
+    class BaseCoordinateFrame(BaseCoordinate):
+
+        def __new__(self):
+            warnings.warn("Please use SkyCoord")
+
+        @abstractpropery # implemented on subclasses
+        def frame(self) -> BaseFrame:
+            ...
+
+    class ICRS(LegacyCoordinateFrame, ICRSFrame):
```
adrn, line 219:
```suggestion
    class ICRS(BaseCoordinateFrame, ICRSFrame):
```

APE26.rst
```diff
@@ -0,0 +1,223 @@
...
+        ...
+
+    class FK5(LegacyCoordinateFrame, FK5Frame):
```
adrn, line 222:
```suggestion
    class FK5(BaseCoordinateFrame, FK5Frame):
```

APE26.rst
```diff
@@ -0,0 +1,223 @@
...
+
+    # === Coordinates (data + frame) ===
+
+    class BaseCoordinate:
+        """Base class for data in a reference frame."""
+        frame: BaseFrame
+        data: BaseRepresentation
+        ...
+
+    class SkyCoord(BaseCoordinate):
```
nstarman, line 201:
```suggestion

class Coordinate(BaseCoordinate):
    """Data in a reference frame."""
    pass

    class SkyCoord(BaseCoordinate):
         """Data in a reference frame, batteries included."""
```

jeffjennings:
Hi @ayshih and @Cadair, we're starting to finalize this APE and are wondering if you anticipate any major concerns with respect to SunPy. The first phase of the APE wouldn't introduce any breaking changes, the second phase (timeline flexible) would add deprecation warnings, and the final phase (several major astropy versions away) would require changes to SunPy. Thanks!

APE26.rst
jeffjennings: Great point, Marten has amended the text to note this in [9e33a95](https://github.com/astropy/astropy-APEs/pull/112/commits/9e33a95f4e74054a022dbbfc09788343bef219b8) - so I'm going to mark this as resolved.

APE26.rst
jeffjennings: Thanks Eero. Marten has amended the text to add this in [9e33a95](https://github.com/astropy/astropy-APEs/pull/112/commits/9e33a95f4e74054a022dbbfc09788343bef219b8) - so I'm going to mark this as resolved.

APE26.rst
```diff
@@ -0,0 +1,219 @@
...
+complicating the codebase unnecessarily.
+
+All the points discussed thus far – separation of concerns and code duplication – 
+concern maintainers. However user experience is the more important consideration. In 
+this arena too, separating frames from data storage has its advantages. Perhaps most 
+importantly, documentation will be more obvious: the methods and attributes are defined 
+on ``SkyCoord``, and sphinx will know how to typeset those. It will also be easier: 
+following the Zen of Python, there should be one clear way to do something. The present 
+overlap leads to confusion where beginner users end up creating ``BaseCoordinateFrame`` 
+objects such as ``ICRS``, when the docs are clear that these are for more advanced users
```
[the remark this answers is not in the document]
jeffjennings: I can see how this is a feature, but I think the improved user clarity is motivated earlier on in the text - currently, a user can initialize an `ICRS` object or a `SkyCoord` object, which "represent the same thing, but return different objects with similar but not identical APIs". The proposed change gives only way to have an object hold both data and frame information - `SkyCoord` - preventing user confusion about subtle differences between `SkyCoord` and a with-data frame like the current `ICRS` (including which one to use, and why their API differs in perhaps unexpected ways).

APE26.rst
```diff
@@ -0,0 +1,219 @@
...
+``SkyCoord.__getattr__``. Indeed, the construction of ``BaseCoordinateFrame`` ends up 
+complicating ``SkyCoord``, which has to manage the coordinate data **through** the stored 
+``BaseCoordinateFrame``.
+
+Another issue with the current implementation of coordinate frames is that the optional 
+inclusion of coordinate data makes the reference frames “multi-modal”. This creates 
+different usage modes (with and without data), each exhibiting different behavior. For 
+instance, some methods such as ``separation`` work fine with frames with data, while 
+doing this on coordinate frames without data results in an error. This kind of 
+multi-modal structure is considered an anti-pattern because it forces any code
```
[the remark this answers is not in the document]
jeffjennings: Thanks Erik - I've reworked the text to reflect this in [46a6d82](https://github.com/astropy/astropy-APEs/pull/112/commits/46a6d822a53863f38951b848e21e9a94cb111213) (including explicitly noting the design choice we propose that departs from APE 5, as you suggested).

APE26.rst
```diff
@@ -0,0 +1,219 @@
...
+reference frames that exist in the ``astropy.coordinates`` ecosystem. For example, the 
+``AltAz`` frame class, which is used when referencing sky coordinates in altitude and 
+azimuth, must contain a location on Earth and a time in order to transform to and from a 
+fixed reference frame like the ICRS. In addition to reference frame information, these 
+frame classes can **also** contain coordinate data (positions, velocities, or other 
+differentials), which are stored internally using the ``Representation`` and 
+``Differential`` classes.
+
+The above reveals a problematic oddity with the coordinate frame implementation – an 
+insufficient `separation of concerns <https://en.wikipedia.org/wiki/Separation_of_concerns>`_.
```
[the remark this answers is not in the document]
jeffjennings: That's a good point - I've reworded the text to reflect this in [46a6d82](https://github.com/astropy/astropy-APEs/pull/112/commits/46a6d822a53863f38951b848e21e9a94cb111213).

APE26.rst
```diff
@@ -0,0 +1,219 @@
...
+Authors (alphabetical): Jeff Jennings, Adrian Price-Whelan, Nathaniel Starkman, Marten van Kerkwijk
+---------------------------------------------------------------------------------------------------
+
+:date-created: 2024 11 04
+:date-last-revised: 2024 11 04
+:date-accepted: 202x xx xx
+:type: Standard Track
+:status: Discussion
+
+Abstract
```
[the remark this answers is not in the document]
jeffjennings: You're right - I've added text to give the context of APE 5, and to motivate what we propose changing and why, in [46a6d82](https://github.com/astropy/astropy-APEs/pull/112/commits/46a6d822a53863f38951b848e21e9a94cb111213).

APE26.rst
```diff
@@ -0,0 +1,219 @@
...
+--------------
+The direct use of coordinate frames instead of ``SkyCoord`` is common. In particular 
+``ICRS`` objects are frequently created with data. Given the prevalent use, it is imperative 
+to maintain backward compatibility and not break the API too quickly. Therefore, we 
+propose implementing this APE through 3 steps (and substeps).
+
+1. Splitting the frame classes into two hierarchies: ones with and without data.
+
+2. Switching ``SkyCoord`` to use the data-less frame classes, and enabling automatic 
+conversion of the with-data frames into ``SkyCoord`` objects.
```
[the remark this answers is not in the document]
jeffjennings: There wouldn't be duplication because the coordination conversion machinery would be in the new (data-less) frame classes, and the legacy (with-data) frame classes would subclass the new (data-less) classes.

pllim:
Thanks! Any thoughts on performance impact?

nstarman:
There should be no performance impact on `SkyCoord`. Small performance improvement on byte representations e.g. file schemas related to the new frame classes since they don't carry data. Memory improvements for the new `Coordinate` class since it doesn't do cacheing.

APE26.rst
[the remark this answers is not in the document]
eteq: Ah, I see that... But I'm a bit concerned that this still doesn't address the fact that there was a *reason* for this (i.e., they are intentionally different APIs because the objects behave differently when transformed)

eteq:
Sorry if I missed it, but I'm still confused about what I see as the two biggest user-facing change here (i.e. the first bullet point in my review):

1. that when you transform a SkyCoord with frame attributes to a different frame, it behaves *differently* than transforming  a low-level class, and both are useful in some  contexts. (can expand more on this if needed)
2. that the initializing a low-level class with data is faster than a SkyCoord, by design, because there's less parsing.

Those were the main reasons for having low-level frame objects separate from SkyCoord, and both required low-level classes to have data. In this proposal, how does one do that?

eerovaher:
> ...when you transform a SkyCoord with frame attributes to a different frame, it behaves _differently_ than transforming  a low-level class, and both are useful in some  contexts. (can expand more on this if needed)

If `SkyCoord` had a method that would remove all the stored frame attributes expect the ones needed for its current frame then that should allow fully replicating the behavior of the current frame instances.

nstarman:
@eteq It looks like we should add more verbiage related to `Coordinate`, but if you scroll down to 208-210, there's a `Coordinate` class which addresses both your points. It's the concrete class of `BaseCoordinate`, like `SkyCoord`, but without any of the bells and whistles like cacheing or parsing. It's the new form of the low-level class, but with the separation between reference frame and coordinate data.

ayshih:
Sorry for taking so long to chime in!  I'm no stranger to the subtleties of frames with data versus SkyCoord (e.g., astropy/astropy#10475), so as a developer I'm definitely sympathetic to the motivation behind this APE.  However, I'm skeptical that there are sufficient *user-facing* benefits for this change to be worth the new development effort.  (For example, I'd be enthusiastic if this change were in tandem with one of the major enhancements I suggested at last year's coordination meeting.)

To be clear, writing the new code in `astropy` core would probably not itself be a huge development effort, although I certainly would not look forward to verification and documentation.  With my SunPy hat on, I don't anticipate any *technical* showstoppers for continuing to extend the `astropy` coordinates framework, but I foresee significant headaches to supporting multiple `astropy` releases across the transitions/deprecations.  The headaches could then be exponentiated for those packages that extend off of `sunpy`.

I also share the concern raised by @eteq just above.  There's the possibility having to support the various use cases with this new class organization might actually end up simply pushing complexity elsewhere in the code (e.g., having to frequently call the attribute-stripping method suggested by @eerovaher).

Just checking: Is the plan to continue to have `SkyCoord` be able to store frame attributes that are not used for the current frame?  The APE reads as if `SkyCoord` stores only frame and data.

mhvk:
> Just checking: Is the plan to continue to have `SkyCoord` be able to store frame attributes that are not used for the current frame? The APE reads as if `SkyCoord` stores only frame and data.

The plan means `SkyCoord` will continue to look effectively the same from a user perspective, so, yes, it would still carry attributes not associated with the frame.  As @nstarman noted, however, the new `Coordinate` class, however, would not - it would just contain the frame with its attributes and the representation.

Aside: I was at the last meeting, but fear I only vaguely recall the enhancements you mentioned suggesting. For my benefit, as well as of those who didn't attend, is there a few-line summary? It might be relevant mostly if the scheme proposed here helps or hinders those suggestions...

nstarman:
> (For example, I'd be enthusiastic if this change were in tandem with one of the major enhancements I suggested at last year's coordination meeting.)

I'm also interested to hear what those were. I was not at the meeting.

> However, I'm skeptical that there are sufficient user-facing benefits for this change to be worth the new development effort.

I think this APE opens up an arena for a wide-ranging set of user-facing benefits. Ones possible when we simultaneously simplify, make more robust, and speed up the code! For example:

1. Frames are much simpler objects. It'll be much easier to define your own. Even dynamically. See [this file](https://github.com/GalacticDynamics/coordinax/blob/main/src/coordinax/_src/frames/xfm.py) for how an analogous implementation of this APE in a jax-compatible library makes it trivial to extend the *frame*work.
2. Frames can have defined memory footprints since they no longer store data. This is faster, more memory efficient, and more compatible with schematization.
3. Since frames don't store data it's easier to attach them to custom objects and have that object duck-type `BaseCoordinate`. Now *users* can define their own coordinate-like objects that do exactly what they need. I don't think it's in scope for this APE, but it would be great in followups to ensure that a `BaseCoordinate` duck type can be used in most places in Astropy. That's a huge opening up of the field for users. A major user-facing benefit. Only possible with this APE.
4. [Think of more things here if requested]


> With my SunPy hat on, I don't anticipate any technical showstoppers for continuing to extend the astropy coordinates framework, but I foresee significant headaches to supporting multiple astropy releases across the transitions/deprecations

👍. But given the outlined plan I'm having trouble anticipating the same set of headaches.
Step 1 should see 0 API changes for people using `ICRS`, etc., or `SkyCoord`. The only new thing is `ICRSFrame`. But if a function needs data then it's trivial to put a check

```
def function(coord: Any):
    if not isinstance(coord, BaseCoordinate):  # BaseCoordinate  covers all of SkyCoord, Coordinate, ICRS, etc
        raise ValueError("oops. this doesn't have data. did you pass one of the new BaseFrame objects?")
```

Step 2 has no effect on user API.

Step 3a adds a deprecation warning, which users will see whenever they define an old-style ICRS object. Not sure how that would affect ``function``.
Step 3b swaps for SkyCoord, so there's no deprecation warning and ``function`` should continue to work without change
step 3c no more old-style ICRS object. No change to ``function``.


> I also share the concern raised by @eteq just above. There's the possibility having to support the various use cases with this new class organization might actually end up simply pushing complexity elsewhere in the code (e.g., having to frequently call the attribute-stripping method suggested by @eerovaher).

See https://github.com/astropy/astropy-APEs/pull/112#issuecomment-2756186671. `SkyCoord` isn't changing. no-cache and simple-parse coordinates will continue to exist with `Coordinate`.


It sounds like there should be some verbiage added to explain what is shown in the Implementation code examples!
Suggestions appreciated!

ayshih:
> Aside: I was at the last meeting, but fear I only vaguely recall the enhancements you mentioned suggesting. For my benefit, as well as of those who didn't attend, is there a few-line summary? It might be relevant mostly if the scheme proposed here helps or hinders those suggestions...

They certainly aren't fleshed-out ideas yet, they would require major changes:

- Separately define true/actual coordinates from apparent coordinates -> add support for object propagation in general
- Decouple origin location and axes orientation a la SPICE -> add support for transforming vector fields

The new scheme proposed here shouldn't hinder any of the above, and it's quite possible that it'd be helpful, if for no other reason than that the internals would be less duplicated/tangled.  My point is that it sure would be nice if we were aiming for something grand at the same time as making these changes.  For example, if people were on board with the second bullet, instead of merely having `Coordinate` be data plus a data-less frame, we would go ahead and make it data plus frame origin plus frame orientation to avoid a future deprecation cycle.

ayshih:
> 1. Frames are much simpler objects. It'll be much easier to define your own. Even dynamically.

Hmm, I'm not sure I follow.  Subclassing `BaseCoordinateFrame` is already simple.  The heavy lifting is in defining the transformations between frames, and you'll still have to do that regardless of the scheme.

> 2. Frames can have defined memory footprints since they no longer store data. This is faster, more memory efficient, and more compatible with schematization.

True, but I don't feel like those benefits would ever be appreciable to users.  I think the slog in virtually all use cases is tied to working with coordinates, not with data-less frames.

> 3. Since frames don't store data it's easier to attach them to custom objects and have that object duck-type `BaseCoordinate`.

Yup, I agree, but as I said in my previous post, I wish we were building toward something now instead of a hypothetical future.

> 👍. But given the outlined plan I'm having trouble anticipating the same set of headaches.

You may be right that it'll not be as painful as I fear.  I'll mull it over more.

ayshih:
> With my SunPy hat on, I don't anticipate any _technical_ showstoppers for continuing to extend the `astropy` coordinates framework

I just realized that there is an approach that we use in SunPy that I'm not immediately sure how we'd adapt to this new scheme.  Some of our frames define frame-specific methods/properties that take advantage of both (1) having access to the coordinate data and (2) being exposed as methods/properties callable on the `SkyCoord`.  As I understand the new scheme, that will no longer be possible?

mhvk:
@ayshih - :+1: to your general point of trying to think a bit bigger at the same time!

> * Decouple origin location and axes orientation a la SPICE -> add support for transforming vector fields

That would be interesting. It might also make something like `SkyOffsetFrame` a little easier. Would there ever be a reason not to have both, though? I.e., if we had origin and orientation separately, would it not be fine for `Frame` to just be a container of those two? (Sorry, too little thought here!)

nstarman:
> Yup, I agree, but as I said in my previous post, I wish we were building toward something now instead of a hypothetical future.

Not hypothetical:

- https://github.com/GalacticDynamics/coordinax
- https://github.com/GalOrrery/interpolated-coordinates

The only reason there aren't more ducktypes is that we don't support them. People don't use what they can't make. Build it and they shall come.

StuartLittlefair:
Just to add a comment to the discussion between @ayshih and @nstarman. I think we should not underestimate the benefits to improving the simplicity of the coordinates code base. Personally I have always tried to avoid work which involves the internals of SkyCoord because the existing code base is so complex. If the proposed APE improves this, and I believe it will, then I believe it offers us *significant* advantages in terms of maintainability, the ability to add new features, and the ability to encourage others to contribute. 

As an unrelated comment, there is an upcoming funding opportunity in the UK to provide support to existing scientific software packages, so I may be able to provide meaningful support to implementation of this APE, if approved.

APE26.rst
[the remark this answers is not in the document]
pllim: Should APE 5 also be updated slightly to link to this one in this PR, so that in the future, someone who stumbles upon APE 5 knows to also go here for the rest of the story?

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+differentials), which are stored internally using the 
+``Representation`` and ``Differential`` classes.
+
+The above reflects a design choice in the coordinate frame 
+implementation in `APE 5 
+<https://github.com/astropy/astropy-APEs/blob/main/APE5.rst>`_ 
+that, following several years of use, we have learned from and 
+think best to revise. We now view it as an insufficient 
+`separation of concerns 
+<https://en.wikipedia.org/wiki/Separation_of_concerns>`_. The
```
pllim, line 53: This link should be moved to where it is first mentioned in the Abstract?

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+The end result of the implementation of this APE are be two separate
+hierarchies of classes: reference frame classes and coordinate
+classes which bring together a reference frame and coordinate
+data. We discuss each class type in turn.
+
+Coordinate frame classes only hold information pertaining to
+the reference frame they represent and never actual coordinate
+data in that reference frame. This is consistent with our
+mathematical framework, the reference frame mediates how
+coordinate data is understood (e.g. distance measures) or
```
pllim, line 145:
```suggestion
coordinate data is understood (e.g., distance measures) or
```

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+interacts (e.g. separation from other coordinates), but the
```
pllim, line 146:
```suggestion
interacts (e.g., separation from other coordinates), but the
```

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+the possible expense of developers having to learn this new
+framework.
+
+All the points discussed thus far – separation of concerns and
+code duplication – concern maintainers. However user experience
+is the more important consideration. In this arena too,
+separating frames from data storage has its advantages. Perhaps
+most importantly, documentation will be more obvious: the methods
+and attributes are defined on ``SkyCoord`` (and ``Coordinate``)
+proper, so sphinx will know how to typeset those, while type
```
pllim, line 120: Not very sure what this means but also maybe unnecessary to call out Sphinx specifically, as our code design should not be dictated by our choice of documentation renderer. But if you choose to keep the text, should capitalize and link, as follows.

```suggestion
proper, so `Sphinx <https://www.sphinx-doc.org/>`_ will know how to typeset those, while type
```

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+checkers can help users in finding and using them properly. It
+will also be easier: following the Zen of Python, there should be
```
pllim, line 122:
```suggestion
will also be easier: following the `Zen of Python <https://peps.python.org/pep-0020/>`_, "There should be
```

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+one clear way to do something. The present overlap leads to
```
pllim, line 123:
```suggestion
one-- and preferably only one --obvious way to do it." The present overlap leads to
```

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+4. Deprecating the legacy with-data frame classes.
+
+   - Emitting warnings when instantiated.
+
+   - Still warn, but return a ``Coordinate``, not an instance of 
+     its class type (by overriding ``__new__``)
+
+   - Remove.
+
+The third step (at stage 3a) is illustrated in the following
```
pllim, line 266: What exactly is `stage 3a` again? I don't see any `a.` sub-bullet until Step 3 above.

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+   ``SkyCoord``, but which does not keep any frame attributes not 
+   in the current frame, and does not have extra features like 
+   caching and flexible input parsing. It will only accept 
+   data-less frame classes.
+
+3. Switching ``SkyCoord`` to use the data-less frame classes, and 
+   enabling automatic conversion of the with-data frames into 
+   ``SkyCoord`` objects.
+
+4. Deprecating the legacy with-data frame classes.
```
pllim, line 257:
```suggestion
4. Deprecating the legacy with-data frame classes, and eventually removing them after a deprecation period that adheres to `APE 2 <https://github.com/astropy/astropy-APEs/blob/main/APE2.rst>`_.
```

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+Implementation
+--------------
+The direct use of coordinate frames instead of ``SkyCoord`` is 
+common. In particular ``ICRS`` objects are frequently created 
+with data. Given the prevalent use, it is imperative to maintain 
+backward compatibility and not break the API too quickly. 
+Therefore, we propose implementing this APE through 4 steps (and 
+substeps).
+
+1. Splitting the frame classes into two hierarchies: ones with
```
pllim, line 244: For these steps, I would like to know more about what is your proposed timeline and exactly how break-y things will get at each step. And what would we do about helping to notify or patch affected downstream packages. So maybe even list each step as a sub-section if needed.

Even nicer is actual example usage code (not the implementation but *usage* of the implementation) and what the expected output would look like, for an example use case that is very common. Doesn't have to be fancy, just have to reflect what most of us use, which is usually something common like `SkyCoord(ra_arr, dec_arr, unit="deg")`.

pllim (commented):
Thanks! My main concern is the downstream effect, which drove some of my comments below. Overall, thank you for the well thought out proposal and links.

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+case for units, which know how to transform from one to another, but 
+are not concerned with how the values are stored (that belongs to 
+``Quantity``). Translating to coordinates, units are like 
+``CoordinateFrame``, the values are the coordinate data 
+(``Representations``), and ``Quantity`` is like ``SkyCoord``.
+
+Having both the frame classes as well as ``SkyCoord`` be able to
+store and handle data has resulted in a large amount of code
+duplication (not just in the code proper, but also in the tests,
+leading to missed combinations). Restructuring the ``Frame``
```
eerovaher, line 82: The statement in the parenthesis is the opposite of what it should be. Having duplicated test code does not lead to missing coverage. Not having duplicated tests is what has caused problems to go unnoticed.

It is also worth remarking that having duplicated code sometimes means quadrupled (not duplicated) tests because we have to test both `BaseCoordinateFrame` and `SkyCoord` methods with both `BaseCoordinateFrame` and `SkyCoord` arguments.

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+classes to remove data storage will allow for much more
+maintainable, de-duplicated code. It will also make it easier to
+contribute: if there is a problem one would like to solve in a
+given method, if one looks in ``SkyCoord``, one will likely find
+that it does not exist, but instead is defined on
+``BaseCoordinateFrame`` and gets dynamically called via
+``SkyCoord.__getattr__``. Indeed, the construction of
```
eerovaher, line 89: I have some doubts whether or not potential new contributors manage to find `SkyCoord.__getattr__()`.

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+is the more important consideration. In this arena too,
+separating frames from data storage has its advantages. Perhaps
+most importantly, documentation will be more obvious: the methods
+and attributes are defined on ``SkyCoord`` (and ``Coordinate``)
+proper, so sphinx will know how to typeset those, while type
+checkers can help users in finding and using them properly. It
+will also be easier: following the Zen of Python, there should be
+one clear way to do something. The present overlap leads to
+confusion wherein beginner users end up creating
+``BaseCoordinateFrame`` objects such as ``ICRS``, when the docs
```
eerovaher, line 125:
```suggestion
``BaseCoordinateFrame`` instances, when the docs
```

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+are clear that these are for more advanced users and that
+``SkyCoord`` is to be preferred. The system will also be less
+fragile. For example, if users manipulate the internal workings
+of ``SkyCoord`` (which is discouraged but possible), the
+coordinate data can become decoupled from the caching that
+``SkyCoord`` performs for speed. With these proposed changes
```
eerovaher, line 131: I am not worried about users who mess with `SkyCoord` internals. The problems they will cause will be their own fault. But it might be worth considering downstream developers who subclass `SkyCoord`.

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+together the reference frame (an instance of a ``BaseFrame``
+subclass) and the coordinate data (``BaseRepresentation``
+objects). We also introduce a new class, ``Coordinate``, which
+is akin to ``SkyCoord`` (containing both frame and data), but
+without extra features like keeping frame attributes not
+associated with the current frame, caching and flexible input
+parsing. In this way ``Coordinate`` operates very similarly to
+the current ``BaseCoordinateFrame`` objects when they have
+data, and is meant to be their direct replacement in the new
+framework as well as a more lightweight and performant
```
eerovaher, line 160: How does the new `Coordinate` class avoid code duplication and user confusion that this APE tries to solve? Why can't we just use `SkyCoord`?

APE26.rst
nstarman: Yeah, the first sub-bullet in Step 3. Except now it's called Step 4. Needs to be updated.

APE26.rst
nstarman: > avoid code duplication and user confusion that this APE tries to solve?

Common base class.

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+<https://github.com/astropy/astropy-APEs/blob/main/APE5.rst>`_, 
+coordinate frames in astropy.coordinates currently store metadata 
+used to construct the frame (i.e. for transforming between frames) 
+and may also store coordinate *data* itself. This duplicates 
+functionality with ``SkyCoord``, which acts as a container for both 
+coordinate data and reference frame information. We propose to 
+change the frame classes such that they only store metadata and 
+never coordinate data, superseding this aspect of the implementation 
+in `APE 5 <https://github.com/astropy/astropy-APEs/blob/main/APE5.rst>`_. 
+This would make the implementation more modular and performant,
```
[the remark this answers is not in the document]
nstarman: > The other reasons you are listing here should be convincing enough.

Agreed, but also...

>>> you can back this claim with measurements

But we can know this even without. Number of times an object is constructed, going through non-fast-path parsing/validation routes, slotting (easy with dataclases), etc.

APE26.rst
nstarman: > Not having duplicated tests is what has caused problems to go unnoticed.

👍 😆.

APE26.rst
nstarman: That is a better example to highlight.

APE26.rst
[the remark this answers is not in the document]
eerovaher: Either back up your claims with actual measurements or don't bring up performance at all. I wouldn't be surprised if the new API allows for faster implementation, but it is not at all obvious to me how much faster it would be. I also suspect that it would be possible to improve performance with the current API (although I don't expect it to be easy because the current API is over-complicated), in which case the performance argument would be weaker.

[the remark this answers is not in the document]
astrofrog: Is the plan to only try an implementation for this once the APE is accepted? For similar APEs/PEPs/etc there is sometimes a prototype implementation, which would allow this to be quantified.

APE26.rst
```diff
@@ -0,0 +1,325 @@
...
+        ...  # it's fast.
+
+    class SkyCoord(BaseCoordinate):
+
+        def __init__(...):  # flexible input parsing
+            ...
+
+Branches and pull requests
+--------------------------
+No direct progress on these changes has yet occurred. Discussion
```
astrofrog, line 192: How much effort would it be to do a prototype implementation? Are the proposed changes here actually going to amount to a lot of lines of code if we exclude changes to tests, docs, and deprecation warnings/layers?

APE26.rst
eerovaher: Am I supposed to learn that from reading this paragraph?

jeffjennings:
Thank you all for your comments. At this point we will leave this PR open and complete a prototype implementation, and return when that is ready and can be demoed. In the interim, we will not address further comments left here. 

- APE 26 authors (@adrn, @jeffjennings, @mhvk, @nstarman)

APE26.rst
nstarman: Yes, in part. See the lower pseudo-code that demonstrates how simple `Coordinate` will be as a subclass of the common base class. That and the assumption that for changes large enough to merit an APE there would need to be a reasonable amount of explanation. For example, even a `See Also` is very useful for helping users decide when to use a `Coordinate` versus a `SkyCoord`. Combined with further docstring and RTD changes which are the natural course of a PR, and I'm not so worried.

APE26.rst
jeffjennings: Done in [d801d8d](https://github.com/astropy/astropy-APEs/pull/112/commits/d801d8dba8f52d81878d3b54c384f9a1be8b862b).

APE26.rst
jeffjennings: Done in [a7c80f2](https://github.com/astropy/astropy-APEs/pull/112/commits/a7c80f2b3ce36297d0ee3904b718ef0cf71c34a8).

APE26.rst
jeffjennings: Updated to be correct and add your point in [9332ae6](https://github.com/astropy/astropy-APEs/pull/112/commits/9332ae62c91c54ed09c4c0b64f7c72c6833687cd).

APE26.rst
jeffjennings: Changed to Nathaniel's text in [cd5471b](https://github.com/astropy/astropy-APEs/pull/112/commits/cd5471b1092a3aae4ae31ce9f3b5348df7e91319).

APE26.rst
jeffjennings: Updated the text accordingly in [062755f](https://github.com/astropy/astropy-APEs/pull/112/commits/062755fab13a96d06e511507412c04eee929b32e).

pllim:
> and return when that is ready and can be demoed

In the light of this development, CoCo will hold off on further reviews for now. Thank you.

pllim:
Possible updates from Astropy Coordination Meeting 2025:

* `ICRS(ra, dec)` returns a frame now; proposal is to keep it around but returning a `SkyCoord` in the future (on the same deprecation timescale) cc @keflavich 
* Clarify more explicitly that your planned deprecation is at least 2 years, not just 1 year.

jeffjennings:
> Possible updates from Astropy Coordination Meeting 2025:
> 
>     * `ICRS(ra, dec)` returns a frame now; proposal is to keep it around but returning a `SkyCoord` in the future (on the same deprecation timescale) cc @keflavich
> 
Our proposed deprecation stages for `ICRS` and all other frames are: 
- keep current API, but internally call the new data-less equivalent class (e.g., `ICRSFrame` for `ICRS`), while emitting a deprecation warning when instantiated. 
- still warn, but return a ``Coordinate``, not an instance of its class type (by overriding ``__new__``)
- remove the legacy frame classes after a suitable deprecation period

Keeping `ICRS` would unfortunately generate much user confusion during and after the full deprecation cycle, as there would now be `ICRS` and `ICRSFrame`. The docstrings and docs would have to explain the difference, and even then there would be confusion. An extensive deprecation cycle for these changes (comment quoted immediately below) has already been proposed, and a search-replace for `ICRS` to `SkyCoord` in downstream code would not be prohibitive.

>     * Clarify more explicitly that your planned deprecation is at least 2 years, not just 1 year.

APE26.rst
jeffjennings: The prototype has now been implemented (see comment below). The proposed changes here would remove many more lines of legacy code than new lines added (the new `Coordinate` class for example is by design minimal), and the bulk of the refactoring is in moving data-aware methods/attributes/properties out of `SkyCoord` and into a new base class (`BaseCoordinate`).

jeffjennings:
Hi all, we now have a prototype and demo available for review. Please provide any comments within **2 weeks, by November 14.** 

We recommend reading through the demo notebook and then taking a look at the prototype implementation as needed (see the APE's "Implementation" section for links). Note that the implementation is not exhaustive, e.g., `SkyCoord` is not fully populated - there's just enough to demonstrate the new framework. 

Signed, 
APE 26 authors (@adrn, @jeffjennings, @mhvk, @nstarman)

APE26.rst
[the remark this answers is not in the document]
nstarman: @jeffjennings made https://github.com/jeffjennings/ape26_scratch/blob/main/demo.ipynb :)

APE26.rst
nstarman: `Coordinate` is fast too! 🎉

APE26.rst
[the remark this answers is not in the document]
eerovaher: The only measurements in that notebook are in code cell 11 and they are comparing the new `Coordinate` and the current `SkyCoord`. But if `Coordinate` is supposed to be the replacement for the current `BaseCoordinateFrame` instances then that is what you should be comparing `Coordinate` with.

It's fine to mention performance in discussions but I wouldn't bring it up in the APE text itself because performance depends a lot on implementation details that are beyond the scope of an APE.

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+relative to another reference frame, which is used to transform 
+coordinates between the reference frames that exist in the 
+``astropy.coordinates`` ecosystem. For example, the ``AltAz`` frame 
+class, which is used when referencing sky coordinates in altitude 
+and azimuth, must contain a location on Earth and a time in order 
+to transform to and from an inertial reference frame like the ICRS. In 
+addition to reference frame information, these frame classes can 
+**also** contain coordinate data (positions, velocities, or other 
+differentials), which are stored internally using the 
+``Representation`` and ``Differential`` classes.
```
eerovaher: There is nothing in `astropy.coordinates` called `Representation` or `Differential`.
```suggestion
``BaseRepresentation`` and ``BaseDifferential`` subclasses.
```

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+think best to revise. We now view it as an insufficient 
+separation of concerns. The 
+coordinate frame class deals with two separate issues: the 
+definition of the frame and the storage of coordinate data within 
+that frame. Ideally the code would be more modularly structured, 
+where the coordinate frame class **only** defines the reference 
+frame. It would then be the purview of another class – like 
+``SkyCoord`` or the newly-proposed ``Coordinate`` – to bring these 
+concerns together and to represent data (using ``Representation`` 
+and ``Differential`` classes) in a given reference frame (using a
```
eerovaher:
```suggestion
concerns together and to represent data (using ``BaseRepresentation`` 
and ``BaseDifferential`` subclasses) in a given reference frame (using a 
```

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+that frame. Ideally the code would be more modularly structured, 
+where the coordinate frame class **only** defines the reference 
+frame. It would then be the purview of another class – like 
+``SkyCoord`` or the newly-proposed ``Coordinate`` – to bring these 
+concerns together and to represent data (using ``Representation`` 
+and ``Differential`` classes) in a given reference frame (using a 
+``CoordinateFrame`` class). As a demonstration of the current state 
+of duplicated functionality, these two initializations effectively 
+represent the same thing, but return different objects with similar 
+but not identical APIs:
```
eerovaher, line 72: Provide examples of the differences.

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+
+    c1 = SkyCoord(1., 2., frame="icrs", units="deg")
+    c2 = ICRS(1. * u.deg, 2. * u.deg)
+
+We suggest the situation should instead be analogous to what is the 
+case for units, which know how to transform from one to another, but 
+are not concerned with how the values are stored (that belongs to 
+``Quantity``). Translating to coordinates, units are like 
+``CoordinateFrame``, the values are the coordinate data 
+(``Representations``), and ``Quantity`` is like ``SkyCoord``.
```
eerovaher:
```suggestion
(``BaseRepresentation``), and ``Quantity`` is like ``SkyCoord``.
```

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+
+Having both the frame classes as well as ``SkyCoord`` be able to
+store and handle data has resulted in a large amount of code
+duplication. It has also required duplicated or even quadrupled 
+tests, in order to test both ``BaseCoordinateFrame`` and 
+``SkyCoord`` methods with both ``BaseCoordinateFrame`` and 
+``SkyCoord`` arguments - where the tests have not caught every
```
eerovaher:
```suggestion
``SkyCoord`` arguments - where the tests have not covered every 
```

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+We suggest the situation should instead be analogous to what is the 
+case for units, which know how to transform from one to another, but 
+are not concerned with how the values are stored (that belongs to 
+``Quantity``). Translating to coordinates, units are like 
+``CoordinateFrame``, the values are the coordinate data 
+(``Representations``), and ``Quantity`` is like ``SkyCoord``.
+
+Having both the frame classes as well as ``SkyCoord`` be able to
+store and handle data has resulted in a large amount of code
+duplication. It has also required duplicated or even quadrupled
```
eerovaher: I don't think code duplication is as big of a problem as it used to be. But reducing code duplication has come at the cost of making `SkyCoord.__getattr__()` even more prevalent. It is worth mentioning that the reliance on `SkyCoord.__getattr__()` means there is at least one method that `SkyCoord` implements accidentally (https://github.com/astropy/astropy/issues/15643) and there might be more.

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+modes (with and without data), each exhibiting different
+behavior. For instance, some methods such as ``separation`` work
+fine with frames with data, while doing this on coordinate frames
+without data results in an error. While this multi-modal
+structure as motivated in `APE 5
+<https://github.com/astropy/astropy-APEs/blob/main/APE5.rst>`_
+can be seen as a benefit for interpretability of the logic, we
+also now see it from the perspective of an anti-pattern. It
+forces any code interacting with the frame classes to handle both
+cases (checking ``frame.has_data``), complicating the codebase.
```
eerovaher, line 133: I suspect that in practice it is more common to avoid code complications by simply assuming that all frames have data, or that none of the frames have data.

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+Classes like ``SkyCoord`` will be composed structures bringing
+together the reference frame (an instance of a ``BaseFrame``
+subclass) and the coordinate data (``BaseRepresentation``
+objects). We also introduce a new class, ``Coordinate``, which
+is akin to ``SkyCoord`` (containing both frame and data), but
+without extra features like keeping frame attributes not
+associated with the current frame, caching and flexible input
+parsing. In this way ``Coordinate`` operates very similarly to
+the current ``BaseCoordinateFrame`` objects when they have
+data, and is meant to be their direct replacement in the new
```
eerovaher, line 187: Is the purpose of `Coordinate` to be a replacement for current frame classes with data? Would it be sensible to define both `SkyCoord` and `Coordinate` if you were designing an API from scratch without any historical baggage?

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+        pass  # no frame attributes
+
+    class FK5Frame(BaseFrame):
+        equinox: TimeAttribute
+
+    # ------
+
+    class BaseCoordinate:
+        frame: BaseFrame
+        data: BaseRepresentation
```
eerovaher, line 208: The pseudocode would be much more informative with the [PEP 695 Type Parameter Syntax](https://peps.python.org/pep-0695/).

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+
+Additionally, multiple pull requests have factored out common 
+code between frames and ``SkyCoord``, showing that there is no 
+proper separation of concern:
+
+- `Allow BaseCoordinateFrames to be stored in tables (by giving 
+  them .info) #16831 <https://github.com/astropy/astropy/pull/16831>`_
+- `Masked frames and SkyCoord #17106 
+  <https://github.com/astropy/astropy/pull/17016>`_ (this was later 
+  removed and instead methods were duplicated)
```
eerovaher, line 244: Use an example that was merged. There should be plenty to choose from.

APE26.rst
```diff
@@ -0,0 +1,331 @@
...
+   ``SkyCoord`` objects.
+
+4. Deprecating the legacy with-data frame classes, and eventually 
+   removing them after a deprecation period that adheres to 
+   `APE 2 <https://github.com/astropy/astropy-APEs/blob/main/APE2.rst>`_.
+
+   - Emitting warnings when instantiated.
+
+   - Still warn, but return a ``Coordinate``, not an instance of 
+     its class type (by overriding ``__new__``)
```
eerovaher: I don't see how overriding `__new__()` in such a manner could possibly be a good idea. If you want to make the change more gradual than APE 2 requires then just make the deprecation period longer.

eerovaher (commented):
1. If the `Coordinate` class is inherently a good idea then it shouldn't need the backing of an APE to get it implemented. If it exists mostly as a replacement for `BaseCoordinateFrame` instances with data then I'm not fully convinced it is required at all. Either way, it seems to me the existence of `Coordinate` should not be dictated by the APE. If there is some good reason to include it in the APE then the APE should justify it better.

2. The design proposed here is the second option from the APE 5 "Alternatives" section. APE 5 seems to reject that option because of a claim that it is impossible to implement the "high-level" classes as generics (although APE 5 does not use the term "[generics](https://typing.python.org/en/latest/reference/generics.html)"). To me it is obvious that the claim is false, which is why it is surprising to me that no-one seems to have called it out when APE 5 was discussed. Nonetheless, because APE 5 does make that claim then this APE should have a paragraph that comments on that. Additionally, the demonstration notebook should contain examples of both spherical representations that do not use RA and Dec, and also non-spherical representations, because that is what APE 5 claims will be difficult to implement.

kbwestfall:

Your chief APE editor is Kyle Westfall (@kbwestfall).

APE26.rst
nstarman: I think so, they aim for slightly different things: a WYSIWYG implementation and something with a bit more user-ergonomics optimizations: more flexible init, caching, etc.
Perhaps the flexible init could be a function, e.g. `make_coordinate()`, but in combination with other features, like caching, IMO a class is a good choice too.

APE26.rst
jeffjennings: Added note on this in [8f7711a](https://github.com/astropy/astropy-APEs/pull/112/commits/8f7711a69e41fc14611b0780332f0aaef9b94e24)

APE26.rst
jeffjennings: Added note on this in [ae93078](https://github.com/astropy/astropy-APEs/pull/112/commits/ae930787c8e44db11a01fad1042705863146e4f9)

APE26.rst
jeffjennings: Sorry if there's any confusion in the text; both examples given ([#16831](https://github.com/astropy/astropy/pull/16831) and [#17016](https://github.com/astropy/astropy/pull/17016)) were merged.

APE26.rst
jeffjennings: Hmm I think the pseudocode is clear in conveying inheritance, but the accompanying prototype now linked in the APE is yet another reference.

APE26.rst
jeffjennings: Yes although both cases still have to be supported in the current framework, while the proposed changes would avoid the issue both in practice and conceptually.

APE26.rst
jeffjennings: Added example in [fa2df27](https://github.com/astropy/astropy-APEs/pull/112/commits/fa2df27a8b7001825e1d883be53146ae3ee47ec4)

APE26.rst
mhvk: I think it might still be useful to include the example code that @nstarman wrote above.

jeffjennings: Thanks Marten, I've added a section for this in [5a14258](https://github.com/astropy/astropy-APEs/pull/112/commits/5a14258f896596b3bd93fd4110b02e587c3ff5b7), and also added links to the prototype and demo notebook.

jeffjennings:
> 1. If the `Coordinate` class is inherently a good idea then it shouldn't need the backing of an APE to get it implemented. If it exists mostly as a replacement for `BaseCoordinateFrame` instances with data then I'm not fully convinced it is required at all. Either way, it seems to me the existence of `Coordinate` should not be dictated by the APE. If there is some good reason to include it in the APE then the APE should justify it better. 
> 

Thanks for your comments Eero. The purpose of the APE has been to propose a fairly large set of changes to the approach and implementation of a core part of `astropy.coordinates`, motivating why the changes diverge from parts of APE 5. And we wanted to discuss whether there's support for these changes. The scope of explaining and justifying `Coordinate` seems more suited to an APE than just a set of PRs in my opinion. The discussion of `Coordinate` in the APE of course does not require that `Coordinate` cannot in the future be changed or replaced. But I think the APE clearly motivates why `Coordiante` is useful in addition to `SkyCoord` and why it makes sense to effectively replace the existing `BaseCoordinateFrame`. Nathaniel's [comment](https://github.com/astropy/astropy-APEs/pull/112#discussion_r2721576172) summarizes some of this very well.

>     2. The design proposed here is the second option from the APE 5 "Alternatives" section. APE 5 seems to reject that option because of a claim that it is impossible to implement the "high-level" classes as generics (although APE 5 does not use the term "[generics](https://typing.python.org/en/latest/reference/generics.html)"). To me it is obvious that the claim is false, which is why it is surprising to me that no-one seems to have called it out when APE 5 was discussed. Nonetheless, because APE 5 does make that claim then this APE should have a paragraph that comments on that. Additionally, the demonstration notebook should contain examples of both spherical representations that do not use RA and Dec, and also non-spherical representations, because that is what APE 5 claims will be difficult to implement.

Great, I've added a note on your APE 5 point to this APE, and added examples with cylindrical coordinates to the demo.

jeffjennings:
> With the acceptance of the revisions to APE 1, we're going to start following the new APE proposal process. While all of the APE editors (for now exactly the same as the CoCo) will contribute to the process, one of the APE editors will be the point-person (chief APE editor) shepherding your team through the process.
> 
> Your chief APE editor is Kyle Westfall (@kbwestfall).
> 
> You can review the next steps in the APE process [here](https://github.com/astropy/astropy-APEs/blob/main/APE1.rst), but just to briefly summarize them here:
> 
>     * Please finalize the draft of your APE and then ping @astropy/ape-editor-team.  This will prompt the APE Editorial team to read the proposed APE and provide any basic editing comments (led by your chief editor).
> 
>     * Following APE 1, this editing process will take no more than a month, but we actually expect this to go much faster for this APE.
> 
>     * Once the basic editing is finished, APE 1 provides a detailed list of the next steps.  The main thing to be aware of is that the APE status will be set to Discussion, the PR will be merged, and the APE editor will send an e-mail to astropy-dev to start the discussion period.  **All discussion should happen on the astropy-dev list**.
> 
>     * Throughout the discussion period, it is up to the APE authors to track the discussion and make updates to the APE.  These changes can be committed in a series of new PRs or kept in a branch that only leads to a single PR after the discussion has reached consensus.  Please keep your APE editor contact updated on your preferred approach.

We prefer implementing any such changes in a single PR.

> 
>     * There is no set timeframe for the discussion period, but it should be at least 2 weeks.

We anticipate 2 weeks will be sufficient.
> 
>     * After this minimum discussion period but otherwise at their discretion, the APE authors must then notify the CoCo that the APE proposal is ready for a final decision.
> 
>     * There is no set period for the CoCo review, but we expect it will take no more than 2-3 weeks.
> 
>     * The CoCo will decide to accept, reject, or request more discussion on specific items in the proposal, which will restart the process from the start of the discussion period.
> 
> 
> Please let us (either your chief APE editor or the full CoCo) know if you have any questions!

Hi @astropy/ape-editor-team and @kbwestfall, thank you for all of this. I've just finalized the APE draft and left a few comments. So we're ready to move forward.

APE26.rst
```diff
@@ -0,0 +1,451 @@
...
+operation as common as transforming reference frames:
+
+.. code-block:: python     
+
+    c1g = c1.transform_to(Galactic())  # Works
+    c2g = c2.transform_to(Galactic())  # Works
+    c1g = c1.transform_to("galactic")  # Works
+    c2g = c2.transform_to("galactic")  # Raises ConvertError
+
+We suggest the situation should instead be analogous to what is the
```
kbwestfall: This sentence is a little awkward.  Maybe instead: "We suggest the situation should instead be analogous to the structure implemented for astropy.units:  The ``Unit`` class is able to transform between units, but it is not concerned with how the associated values are stored (which is instead handled by ``Quantity``)."

APE26.rst
```diff
@@ -0,0 +1,451 @@
...
+``BaseCoordinateFrame`` ends up complicating ``SkyCoord``, which
+has to manage the coordinate data **through** the stored
+``BaseCoordinateFrame``.
+
+Another issue with the current implementation of coordinate
+frames is that the optional inclusion of coordinate data makes
+the reference frames “multi-modal”. This creates different usage
+modes (with and without data), each exhibiting different
+behavior. For instance, some methods such as ``separation`` work
+fine with frames with data, while doing this on coordinate frames
```
kbwestfall: Consider something like: "work fine for coordinate-frame instances that contain data, but they lead to faults for instances without data."

APE26.rst
```diff
@@ -0,0 +1,451 @@
...
+more clear, introspectable, and robust system.
+
+Finished Product
+----------------
+The end result of the implementation of this APE will be two separate
+hierarchies of classes: reference frame classes and coordinate
+classes which bring together a reference frame and coordinate
+data. We discuss each class type in turn.
+
+Coordinate frame classes only hold information pertaining to
```
kbwestfall: Here you use "coordinate frame classes," but you use "reference frame classes" above.  It might help clarity and readability to be more (or perhaps overly) consistent with the nomenclature.

APE26.rst
```diff
@@ -0,0 +1,451 @@
...
+
+We also provide a fuller `prototype 
+<https://github.com/jeffjennings/ape26_scratch/blob/main/ape26_prototype.py>` 
+and a
+`demo notebook <https://github.com/jeffjennings/ape26_scratch/blob/main/demo.ipynb>`
+implementing these changes.
+
+Usage patterns
+--------------
+The following pseudocode presents typical use cases of the
```
kbwestfall, line 368: I suggest including a reference to this section when you first introduce the implementation steps so that readers know examples are provided.

APE5.rst
```diff
@@ -28,6 +28,10 @@ provides a more user-friendly interface.  The transformation  framework will
 remain essentially unchanged, except that it will operate specifically on the
 low-level classes.
 
+UPDATE: Portions of the implementation presented here are amended in `APE 26
```
kbwestfall: I'm not exactly sure of the protocol, but I expect you need to add a `revised-by:` section at the top of the file that briefly describes this revision.  See APE 1 for an example.

You also might use an rst directive (like `note` or `important`) to highlight this.

(Although not an editing comment, this also begs the question about whether APE26 should be written to supersede APE5.  I don't know APE5 well enough to comment, but it may be odd to have an APE that is only superseded in part.)

kbwestfall (changes requested):
Here's my editorial review, which includes mostly minor wording recommendations.  Thanks!

APE26.rst
jeffjennings: Great, thanks Kyle! Updated in [a6fa590](https://github.com/astropy/astropy-APEs/pull/112/commits/a6fa590b32312305f1f792e25eed31c027a02e92)

APE26.rst
jeffjennings: Updated in [a6fa590](https://github.com/astropy/astropy-APEs/pull/112/commits/a6fa590b32312305f1f792e25eed31c027a02e92)

APE26.rst
jeffjennings: Updated in [a6fa590](https://github.com/astropy/astropy-APEs/pull/112/commits/a6fa590b32312305f1f792e25eed31c027a02e92)

APE26.rst
jeffjennings: Updated in [a6fa590](https://github.com/astropy/astropy-APEs/pull/112/commits/a6fa590b32312305f1f792e25eed31c027a02e92)

APE5.rst
jeffjennings: Good point -  I added a note to APE 5 in [19bfe90](https://github.com/astropy/astropy-APEs/pull/112/commits/19bfe90ec5f9b49ae07d41d69409369d8cb70b84). 

I agree that this APE only supersedes a part of APE5, and I think that trying to 'fully' supersede APE5 with APE26 might just end up being less clear than having readers of one refer to the other if needed to understand what has changed.

jeffjennings:
> Here's my editorial review, which includes mostly minor wording recommendations. Thanks!

Thank you for such a fast review! I've incorporated all of your suggestions, but of course let me know if I can do anything else before the next step.

kbwestfall (approved):

PR #118, opened 2025-07-16
APE0: Adding language about emeritizing voting members for non-voting

ceb8:
Per discussion at the Coordination Meeting, this PR adds language stating that the Coordination Committee can ask a voting member to move themself to emeritus status if they have not voted in the past 4 elections, even if they are otherwise active in the project.

Notes from the coordination meeting: https://docs.google.com/document/d/17zwz_7g4XdUOGR3oEtUQAqNbdYTDnxEhrxfDKM0mt7Q/edit?tab=t.0#heading=h.w0w9mixvcd44

### DIFF
--- a/APE0.rst
+++ b/APE0.rst
@@ -331,8 +331,11 @@ Term and Active/Emeritus Status
 Voting Members have no term limits. Voting Members who have stopped
 contributing are encouraged to declare themselves as *emeritus*. Those who have
 not made any significant contribution for two years may be asked to move
-themselves to the *emeritus* category by the Coordination Committee. If no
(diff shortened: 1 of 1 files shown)


jehturner:
I think this makes sense, but we might need some procedure for restoring emeritus members who become active again (which I don't think has really happened so far, but could if someone's commitments change). That shouldn't necessarily hold up this vote.

ceb8:
@jehturner Excellent point. Well, cross that bridge in the future. 😛

ceb8:
Merged based on successful Voting Member vote on the matter.

PR #120, opened 2025-08-20
APE 21: Capitalize status

pllim:
I think this should be uncontroversial. Just capitalizing so it is consistent with all the other accepted APEs.


ceb8 (approved):
LGTM

